PR #6, opened 2017-06-07
A very User Friendly, very Enterprise Spinner™

woju:
Since Admin API, qvm-ls takes a long time to complete. Therefore,
Corporate Headquarters commanded that a Enterprise Spinner is to be
implemented and mandated it's use unto us.

We take amusement from its endless gyrations.

### DIFF
--- a/qubesadmin/spinner.py
+++ b/qubesadmin/spinner.py
@@ -6,18 +6,17 @@
 # Copyright (C) 2017  Wojtek Porczyk <woju@invisiblethingslab.com>
 #
 # This program is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


qubesadmin/spinner.py
```diff
@@ -0,0 +1,148 @@
...
+
+#
+# The Qubes OS Project, https://www.qubes-os.org/
+#
+# Copyright (C) 2017  Wojtek Porczyk <woju@invisiblethingslab.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
```
marmarek: LGPL

marmarek (changes requested):

PR #7, opened 2017-06-08
Miscellaneous fixes around qvm-ls

woju:
There is not a mention about PV/HVM setting (other than `FLAGS` aka `STATUS`). Should that remain that way?

The old field `STATUS` got renamed to `FLAGS` not to confuse that with `STATE`, which is `get_power_state()`.


marmarek:
> There is not a mention about PV/HVM setting (other than FLAGS aka STATUS). Should that remain that way?

Yes, that's ok. All VMs will be HVM in 4.0. And if you want, you can enable "hvm" column.

marmarek (approved):

PR #8, opened 2017-06-14
Incremental development update

marmarek:
- features improvements
- fix qvm-template-postinstall (calling qubes.PostInstall service)

See also devel-2-qvm-run-1 and devel-2-qvm-run-2 branches - two different versions of qvm-run fix. Both fails on tests (no idea how to easily fix them), but both works otherwise.

### DIFF
--- a/qubesadmin/tools/qvm_template_postprocess.py
+++ b/qubesadmin/tools/qvm_template_postprocess.py
@@ -168,29 +168,36 @@ def post_install(args):
     if not args.skip_start:
         # just created, so no need to save previous value - we know what it was
         vm.netvm = None
-        vm.start()
+        # temporarily enable qrexec feature - so vm.start() will wait for it;
+        # if start fails, rollback it
+        vm.features['qrexec'] = True
         try:
-            vm.run_service_for_stdio('qubes.PostInstall')
-        except qubesadmin.exc.QubesVMError:
-            vm.log.error('qubes.PostInstall service failed')
-        vm.shutdown()
-        if have_events:
-            try:
-                # pylint: disable=no-member
-                qubesadmin.events.utils.wait_for_domain_shutdown(vm,
-                    qubesadmin.config.defaults['shutdown_timeout'])
-            except qubesadmin.exc.QubesVMShutdownTimeout:
-                vm.kill()
-            asyncio.get_event_loop().close()
+            vm.start()
+        except qubesadmin.exc.QubesException:
+            del vm.features['qrexec']
         else:
-            timeout = qubesadmin.config.defaults['shutdown_timeout']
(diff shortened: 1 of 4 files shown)


qubesadmin/tools/qvm_template_postprocess.py
```diff
@@ -168,29 +168,36 @@ def post_install(args):
...
+                vm.log.error('qubes.PostInstall service failed')
+            vm.shutdown()
+            if have_events:
+                try:
+                    # pylint: disable=no-member
+                    qubesadmin.events.utils.wait_for_domain_shutdown(vm,
+                        qubesadmin.config.defaults['shutdown_timeout'])
+                except qubesadmin.exc.QubesVMShutdownTimeout:
+                    vm.kill()
+                asyncio.get_event_loop().close()
```
woju, line 191: This line will hurt anyone who tried to run this with asyncio. Sorry I didn't catch this earlier.

marmarek: Someone needs to close event loop, any better candidate? This is standalone tool, not a library function.

woju: As to the "standalone tool", I'd imagine someone writing some automatic tool wouldn't want rewrite this function, but just `from qubesadmin.tools.qvm_template_postprocess import post_install`, even if undocumented. This line makes that impossible for no apparent reason.

As to the loop: who started it in the first place? It's used only in `qubesadmin.events.utils.wait_for_domain_shutdown()`, isn't it? If so, I think this is backwards. I'd rewrite `wait_for_domain_shutdown()` *and* `post_install()` as coroutines and `run_` the second one `_until_complete()` from `main()`, then close the loop. IOW, get the loop handling to the outermost context.

I don't think this would cause compat problems, for this package cannot be installed for python<3.4 anyway.

marmarek: See b3558836 including docstring for reasoning.

marmarek: Anyway, setting event loop can be extracted from there.

woju: Reasoning in that docstring is backwards. The loop belongs to `main()`.

If you'd follow the advice in docstring and really closed the loop, the function would throw up the second time it was called, because you cannot `.ensure_future()` on a closed loop, much less `.run_*()` it. `asyncio.get_event_loop()` does not create new loop.

woju: OK. But that wasn't the point of this pull request. I'll accept and file a bug for it.

woju (approved):

PR #12, opened 2017-06-26
Various improvements

marmarek:
- qvm-shutdown fixes
- qvm-firewall API update (depends on https://github.com/QubesOS/qubes-core-admin/pull/119)
- two-stage volume clone (depends on https://github.com/QubesOS/qubes-core-admin/pull/120)
- do not try to clone `uuid` property
- qvm-template-postprocess fixes

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -360,7 +360,6 @@ class QubesBase(qubesadmin.base.PropertyHolder):
                         raise
 
             try:
-                dst_vm.firewall.policy = src_vm.firewall.policy
                 dst_vm.firewall.save_rules(src_vm.firewall.rules)
             except qubesadmin.exc.QubesException as e:
                 self.log.error('Failed to set firewall: %s', e)
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -327,7 +327,7 @@ class QubesBase(qubesadmin.base.PropertyHolder):
             assert isinstance(dst_vm, qubesadmin.vm.QubesVM)
             for prop in src_vm.property_list():
                 # handled by admin.vm.Create call
-                if prop in ('name', 'qid', 'template', 'label'):
+                if prop in ('name', 'qid', 'template', 'label', 'uuid'):
                     continue
                 if src_vm.property_is_default(prop):
                     continue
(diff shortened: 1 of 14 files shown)


qubesadmin/tools/qvm_shutdown.py
```diff
@@ -48,37 +48,66 @@
...
-
-    args.app.log.info(
-        'Killing remaining qubes: {}'
-        .format(', '.join([str(vm) for vm in current_vms])))
-    for vm in current_vms:
-        vm.kill()
-
+    # care about order only when --wait is used:
+    if args.wait:
+        domains_and_distance = [(vm, distance_from_netvm(vm))
```
woju: I don't think this is right, since extensions can introduce other dependencies (like device attached for example) and disallow vm shutdown for that (or other) reason. This code assumes that the only dependency between vms is expressed as netvms.

I'd write it as (this is only a rough idea):
```python
@asyncio.coroutine
def shutdown_vm(vm):
  yield from vm.shutdown()
  return vm

for i in range(max_rounds): # maybe 7
  if not vms_to_shutdown:
    break
  done, pending = yield from asyncio.wait((shutdown_vm(vm) for vm in vms_to_shutdown), timeout=20)
  # (or whatever timeout is deemed OK)
  # TODO handle pending and don't call vm.shutdown() more than once
  for fut in done:
    try:
      vms_to_shutdown.remove(fut.result())
    except QubesException:  # if vm.shutdown() raises exception, it gets re-raised by fut.result()
      pass

# after max_rounds, they are fair game
if vms_to_shutdown:
  yield from asyncio.wait(vm.kill() for vm in vms_to_shutdown)
```

That way we allow for more flexible ordering.

woju (changes requested):

qubesadmin/tools/qvm_shutdown.py
marmarek: It is more complex than this - if `vm.shutdown()` raise an exception, VM should go to the next round. But if timeout, it should be killed, at the end of the same round. Otherwise for example hanging `untrusted` VM would block attempt to clean shutdown `sys-net` and `sys-firewall`.

marmarek: For now I've dropped qvm-shutdown rework from this PR, will work on it later. @woju please review other changes.

woju (approved):

PR #13, opened 2017-07-07
qvm-shutdown rework and bug fixes

marmarek:
The second approach to qvm-shutdown rework, discussed in https://github.com/QubesOS/qubes-core-admin-client/pull/12#discussion_r125371257


woju (approved):
LGTM

PR #14, opened 2017-07-07
tools/qvm-run: first approach

marmarek:
Launch stdin copy loop in a separate process (multiprocessing.Process)
and terminate it when target process is terminated.
Another idea here was threads, but there is no API to kill a thread
waiting on read().

This is alternative to #15. Both fails at tests.


woju (approved):

marmarek:
@woju what to do about tests here?

woju:
I dunno, maybe fix them? :stuck_out_tongue_winking_eye:

marmarek:
> I dunno, maybe fix them? 😜

Annotated with "expected failure".

PR #15, opened 2017-07-07
tools/qvm-run: second approach

marmarek:
Use 'cat' program to copy stdin to the child.

This is alternative to #14. Both fails at tests.


woju:
No good good reason for choosing one over another. This one is marginally more dependent on external tool, namely `cat`, being in $PATH. But if `cat` didn't work, users would have worse problems anyway.

woju:
Feel free to reopen if #14 wouldn't work out for some other reason.

PR #16, opened 2017-07-17
qvm-backup-restore and underlying code

marmarek:
Move qvm-backup-restore from qubes-core-admin and modify it to use Admin API.

This is only restore code, making backups will be in separate PR.

Fixes QubesOS/qubes-issues#1214


marmarek:
Test coverage isn't as bad as reported here - coverage isn't collected from multiprocess.Process instances. And adding `--concurrency=multiprocessing` break tests for some reason.

PR #23, opened 2017-08-09
qvm-run --dispvm and other improvements

marmarek:
This is actual code for QubesOS/qubes-issues#2974
But while touching this, also few other fixes.

### DIFF
--- a/doc/manpages/qvm-features.rst
+++ b/doc/manpages/qvm-features.rst
@@ -65,11 +65,16 @@ this feature to :py:obj:`True` enable gui-agent based GUI (i.e. with support of
 tools installed inside of qube). Not setting this feature at all, enable showing
 VGA emulated output.
 
+Default: show emulated VGA output only
+
 qrexec
 ^^^^^^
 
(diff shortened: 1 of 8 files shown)


doc/manpages/qvm-features.rst
```diff
@@ -45,7 +45,96 @@ have some specific capability.
 Some extensions interpret the values as boolean. In this case, the empty string
 means :py:obj:`False` and non-empty string (commonly ``'1'``) means
 :py:obj:`True`. An absence of the feature means "default", which is
-extension-dependent.
+extension-dependent. In most cases the default value for feature is retrieved
+from a qube template.
+
+List of known features
```
woju, line 51: I'd like to see explicit "default: \<something\>" at the end of each block to know, what does it mean if there is no such feature set. Even if that would repeat something in the description, like gui.

And some mention that for the features mentioned this list is authoritative and if in some place there is another default, then said other place is wrong.

woju (approved):
LGTM

PR #27, opened 2017-09-05
tools: restore qvm-start --cdrom and similar options

marmarek:
Booting a VM from cdrom require attaching the device before VM startup,
which is possible only in persistent mode. But for qvm-start --cdrom adding
a cdrom only temporarily, use new update_persistence() function to convert
the assignment to temporary one.

This depends on QubesOS/qubes-core-admin/pull#146 and also will not work
without QubesOS/qubes-vmm-xen#12

Fixes QubesOS/qubes-issues#3055

### DIFF
--- a/doc/manpages/qvm-start.rst
+++ b/doc/manpages/qvm-start.rst
@@ -15,7 +15,7 @@
 Synopsis
 --------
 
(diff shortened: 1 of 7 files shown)


marmarek:
@woju ping

PR #28, opened 2017-09-27
qvm-start: exit if domain running and no --skip-if-running

yomimono:
An attempted fix for https://github.com/QubesOS/qubes-issues/issues/3130 .

### DIFF
--- a/qubesadmin/tools/qvm_start.py
+++ b/qubesadmin/tools/qvm_start.py
@@ -147,6 +147,10 @@ def main(args=None, app=None):
     for domain in args.domains:
         if args.skip_if_running and domain.is_running():
             continue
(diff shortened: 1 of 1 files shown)


qubesadmin/tools/qvm_start.py
```diff
@@ -147,6 +147,10 @@ def main(args=None, app=None):
     for domain in args.domains:
         if args.skip_if_running and domain.is_running():
             continue
+        elif domain.is_running():
+            exit_code = 1
+            parser.print_error("domain is already running")
```
marmarek, line 152: This message should include domain name, because you may specify multiple of them. I'll add after merging.

marmarek (approved):

PR #29, opened 2017-10-05
Avoid admin.vm.List calls when constructing wrapper objects

marmarek:
The main purpose of this PR is to avoid race condition in events handling - when application got an event, VM may be already removed. And at this point, it's impossible to get VM class, because admin.vm.List will fail. But with the current code, it is required for constructing wrapper object.
The main reason is domain-shutdown event for DispVMs, but the race applies to other cases too. And also importantly - some mgmt VMs may not be allowed to list domains, but still access some actions. This approach is simpler than filtering response to admin.vm.List.

This PR introduce "blind mode", where domain wrapper object is constructed without checking if given domain exists (when referenced by a name directly). In such a case domain class is not available, so return QubesVM instance. After some consideration, abandon individual classes for different VM types and always use QubesVM, even not in blind mode - to avoid various corner cases depending on enabling or not blind mode. In fact domain class is rarely used by Admin API _client_ side - all domain properties are enforced by the server side. In fact this also makes extending qubesd easier, because one does not need to provide appropriate extensions (new domain classes) to client side anymore.

The introduced blind mode is not enabled here by default. It requires careful evaluation on case-by-case basis (especially on error reporting - by default errors with blind mode could be very misleading). The plan is to enable it in qubes-dbus first. Then introduce generic --blind-mode switch to all qvm-* tools. And only then consider enabling in some other places by default (if needed).

This PR contains also few bug fixes, beside the above main functionality. If you like to have them in separate PR, just ask.

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -86,12 +86,14 @@ class VMCollection(object):
         if not self.app.blind_mode and item not in self:
             raise KeyError(item)
         if item not in self._vm_objects:
-            if self.app.blind_mode:
-                cls = qubesadmin.vm.QubesVM
-            else:
-                cls = qubesadmin.utils.get_entry_point_one(VM_ENTRY_POINT,
-                    self._vm_list[item]['class'])
-            self._vm_objects[item] = cls(self.app, item)
+            cls = qubesadmin.vm.QubesVM
+            # provide class name to constructor, if already cached (which can be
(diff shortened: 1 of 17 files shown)


marmarek:
This needs to be merged/released together with appropriate changes in qubes-manager (settings window) I will push tomorrow. And probably some package dependencies will need to be added.

PR #32, opened 2017-10-16
Add CentOS support and fix python3 dependencies for CentOS

fepitre:
python3-sphinx and python3-daemon has no python34 packages. is it strictly necessary for now?


marmarek:
sphinx is required to build manpages

fepitre:
Yes. As I told you in my recap mail, I only put the python2 version for these two packages.

PR #35, opened 2017-11-07
Make qvm-run work for non-blocking stdin

qubesuser:
The main process sometimes sets fd 1 to O_NONBLOCK, and since in the
terminal case fd 0 and 1 are the same fd, this also results in fd 0
being non-blocking, causing qvm-run to crash with EAGAIN.

So just make the code work for both blocking and non-blocking stdin.

### DIFF
--- a/qubesadmin/tools/qvm_run.py
+++ b/qubesadmin/tools/qvm_run.py
@@ -26,7 +26,7 @@ import sys
 
 import multiprocessing
 
-import fcntl
+import select
 
 import qubesadmin.tools
 import qubesadmin.exc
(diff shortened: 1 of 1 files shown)


qubesadmin/tools/qvm_run.py
```diff
@@ -114,10 +114,10 @@ def copy_stdin(stream):
...
     # multiprocessing.Process have sys.stdin connected to /dev/null, use fd 0
     #  directly
-    flags = fcntl.fcntl(0, fcntl.F_GETFL)
-    flags &= ~os.O_NONBLOCK
-    fcntl.fcntl(0, fcntl.F_SETFL, flags)
-    for data in iter(lambda: os.read(0, 65536), b''):
+    while True:
+        # select so this code works even if fd 0 is non-blocking
+        select.select([0], [], [])
+        data = os.read(0, 65536)
```
marmarek: You need to also handle read returning `b''` (EOF).

qubesadmin/tools/qvm_run.py
```diff
@@ -247,3 +247,4 @@ def main(args=None, app=None):
 
 if __name__ == '__main__':
     sys.exit(main())
+
```
marmarek: Stray empty line...

marmarek (changes requested):

qubesuser:
Oops, fixed.

marmarek (approved):

PR #37, opened 2017-11-10
Simple client optimizations

qubesuser:
This is a set of optimizations consisting of the changes in the "speed up qvm-ls" branch that don't make architectural or API changes.

In particular, startup time of all tools is reduced by dropping the use of the "console entry points" system and avoiding importing the pkg_resources module, which are very slow and don't seem to provide any added value, in favor of just generating simple scripts that directly import the desired module and call main

### DIFF
--- a/qubesadmin/tools/qvm_ls.py
+++ b/qubesadmin/tools/qvm_ls.py
@@ -58,7 +58,7 @@ class Column(object):
 
     def __init__(self, head, attr=None, doc=None):
         self.ls_head = head
-        self.__doc__ = doc if doc is None else qubesadmin.utils.format_doc(doc)
+        self.__doc__ = doc
 
         # intentionally not always do set self._attr,
         # to cause AttributeError in self.format()
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 
 import os
 import setuptools
+import setuptools.command.install
 import sys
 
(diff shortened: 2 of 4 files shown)


qubesadmin/tools/qvm_ls.py
```diff
@@ -58,7 +58,7 @@ class Column(object):
 
     def __init__(self, head, attr=None, doc=None):
         self.ls_head = head
-        self.__doc__ = doc if doc is None else qubesadmin.utils.format_doc(doc)
+        self.__doc__ = doc
```
marmarek, line 61: It was needed before throwing Admin API in between - there it applied also to all VM properties, where markup is used sometimes. But indeed now it isn't needed anymore.

setup.py
```diff
@@ -19,9 +20,29 @@ def get_console_scripts():
...
+               f.write(
+"""#!/usr/bin/python3
+from {} import main
+import sys
+if __name__ == '__main__':
+	sys.exit(main())
+""".format(pkg))
+
+           os.chmod(path, 0o755)
+        setuptools.command.install.install.run(self)
```
marmarek, line 45: core-admin could use this optimization too. It would be specifically good for `qrexec-policy` tool.

marmarek (approved):

PR #42, opened 2017-12-01
Fix issue #3303

tasket:
This allows qvm-backup-restore to perform verify-only operations whether or not archive contains vm names that match live qubesdb.
issue #3303

### DIFF
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -1775,7 +1775,8 @@ class BackupRestore(object):
                 restore_info['dom0'].good_to_go:
             vms_dirs.append(os.path.dirname(restore_info['dom0'].subdir))
             vms_size += restore_info['dom0'].size
-            handlers[restore_info['dom0'].subdir] = (self._handle_dom0, None)
+            if not self.options.verify_only:
+                handlers[restore_info['dom0'].subdir] = (self._handle_dom0, None)
         try:
             self._restore_vm_data(vms_dirs=vms_dirs, vms_size=vms_size,
                 handlers=handlers)
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -70,6 +70,7 @@ _re_alphanum = re.compile(r'^[A-Za-z0-9-]*$')
 _tar_msg_re = re.compile(r".*#[0-9].*restore_pipe")
 _tar_file_size_re = re.compile(r"^[^ ]+ [^ ]+/[^ ]+ *([0-9]+) .*")
 
+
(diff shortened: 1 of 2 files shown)


qubesadmin/backup/restore.py
```diff
@@ -1356,7 +1357,9 @@ def _restore_vm_data(self, vms_dirs, vms_size, handlers):
                     # This all means that if the file was correctly verified
                     # + decrypted, we will surely access the right file
                     filename = self._verify_and_decrypt(filename)
-                to_extract.put(os.path.join(self.tmpdir, filename))
+
+                if not self.options.verify_only:
+                    to_extract.put(os.path.join(self.tmpdir, filename))
```
marmarek: If not sending file to be extracted, it should be removed here. Otherwise you'll need a lot of space (=to hold the whole backup) for verification process.

tasket: Ah, missed that bc dir is removed if there is no exception.

I'll have it `unlink(filename)`. However, the current .enc file will still be left behind if a decrypt error occurs.

marmarek: > However, the current .enc file will still be left behind if a decrypt error occurs.

Yes, this is to ease finding out what happened. Maybe file is truncated, or sth? Do you think it is a bad idea? This may be also a file somehow manipulated by a 3rd-party...

tasket: Its a separate issue, but yes it can have privacy impact to leave 100M chunk there. Could clean it up if no `-v` or debug option, or print message stating data remains in $dir.

marmarek: +1 for cleaning if no verbose/debug enabled.

tasket: Was going to handle it in `_verify_and_decrypt()` etc, but now thinking it can be covered in `restore_do()` here:
```
        try:
            self._restore_vm_data(vms_dirs=vms_dirs, vms_size=vms_size,
                handlers=handlers)
        except QubesException as err:
            if self.options.verify_only:
                raise
            else:
                self.log.error('Error extracting data: ' + str(err))
                self.log.warning(
                    "Continuing anyway to restore at least some VMs")
```

However, that "Continuing" warning is confusing me. Its false... right? I don't see continuance from this point.

marmarek: You're right. This applied to v1 (Qubes 1.0) backup format, where metadata was restored after data. But support for this backup format is dropped.

tasket: Added a new commit.

marmarek (approved):

PR #44, opened 2017-12-29
VM removal safeguards (was: Add --yes...)

tasket:
Have `qvm-remove` prompt user unless `--yes` option specified.

### DIFF
--- a/qubesadmin/tools/qvm_remove.py
+++ b/qubesadmin/tools/qvm_remove.py
@@ -28,14 +28,26 @@ from qubesadmin.tools import QubesArgumentParser
 parser = QubesArgumentParser(description=__doc__,
                              want_app=True,
                              vmname_nargs='+')
+parser.add_argument("--yes", action="store_true", dest="no_confirm",
+    default=False, help="Do not prompt for confirmation")
+
 
 
 def main(args=None, app=None):  # pylint: disable=missing-docstring
     args = parser.parse_args(args, app=app)
-    for vm in args.domains:
-        del args.app.domains[vm.name]
+    go_ahead = ""
+    if not args.no_confirm:
+        print("This will completely remove the selected VM(s)...")
+        go_ahead = input("Are you sure? [y/N] ").upper()
 
-    return 0
+    if args.no_confirm or go_ahead == "Y":
+        for vm in args.domains:
+            del args.app.domains[vm.name]
+        return 0
+    else:
+        print("Remove cancelled.")
+        return 1
+
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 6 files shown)


qubesadmin/tools/qvm_remove.py
```diff
@@ -28,14 +28,26 @@
 parser = QubesArgumentParser(description=__doc__,
                              want_app=True,
                              vmname_nargs='+')
+parser.add_argument("--yes", action="store_true", dest="no_confirm",
```
marmarek: Lets name it as in other tools: --force, -f

qubesadmin/tools/qvm_remove.py
```diff
@@ -28,14 +28,26 @@
...
+
 
 
 def main(args=None, app=None):  # pylint: disable=missing-docstring
     args = parser.parse_args(args, app=app)
-    for vm in args.domains:
-        del args.app.domains[vm.name]
+    go_ahead = ""
+    if not args.no_confirm:
+        print("This will completely remove the selected VM(s)...")
```
marmarek, line 40: I'd repeat VMs list here.

marmarek (changes requested):

qubesadmin/tools/qvm_remove.py
marmarek: Or at least add short version: `-y`.

tasket:
Just thought I'd mention qvm_template_postprocess uses "--really" and qvm_backup uses "--yes"
:)

tasket:
Avoid cloning installed_by_rpm: Copying `True` status doesn't make sense here, so new VM gets default.

marmarek:
I've pushed this from my account for travis check. Results:
 - missing --force, -f in documentation (`doc/manpages/qvm_remove.rst`). 
 - unit tests needs to be adjusted - currently wait for interactive confirmation

https://travis-ci.org/marmarek/qubes-core-admin-client/builds/325474606

tasket:
OK, added to manpage.

tasket:
The testing framework is unfamiliar to me. I tried looking at test modules for commands I know accept input, but I don't have time to delve into it now.

marmarek:
I've pushed tests update. One remaining thing is pylint complaining about:
```
************* Module qubesadmin.tools.qvm_remove
R: 45, 4: Unnecessary "else" after "return" (no-else-return)
```

tasket:
Ah, OK.

Also, review request (add vm listing) is also satisfied, I think.

tasket:
I did a `pull` to get your commit, and then a `push`. What do you suggest?

marmarek:
> I did a pull to get your commit, and then a push. What do you suggest?

If you want to avoid merge commit, you could do that either in opposite order, or use `pull --rebase`.

marmarek (approved):

PR #47, opened 2018-01-10
qvm-device/block: Alias '--ro' to '--option read-only=yes'

rustybird:
It's quicker to type and compatible with pre-R4.0 habits/scripts.

[Note that this adds an `--ro` option to the `attach` help message for _all_ device classes. I don't know if there's a good way to restrict it to `block`.]


marmarek (approved):
It's ok. Maybe in the future some other device classes will have such option?

PR #49, opened 2018-01-12
Fix qvm-run --dispvm some-command

marmarek:
There were two similar problems related to waiting for user session in newly
created DispVM.

Fixes QubesOS/qubes-issues#3012

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -494,6 +494,7 @@ class QubesLocal(QubesBase):
         :param str user: username to run service as
         :param str localcmd: Command to connect stdin/stdout to
         :param bool wait: wait for remote process to finish
+        :param int connect_timeout: qrexec client connection timeout
         :rtype: subprocess.Popen
         '''
 
@@ -516,6 +517,8 @@ class QubesLocal(QubesBase):
             user = 'DEFAULT'
         if not wait:
             qrexec_opts.extend(['-e'])
+        if 'connect_timeout' in kwargs:
+            qrexec_opts.extend(['-w', str(kwargs.pop('connect_timeout'))])
         kwargs.setdefault('stdin', subprocess.PIPE)
         kwargs.setdefault('stdout', subprocess.PIPE)
         kwargs.setdefault('stderr', subprocess.PIPE)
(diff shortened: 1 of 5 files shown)


rustybird:
This wouldn't fix the `qvm-run-vm --dispvm` variety, right?

What do you think about a slightly different approach: Instead of mangling the command, the service could be changed to `qubes.VMShell+WaitForSession` for the dvm case, in both `__init__.py` and `qvm-run-vm`. Then let `qubes.VMShell` wait when given the `WaitForSession` argument. I can work on PRs if you want.

rustybird:
Proposed `qubes.VMShell` service:

    #!/bin/sh

    if [ "$1" = "WaitForSession" ]; then
        id -un | /etc/qubes-rpc/qubes."$1"
    fi </dev/null >/dev/null 2>&1

    exec /bin/bash

**Edit:** Changed `PATH=/usr/local/etc/qubes-rpc:/etc/qubes-rpc` to hardcoded `/etc/qubes-rpc`; changed `>&2` to `2>&1`.

marmarek:
> What do you think about a slightly different approach: Instead of mangling the command, the service could be changed to qubes.VMShell+WaitForSession for the dvm case, in both __init__.py and qvm-run-vm. Then let qubes.VMShell wait when given the WaitForSession argument. I can work on PRs if you want.

This is brilliant! Especially because it will not break qvm-run on non-updated templates (which would be the case for separate service).

rustybird:
Okay then, I'll submit the PRs tomorrow. :)

marmarek:
> Then let qubes.VMShell wait when given the WaitForSession argument.

Instead of modifying `/etc/qubes-rpc/qubes.VMShell` script, better set `wait-for-session=1` in `/etc/qubes/rpc-config/qubes.VMShell+WaitForSession`. This way qrexec-agent will wait for session _before_ starting `qubes.VMShell` service, so it will get appropriate environment initialized there.

rustybird:
> Instead of modifying `/etc/qubes-rpc/qubes.VMShell` script, better set `wait-for-session=1` in `/etc/qubes/rpc-config/qubes.VMShell+WaitForSession`.

Thanks for the hint - done.

- dom0: https://github.com/QubesOS/qubes-core-admin-client/pull/51
- VM: https://github.com/QubesOS/qubes-core-agent-linux/pull/84

PR #52, opened 2018-02-02
Tasket02

tasket:
Get dom0 restore working for v3 and v4 archives. This changes old restore path behavior to deposit entire dom0-home archive into a subdir.
_handle_dom0() signature is changed to accept stream (uncompressed tar archive).
https://github.com/QubesOS/qubes-issues/issues/3467


qubesadmin/backup/core2.py
```diff
@@ -348,7 +351,7 @@ def load(self):
 
         self.globals['default_kernel'] = tree.getroot().get("default_kernel")
 
-        vm_classes = ["AdminVM", "TemplateVm", "TemplateHVm",
+        vm_classes = ["AdminVm", "TemplateVm", "TemplateHVm",
```
marmarek, line 359: Ough, we definitely need some tests for this code...

marmarek (approved):

qubesadmin/backup/core2.py
tasket: FWIW, I did perform manual tests on my 3.2 archives. I looked at the qubes.xml and saw that the elements were being tagged "AdminVm" not "AdminVM" (IIRC 4.0 xml saves with the latter case). Change results in R3.2 dom0 being picked up instead of ignored, and resulting in-memory class is "AdminVM" same as with R4.0 handling.

Not sure what errors we would anticipate with a test... maybe dom0 missing from the selection list (which was the original problem, actually).

tasket: Note in core2.py:
~~~
        elif element.get('qid') == '0':
            kwargs['dir_path'] = element.get('dir_path')
            vm.klass = "AdminVM"
~~~

marmarek: > Not sure what errors we would anticipate with a test... maybe dom0 missing from the selection list (which was the original problem, actually).

Yes, exactly this. And since the restore no longer override original user home, we can even compare restored data.

> Note in core2.py:

Yes, but it only iterates over elements listed in `vm_classes`...

marmarek:
Pushed tests here.

PR #55, opened 2018-02-22
qvm-backup[-restore] fixes

rustybird:
!!! Depends on https://github.com/QubesOS/qubes-core-admin/pull/198 !!!

### DIFF
--- a/doc/manpages/qvm-backup-restore.rst
+++ b/doc/manpages/qvm-backup-restore.rst
@@ -53,11 +53,6 @@ Options
 
    Restore VMs that are already present on the host under different names
 
-.. option:: --replace-template=REPLACE_TEMPLATE
-
-    Restore VMs using another template, syntax:
-    ``old-template-name:new-template-name`` (might be repeated)
-
 .. option:: --exclude=EXCLUDE, -x EXCLUDE
 
     Skip restore of specified VM (might be repeated)
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -1872,6 +1872,8 @@ class BackupRestore(object):
                 continue
 
             for prop, value in vm.properties.items():
+                if prop == 'dispid':
+                    continue
                 # exclude VM references - handled manually according to
                 # restore options
                 if prop in ['template', 'netvm', 'default_dispvm']:
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -1883,6 +1883,12 @@ class BackupRestore(object):
                         del self.app.domains[new_vm.name]
                     continue
 
+            # restore this property early to be ready for dependent DispVMs
+            prop = 'template_for_dispvms'
+            value = vm.properties.get(prop, None)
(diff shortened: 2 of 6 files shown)


marmarek:
See travis log - some tests needs to be adjusted and also manual page updated.

rustybird:
I've fixed the `qvm-backup-restore` manpage and the `qvm-backup` tests, but can't wrap my head around the failing test for `qvm-backup-restore`. It must have something to do with `--rename-conflicting` no longer effectively being the default behavior, but adding that option to the `qubesadmin.tools.qvm_backup_restore.main()` call didn't change anything.

marmarek:
I think it's because you override `options.exclude`, instead of modifying it in-place. This means `exclude_list` variable in the test is not really used.
This is ok, modify the test to look for `mock_backup.options.exclude` instead.

rustybird:
That did the trick.

rustybird:
The three `restore-py:` commits seem to mostly fix DispVM restore. But it doesn't work when both a DispVM and its dvm template are being restored at the same time - the `template_for_dispvms` property is set too late:

```
qubesadmin.backup: -> Restoring crap-dvm...
qubesadmin.backup: -> Restoring disp4418...
qubesadmin.backup: Error restoring VM disp4418, skipping: template for DispVM (crap-dvm) needs to have template_for_dispvms=True
[...]
qubesadmin.backup: -> Done.
qubesadmin.backup: -> Please install updates for all the restored templates.

$ qvm-prefs crap-dvm template_for_dispvms
True
```

marmarek:
This can be probably fixed by adjusting `_restore_vms_metadata()` to also handle `template_for_dispvms` property (just after `add_new_vm()` call).

rustybird:
The latest two commits do just that. So far so good:

```
[...]
qubesadmin.backup: -> Restoring disp4418...
qubesadmin.backup: Extracting data: 0 to restore
qubesadmin.backup: Error extracting data: unable to read the qubes backup file /var/tmp/disp (2): b'tar: vm14: Not found in archive\ntar: Exiting with failure status due to previous errors\n'
qubesadmin.backup: -> Done.
qubesadmin.backup: -> Please install updates for all the restored templates.
```

I guess the (cosmetic?) tar error could be gotten rid of by always adding a `vmN` directory, even if empty, to the backup somehow.

rustybird:
BTW I'm currently waiting for a huge backup to complete, then I'll test if qvm-backup-restore can handle it.

rustybird:
> I guess the (cosmetic?) tar error

Not cosmetic, it messes with restore of other VMs.

> could be gotten rid of by always adding a vmN directory, even if empty, to the backup somehow.

~~~OMG what have I stepped into. At this point I'd rather just prevent DispVM backup again.~~~

rustybird:
Previous despair notwithstanding, everything works [now](https://github.com/QubesOS/qubes-core-admin/pull/198/commits/6886e70aa965d75c603e67004ebe058e0bf3f3ef).

marmarek (approved):
Thanks!

PR #56, opened 2018-02-25
qvm-backup-restore: also handle absent --ignore-missing

rustybird:
This was forgotten in 2d8bade.

### DIFF
--- a/qubesadmin/tools/qvm_backup_restore.py
+++ b/qubesadmin/tools/qvm_backup_restore.py
@@ -220,9 +220,8 @@ def main(args=None, app=None):
         # unreachable - error_runtime will raise SystemExit
         return 1
 
-    if args.ignore_missing:
-        backup.options.use_default_template = True
-        backup.options.use_default_netvm = True
+    backup.options.use_default_template = args.ignore_missing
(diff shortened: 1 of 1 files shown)


qubesadmin/tools/qvm_backup_restore.py
```diff
@@ -220,9 +220,8 @@ def main(args=None, app=None):
         # unreachable - error_runtime will raise SystemExit
         return 1
 
-    if args.ignore_missing:
-        backup.options.use_default_template = True
-        backup.options.use_default_netvm = True
+    backup.options.use_default_template = \
+    backup.options.use_default_netvm = args.ignore_missing
```
marmarek: Code style: either correctly indent line continuation, or assign `args.ignore_missing` to both separately.

marmarek (changes requested):

qubesadmin/tools/qvm_backup_restore.py
rustybird: > or assign `args.ignore_missing` to both separately.

Done.

marmarek (approved):

PR #64, opened 2018-04-09
qvm-backup-restore: Add --restore-prefix option

dsg22:
Add option to always prefix restored VM name with a given prefix.

Handy in concert with the paranoid restore mode, if you just want to grab files from an old restore without polluting the VM namespace if you're planning on setting up all new VMs.

### DIFF
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -745,6 +745,8 @@ class BackupRestoreOptions(object):
         #: automatically rename VM during restore, when it would conflict
         # with existing one
         self.rename_conflicting = True
+        # Prefix all names of restored VMs with given prefix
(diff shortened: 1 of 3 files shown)


qubesadmin/backup/restore.py
```diff
@@ -1459,9 +1464,10 @@ def restore_info_verify(self, restore_info):
 
             if not self.options.verify_only and \
                     vm_info.name in self.app.domains:
```
marmarek, line 1466: This adds prefix only to VMs that already exists on the system. I guess it isn't what you wanted.

marmarek (changes requested):

dsg22:
Closing this PR due to problems pointed out in review. I will open a new one once I have a new patch and have tested it better.

PR #67, opened 2018-04-30
tools/qvm-firewall: Show EXPIRE column in list output

pgerber:
Sample output:

```
[user@dom0 ~]$ qvm-firewall sys-whonix add --before 0 action=accept dsthost=8.8.8.8 proto=tcp dstports=53 expire=+1800
[user@dom0 ~]$ qvm-firewall sys-whonix add --before 0 action=accept dsthost=8.8.8.8 proto=udp dstports=53 expire=+1800
[user@dom0 ~]$ qvm-firewall sys-whonix
NO  ACTION  HOST            PROTOCOL  PORT(S)  SPECIAL TARGET  ICMP TYPE  EXPIRE  COMMENT
0   accept  8.8.8.8/32      udp       53       -               -          +1791s  -
1   accept  8.8.8.8/32      tcp       53       -               -          +1790s  -
2   drop    10.0.0.0/8      -         -        -               -          -       private address
3   drop    169.254.0.0/16  -         -        -               -          -       private address
4   drop    172.16.0.0/12   -         -        -               -          -       link local
5   drop    192.168.0.0/16  -         -        -               -          -       private address
6   drop    fc00::/8        -         -        -               -          -       site local
7   drop    fd00::/8        -         -        -               -          -       unique site local
8   drop    fe80::/10       -         -        -               -          -       link local
9   accept  -               tcp       -        -               -          -       -
10  drop    -               -         -        -               -          -       -
```

I also looked at the output with --raw, I'm wondering if it should be slightly adjusted. The output currently looks like this:

```
action=accept proto=udp dst4=8.8.8.8/32 dstports=50-50 expire=1525105863
action=accept proto=tcp dst4=8.8.8.8/32 dstports=50-50 expire=1525105729
action=drop dst4=10.0.0.0/8 comment=private address
action=drop dst4=169.254.0.0/16 comment=private address
action=drop dst4=172.16.0.0/12 comment=link local
action=drop dst4=192.168.0.0/16 comment=private address
action=drop dst6=fc00::/8 comment=site local
action=drop dst6=fd00::/8 comment=unique site local
action=drop dst6=fe80::/10 comment=link local
action=accept proto=tcp
action=drop
```

I'm not sure what the intention of --raw originally was but I often use it as a template for writing new rules. Hence, I'd like change `dst4` and `dst6` to `dsthost`  and the comments to be quoted (`comment='private address'`). That way the output can be used as input for `qvm-firewall … add` again. Since I didn't know if there is a particular reason for the output being that way, apart for it matching the underlying API, I wondered if this would be a change you'd be willing to accept. I'd be happy to implement it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 *.pyo
 build/
+doc/_build/
 pkgs/
 .coverage
 __pycache__
--- a/doc/manpages/qvm-firewall.rst
+++ b/doc/manpages/qvm-firewall.rst
@@ -31,11 +31,11 @@ Options
 
 .. option:: --reload, -r
 
-   force reloading rules even when unchanged
+   force reload of rules even when unchanged
 
 .. option:: --raw
 
-   Print raw rules when listing
+   in combination with :option:`--list`, print raw rules
 
 
 Actions description
@@ -45,7 +45,8 @@ Available actions:
 
 * add - add specified rule. See `Rule syntax` section below.
 
-* del - delete specified rule. Can be selected either by rule number using :option:`--rule-no`, or specifying rule itself.
+* del - delete specified rule. The rule to remove can be selected either by rule number using :option:`--rule-no`
+  or by specifying the rule itself using the same syntax used for adding it.
 
 * list - list all the rules for a given VM.
 
@@ -59,8 +60,8 @@ A single rule is built from:
  - action - either ``drop`` or ``accept``
  - zero or more matches
 
-Selected action is applied on given packet when all specified matches do match,
-further rules are not evaluated. If none of the rules match, default action
+Selected action is applied to packets when all specified matches match,
+further rules are not evaluated. If none of the rules match, the default action
 (``policy``) is applied.
 
 Supported matches:
@@ -76,9 +77,9 @@ Supported matches:
  - ``proto`` - specific IP protocol. Supported values: ``tcp``, ``udp``,
    ``icmp``.
 
- - ``dstports`` - destination port or ports range. Can be either a single port,
+ - ``dstports`` - destination port or ports range. Can be either a single port
    or a range separated by ``-``. Valid only together with ``proto=udp`` or
- ``proto=tcp``.
+   ``proto=tcp``.
 
  - ``icmptype`` - ICMP message type, specified as numeric value. Valid only
    together with ``proto=icmp``.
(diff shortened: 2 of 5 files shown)


marmarek:
`--raw` provide firewall rules as for [admin.vm.firewall.Get](https://www.qubes-os.org/doc/admin-api/). See also https://www.qubes-os.org/doc/vm-interface/#firewall-rules-in-4x

Maybe better add `dst4` and `dst6` handling to qvm-firewall?

Also take a look at travis output - pylint complains about one too long line...

pgerber:
Shortened the line, expecting Travis to be happy now.

Adding `dst4` and `dst6` seems reasonable to me. Doesn't solve the issue with spaces in comments, but unlike with `dst[46]` that need to be rewriting to `dsthost`, it's rather obvious that quoting is needed. So, let's do it that way, shouldn't be an intrusive change.

marmarek:
> it's rather obvious that quoting is needed

On qvm-firewall command - yes. But on Admin API level, comment is expected to be the last field, unquoted.

marmarek (approved):

pgerber:
@marmarek, I looked at the changes necessary to get `dst4`/`dst6` working. The code for distinguishing IPv4, IPv6 and hostnames is rather hacky. I guess to keep compatibility with Python2. I wondered when you're planning to phase out Python2. It might be worth waiting and properly implementing it using Python3's ipaddress module.

marmarek:
> I wondered when you're planning to phase out Python2

We need python2 at least as long as salt is python2. In theory salt >=2017.7 supports python3, but it is still packaged as python2 (at least in Fedora and Debian).

Note that tools are explicitly excluded from python2 compatibility (see README). Anyway, I don't see how it affect this issue. Translating `dst4`/`dst6` to `dsthost` does not strictly require distinguishing address family. The worst that can happen is `dst4` will accept v6 address and be translated to `dst6` by underlying layer (or the other way around). IMO not a problem at qvm-firewall tool layer.

pgerber:
> The worst that can happen is dst4 will accept v6 address and be translated to dst6 by underlying layer (or the other way around). IMO not a problem at qvm-firewall tool layer.

You're right, it shouldn't matter. I decided to just make `dsthost`, `dst4` and `dst6` synonyms.

I pushed another commit improving comprehensibility of the --help output and the manpage or at least I believe it is an improvement. Also, I made it so the term match and rule are clearly distinguished. Before the term rule was used for both, a list predicates that make up a rule as well as for the predicate itself. I now renamed predicates from rules to matches which is the term used in the manpage.

marmarek (approved):

PR #68, opened 2018-06-07
Change qvm_prefs to handle "None" as Property None

jimtahu:
This changes to accept the string none as the value None,
as if an empty string was entered. This allows setting the
netvm to "None" as described in QubesOS/qubes-issues#3942

### DIFF
--- a/qubesadmin/base.py
+++ b/qubesadmin/base.py
@@ -275,6 +275,8 @@ class PropertyHolder(object):
                 value = value.name
             if value is None:
                 value = ''
+            if str(value).lower() == "none":
+                value = ''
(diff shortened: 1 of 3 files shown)


marmarek (changes requested):
"none" string could be a proper *string* value for some properties, so this change is wrong, at least in this place.
If anything, `qvm-prefs` tool could interpret value `none` as None VM, maybe only for selected properties (those known to be of VM type - `netvm`, `default_dispvm`, `template`).

jimtahu:
Will do

marmarek (approved):

PR #73, opened 2018-07-18
Helper function that lists where a given VM is used

marmarta:
A helper function to list vm's usage added to the utils.

### DIFF
--- a/qubesadmin/utils.py
+++ b/qubesadmin/utils.py
@@ -130,7 +130,7 @@ def vm_usage(app, reference_vm):
 
     for prop in global_properties:
         if reference_vm == getattr(app, prop, None):
-            result.append([None, prop])
+            result.append((None, prop))
 
     vm_properties = ['template', 'netvm', 'default_dispvm']
 
@@ -138,6 +138,6 @@ def vm_usage(app, reference_vm):
         for prop in vm_properties:
             if reference_vm == getattr(vm, prop, None) and \
                     not vm.property_is_default(prop):
-                result.append([vm, prop])
+                result.append((vm, prop))
 
     return result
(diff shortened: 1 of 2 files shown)


qubesadmin/utils.py
```diff
@@ -115,3 +115,29 @@ def updates_vms_status(qvm_collection):
             # "mixed"
             return None
     return status
+
+# Helper function that returns a list of all the places a given VM is used in.
+# Output is a list of tuples (property_holder, property_name), with None as
+# property_holder for global properties
```
marmarek: This should be in docstring. pylint told you this already ;)

qubesadmin/utils.py
```diff
@@ -115,3 +115,29 @@ def updates_vms_status(qvm_collection):
...
+
+
+def vm_usage(app, reference_vm):
```
marmarek: Could you make this function name more specific? IMO `vm_usage` could be confused with disk usage.

qubesadmin/utils.py
```diff
@@ -115,3 +115,29 @@ def updates_vms_status(qvm_collection):
...
+    for prop in global_properties:
+        if reference_vm == getattr(app, prop, None):
+            result.append([None, prop])
+
+    vm_properties = ['template', 'netvm', 'default_dispvm']
+
+    for vm in app.domains:
+        for prop in vm_properties:
+            if reference_vm == getattr(vm, prop, None) and \
+                    not vm.property_is_default(prop):
```
marmarek, line 140: A little optimization: switch conditions order (most properties will have default value).

marmarek (changes requested):
Besides the comments above, this could really use some unit test (handling various corner cases, like missing property).

marmarek (approved):

PR #83, opened 2018-12-02
Simplify class

GammaSQ:
The `--class` option should accept lowercase-definitions, especially "dispvm", since all other commands use lowercase-options. (e.g. `qvm-run --dispvm`)
The easiest way to implement this backwards-compatibly was to lowercase the argument, titlecase it and replace "vm" with "VM". Because it was easy to implement, I decided to allow omission of the trailing "vm", so `dispvm` and `disp` would both be valid.

First commit is the functional implementation, second I changed the `--help-classes`-output to show lowercase, "vm"-removed class options.

### DIFF
--- a/qubesadmin/tools/qvm_create.py
+++ b/qubesadmin/tools/qvm_create.py
@@ -41,7 +41,7 @@ parser = qubesadmin.tools.QubesArgumentParser()
 
 parser.add_argument('--class', '-C', dest='cls',
     default='AppVM',
-    help='specify the class of the new domain (default: %(default)s)')
+    help='specify the class of the new domain, case insensitive (default: %(default)s)')
 
 parser.add_argument('--property', '--prop',
     action=qubesadmin.tools.PropertyAction,
@@ -97,6 +97,7 @@ def main(args=None, app=None):
     if args.help_classes:
         vm_classes = args.app.qubesd_call('dom0', 'admin.vmclass.List').decode()
         vm_classes = vm_classes.splitlines()
+        print("case insensitive, trailing 'VM' optional")
         print('\n'.join(sorted(vm_classes)))
         return 0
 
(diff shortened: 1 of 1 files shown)


GammaSQ:
For some unknown reason, I get errors when signing this commit. Will debug tomorrow.

qubesadmin/tools/qvm_create.py
```diff
@@ -131,23 +133,24 @@ def main(args=None, app=None):
             args.properties['label'],
             ', '.join(args.app.labels)))
 
+    cls = args.cls.lower().replace("vm", "").title() + "VM"
```
marmarek, line 136: This is risky, as it will replace all occurrences of "vm", not only the final one. While in the current set of classes it isn't a problem, the system is extensible and new classes can be added. For the same reason, I think forcing "VM" at the end regardless of what user provided is a bad idea.

qubesadmin/tools/qvm_create.py
```diff
@@ -97,7 +97,9 @@ def main(args=None, app=None):
     if args.help_classes:
         vm_classes = args.app.qubesd_call('dom0', 'admin.vmclass.List').decode()
         vm_classes = vm_classes.splitlines()
-        print('\n'.join(sorted(vm_classes)))
+        print("case insensitive, trailing 'VM' optional")
+        clsList=map(lambda x: x.lower().replace("vm", ""),sorted(vm_classes))
```
marmarek, line 101: See the comment below about replacing "vm".

marmarek (changes requested):
I think this is too far. There is no guarantee that all VM classes will be named this way. In fact, our current approach is CamelCase - it's mere coincidence that all the names are single words. If we ever decide to add `TemplateBasedVM` class for example (which in fact [is discussed](https://github.com/QubesOS/qubes-doc/pull/748#issuecomment-442814749)), this will break.

Additionally, adding such flexibility in one place only, would make more harm than good, because it will be inconsistent ("why using 'app' doesn't work?! it worked with qvm-create", or "is 'app' the same as 'AppVM'?").

On the other hand, convenient aliases for common cases would be a good idea. What about adding `--standalone` option, equivalent to `--class StandaloneVM`, and `--template` for `--class TemplateVM`? Bonus point - the `--standalone` will bring back compatibility with `qvm-create` from R3.2.

GammaSQ:
Did take up that suggestion, new pull request here: https://github.com/QubesOS/qubes-core-admin-client/pull/84
(though `--template` was not implemented for obvious reasons! ;) )

PR #84, opened 2018-12-03
qvm-create: added --standalone and --disp shortcuts

GammaSQ:
`--standalone` = `--class StandaloneVM`
`--disp` = `--class DispVM --label red --template`

Currently, both shortcuts take precedence over a specified `--class`. The `--label` and `--template` option of the `--disp`-shortcut are implemented using `args.properties.setdefault`, so specified `--label` and `--template` take precedence.

I'm uncertain which behavior would be expected of `--standalone --class [not StandaloneVM]`, so I'd change it according to other peoples expectations.

Sidenote: I'd like to label dispVMs with the usual red recycle-arrow, but don't know which label I would have to specify for that?

after a recommendation in a previous PR: https://github.com/QubesOS/qubes-core-admin-client/pull/83#pullrequestreview-180583699

### DIFF
--- a/qubesadmin/tools/qvm_create.py
+++ b/qubesadmin/tools/qvm_create.py
@@ -47,9 +47,9 @@ parser.add_argument('--standalone',
     action="store_true",
     help=' shortcut for --class StandaloneVM')
 
-parser.add_argument('--disp'
-    action=qubesadmin.tools.SinglePropertyAction,
-    help='alias for --class DispVM --label red --template <argument or default DVM>')
+parser.add_argument('--disp',
+    action="store_true",
+    help='alias for --class DispVM --label red')
 
 parser.add_argument('--property', '--prop',
     action=qubesadmin.tools.PropertyAction,
@@ -121,12 +121,11 @@ def main(args=None, app=None):
     if args.one_pool:
         pool = args.one_pool
 
-    if 'disp' in args.properties:
+    if args.disp:
         args.properties.setdefault('label', 'red')
         args.cls = 'DispVM'
-        args.properties.setdefault('template', args.properties['disp'])
 
-    if 'standalone' in args.properties:
+    if args.standalone:
         args.cls = 'StandaloneVM'
 
     if 'label' not in args.properties:
--- a/qubesadmin/tools/qvm_create.py
+++ b/qubesadmin/tools/qvm_create.py
@@ -43,6 +43,14 @@ parser.add_argument('--class', '-C', dest='cls',
     default='AppVM',
     help='specify the class of the new domain (default: %(default)s)')
 
+parser.add_argument('--standalone',
+    action="store_true",
+    help=' shortcut for --class StandaloneVM')
+
+parser.add_argument('--disp'
+    action=qubesadmin.tools.SinglePropertyAction,
+    help='alias for --class DispVM --label red --template <argument or default DVM>')
+
 parser.add_argument('--property', '--prop',
     action=qubesadmin.tools.PropertyAction,
     help='set domain\'s property, like "internal", "memory" or "vcpus"')
@@ -112,6 +120,14 @@ def main(args=None, app=None):
                     'Pool argument must be of form: -P volume_name=pool_name')
     if args.one_pool:
         pool = args.one_pool
+
+    if 'disp' in args.properties:
+        args.properties.setdefault('label', 'red')
+        args.cls = 'DispVM'
+        args.properties.setdefault('template', args.properties['disp'])
+
+    if 'standalone' in args.properties:
+        args.cls = 'StandaloneVM'
 
     if 'label' not in args.properties:
         parser.error('--label option is mandatory')
--- a/doc/manpages/qvm-create.rst
+++ b/doc/manpages/qvm-create.rst
@@ -24,6 +24,14 @@ Options
 
    Decrease verbosity.
 
+.. option:: --standalone
+
+   shortcut for --class StandaloneVM, see below
+
+..option :: --disp
+
+   shortcut for --class DispVM --label red, see below
+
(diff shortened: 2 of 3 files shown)


qubesadmin/tools/qvm_create.py
```diff
@@ -43,6 +43,14 @@
     default='AppVM',
     help='specify the class of the new domain (default: %(default)s)')
 
+parser.add_argument('--standalone',
+    action="store_true",
+    help=' shortcut for --class StandaloneVM')
+
+parser.add_argument('--disp'
+    action=qubesadmin.tools.SinglePropertyAction,
```
marmarek: I think this should be just `action="store"` (there is no "disp" property, so `SinglePropertyAction` is not applicable here).
(besides syntax error - missing coma - pointed by travis)
Additionally, in this form the argument isn't optional. For optional, add `nargs=argparse.OPTIONAL`. But see below about using optional arg at all.

qubesadmin/tools/qvm_create.py
```diff
@@ -113,6 +121,14 @@ def main(args=None, app=None):
     if args.one_pool:
         pool = args.one_pool
 
+    if 'disp' in args.properties:
```
marmarek:
```suggestion
    if args.disp:
```

qubesadmin/tools/qvm_create.py
```diff
@@ -113,6 +121,14 @@ def main(args=None, app=None):
...
+        args.properties.setdefault('label', 'red')
+        args.cls = 'DispVM'
+        args.properties.setdefault('template', args.properties['disp'])
+
+    if 'standalone' in args.properties:
```
marmarek:
```suggestion
    if args.standalone:
```

qubesadmin/tools/qvm_create.py
```diff
@@ -113,6 +121,14 @@ def main(args=None, app=None):
     if args.one_pool:
         pool = args.one_pool
 
+    if 'disp' in args.properties:
+        args.properties.setdefault('label', 'red')
+        args.cls = 'DispVM'
+        args.properties.setdefault('template', args.properties['disp'])
```
marmarek: `args.disp`. But IMO optional "template" argument to `--disp` may be confusing. Explicit `qvm-create --disp -t ...` is clearer.

marmarek:
> Sidenote: I'd like to label dispVMs with the usual red recycle-arrow, but don't know which label I would have to specify for that?

The recycle-arrow is added automatically based on VM class.

marmarek (changes requested):
In addition to comments inline, this could use some unit tests. See qubesadmin/tests/tools/qvm_create.py, `test_003_properties` looks like a good example to base new test on.

qubesadmin/tools/qvm_create.py
GammaSQ: I was going for a consistent coding-style with this. If this should be changed, it should be changed across the whole file, which would constitute another pull request IMO?

qubesadmin/tools/qvm_create.py
GammaSQ: Is it important _not_ to have the ability to specify a template w/o `-t`?

qubesadmin/tools/qvm_create.py
marmarek: It isn't about coding-style. This does completely different thing. `args.properties` is set by `SinglePropertyAction`, and is about actual VM properties (like `name`, `label`, etc). Not arbitrary name.

qubesadmin/tools/qvm_create.py
marmarek: Yes, because it isn't obvious what this argument is. This applies to (almost?) all optional arguments to options. If you add optional command positional arguments to the mix, it is not only confusing to human, but also to the computer - for example in `--opt arg1 arg2`, `arg1` could be interpreted as optional argument to `--opt`, or a positional argument and `--opt1` without its argument.

If you read (in documentation for example) `qvm-create --disp some-name other-name`, you really need to check some extra doc to understand what is going on. On the other hand, `qvm-create --disp --template=some-name other-name` is much clearer.

qubesadmin/tools/qvm_create.py
GammaSQ: Sorry, I misunderstood the code there.

qubesadmin/tools/qvm_create.py
GammaSQ: Okay, I get it, will do!

GammaSQ:
> In addition to comments inline, this could use some unit tests. See qubesadmin/tests/tools/qvm_create.py, `test_003_properties` looks like a good example to base new test on.

Will do.

> The recycle-arrow is added automatically based on VM class.

If I start VMs created with `qvm-create --class DispVM --label red ...`, they don't get a recycle-arrow. Is this a bug or do the python-tools behave differently from the commandline-tools?

marmarek:
> Is this a bug or do the python-tools behave differently from the commandline-tools?

I'd say so. On the other hand, those are slightly different cases:
 - "Disposable: ..." menu entries with recycle icon - those are for DVM template, where launching it create _new_ DispVM
 - DispVM itself menu entries (as I understand - no recycle icon) - launching it will start this very DispVM

The practical difference is very subtle - if you launch such menu entry multiple times, in the first case you'll get separate DispVMs for each call, in the second case you'll get just one DispVM for all of them.

marmarek:
Now [travis complains](https://travis-ci.com/QubesOS/qubes-core-admin-client/jobs/163815413#L3292) that man page (in doc/manpages/qvm_create.rst) lack doc for `--disp` and --standalone`...
The code looks good.

marmarek (approved):

PR #85, opened 2018-12-03
 get_label now accepts integer and has correct return-type 

GammaSQ:
implementation had several issues:
No integer would be accepted (`.isdigit` isn't implemented by `int`-type)
`i` is a `str`-type, so `i.index` is a function
return type would have been `str`, not `label.Label`

new implementation takes advantage of `dict.values()` function, which returns an array. `int(int)` returns `int` btw.

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -216,10 +216,8 @@ class QubesBase(qubesadmin.base.PropertyHolder):
             pass
 
         # then search for index
-        if label.isdigit():
-            for i in self.labels:
-                if i.index == int(label):
-                    return i
+        if type(label) == int or label.isdigit():
+            return self.labels.values[int(label)]
 
         raise KeyError(label)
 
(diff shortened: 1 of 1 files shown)


qubesadmin/app.py
```diff
@@ -216,10 +216,8 @@ def get_label(self, label):
             pass
 
         # then search for index
-        if label.isdigit():
-            for i in self.labels:
-                if i.index == int(label):
-                    return i
+        if type(label) == int or label.isdigit():
+            return self.labels.values()[int(label)]
```
marmarek: Label.index is _not_ the same as index in what `values()` returns - dict-like objects in Python generally don't have stable ordering. Just adding `isinstance(label, int)` (not `type(...)`) to the previous version would be enough.

marmarek (changes requested):

qubesadmin/app.py
GammaSQ: Ah! And there I was wondering why this code is so non-pythonian. Turns out, I completely misunderstood the problem. So I kept the ability to accept `int`-type arguments and corrected the for-loop. New commit added.

marmarek:
See travis:
```
************* Module qubesadmin.app
qubesadmin/app.py:224:11: C0123: Using type() instead of isinstance() for a typecheck. (unidiomatic-typecheck)
```

marmarek (approved):

PR #91, opened 2019-03-09
Remove policy option from man page

anadahz:
Firewall policy is now hardcoded to 'drop'.

QubesOS/qubes-issues#2869


marmarek (approved):
Thanks!

PR #93, opened 2019-05-26
qvm-prefs: possibility to only show non-default properties

maltesl:
qvm-prefs usually shows a large number of properties that are set to default,
so it would be useful to have a way to only show those properties that are
non-default. One could either add a new option like --show-only-non-default
(or perhaps --hide-default), or instead change the normal behavior and recover
the old one with e.g. --show-all.
As there was already a TODO in the source file to only show non-default
properties, I went with the second option, but I would be happy to rework
to go with the first one.

### DIFF
--- a/doc/manpages/qubes-prefs.rst
+++ b/doc/manpages/qubes-prefs.rst
@@ -19,6 +19,10 @@ Options
 
    List available properties with short descriptions and exit.
 
+.. option:: --hide-default
+
+   Do not show properties that are set to the default value.
+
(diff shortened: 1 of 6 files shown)


marmarek:
Better to have the opposite option - show all properties by default (preserve current behavior), but add an option to show only non-default ones. `--hide-default` is ok.

maltesl:
Ok, I reworked it now to add a `--hide-default` option instead.

marmarek (approved):

PR #94, opened 2019-06-07
Add required packages for bionic build

unman:
Add required packages for bionic build.
Fix rogue comma in qvm-backup.rst
Closes QubesOS/qubes-issues#5081

### DIFF
--- a/Makefile.builder
+++ b/Makefile.builder
@@ -1,2 +1,10 @@
 RPM_SPEC_FILES := rpm_spec/qubes-core-admin-client.spec
-DEBIAN_BUILD_DIRS := debian
+ifneq ($(filter $(DISTRIBUTION), debian qubuntu),)
+	DEBIAN_BUILD_DIRS := debian
+	SOURCE_COPY_IN := source-debian-quilt-copy-in
+endif
+source-debian-quilt-copy-in:
+	if [ $(DIST) == bionic ] ; then \
+		sed -i '/python3-sphinx/a \ python3-xcffib,\n\ python3-daemon, ' $(CHROOT_DIR)/$(DIST_SRC)/debian/control ;\
+	fi
+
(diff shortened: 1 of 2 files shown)


Makefile.builder
```diff
@@ -1,2 +1,10 @@
 RPM_SPEC_FILES := rpm_spec/qubes-core-admin-client.spec
-DEBIAN_BUILD_DIRS := debian
+ifneq ($(filter $(DISTRIBUTION), debian qubuntu),)
+	DEBIAN_BUILD_DIRS := debian
+	SOURCE_COPY_IN := source-debian-quilt-copy-in
+endif
+source-debian-quilt-copy-in:
+	if [ $(DIST) == bionic ] ; then \
+		sed -i '/python3-sphinx/a \ python3-xcffib,\n\ python3-daemon, ' $(CHROOT_DIR)/$(DIST_SRC)/debian/control ;\
```
marmarek, line 8: Isn't it needed on plain Debian too? In such a case, better to add into `debian/control` directly. Generally, I'd like to avoid (if possible) sed-ing sources from Makefile.builder.

marmarek (changes requested):

Makefile.builder
unman: No it isn't - stretch and buster build fine without this.
I understand your reluctance though. Is there another way?

marmarek: The build indeed succeed, but it complains about those packages. See https://github.com/QubesOS/build-logs/blob/master/build-debian4/log_2019-03-07_03-50-16#L1801
So, better add them directly to debian/control.

unman: Ok - I've separated out the extraneous comma issue that was stopping build in bionic, and put in distinct [PR](https://github.com/QubesOS/qubes-core-admin-client/pull/95). 

So I'm wondering now if we _need_ those extra packages at all.
I'll investigate and see what's happening with/without (on various Debian/Ubuntu) - it may be that the warnings are of no great significance.

PR #96, opened 2019-06-15
WIP: Move parsing from qubesadmin.tools to qubesadmin.toolparsers

maltesl:
This is part of a project to deduplicate some similar code for parsing that occurs in both qubes-mgmt-salt-dom0-qvm and qubes-core-admin-client. For this it would be very helpful if the parsers used in `qubesadmin.tools` were accessible also in Python 2 (as the current salt modules we use are using Python 2), so this pull request is work in progress to move the parsers into their own module, `qubesadmin.toolparsers`, which can then be supported also under Python 2.

In order to preserve backward compatibility, `qubesadmin.tools` imports the functions and classes moved to `qubesadmin.toolparsers` into its namespace, and the tools also import their `get_parser` functions into their namespace. Some of this (particularly importing `get_parser` into the tools) might not be needed once the other places of the Qubes core stack that use these things have been changed.


marmarek (approved):
This generally looks fine, but lets see first if it's enough for https://github.com/QubesOS/qubes-mgmt-salt-dom0-qvm/pull/10

maltesl:
Closing as it is multiple years since I last worked on this.

PR #98, opened 2019-08-02
Refactor qvm-check

fepitre:


qubesadmin/tools/qvm_check.py
```diff
@@ -19,58 +20,93 @@
...
+    filters = get_filters(args)
+    filtered_domains = set(domains)
+    if filters:
+        for filt in filters:
+            status.append(filt['status'])
+            check = filt['check']
+            filtered_domains = filtered_domains.intersection(
+                [vm for vm in domains if check(vm)])
+
+        filtered_domains = list(filtered_domains)
```
marmarek: If you move this few lines below, you could avoid converting it back to `set` in the next line. Maybe even move it into `print_msg` function.

marmarek (changes requested):
You first unwrap long lines, to wrap them again few commits later. Please don't.

marmarek (approved):

PR #100, opened 2019-08-10
Listing class

fepitre:


qubesadmin/tools/qvm_device.py
```diff
@@ -272,6 +272,12 @@ def main(args=None, app=None):
     devclass = None
     if basename.startswith('qvm-') and basename != 'qvm-device':
         devclass = basename[4:]
+
+    curr_action = os.path.basename(sys.argv[1])
```
marmarek: basename isn't needed here, just look at `sys.argv[1]`.

marmarek (changes requested):
Besides comment below, please also add new action to qvm-device --help and the man page.

marmarek (approved):

PR #101, opened 2019-08-20
Check tmp free space before start extraction

### LINKED ISSUE #4791: qvm-backup-restore should check if there is enough disk space / should continuously make sure to not run out of disk space
### Qubes OS version:

R4.0

### Affected component(s):

dom0 qvm-backup-restore

---

Backup.
Restore the backup which fails for some reason.
Try restore backup again.
Will fail due to low disk space?

### Expected behavior:

Check for enough disk space.

### Actual behavior:

Run out of disk space.

### General notes:

Running out of disk space risks a no longer bootable system.

Files it stored temporarily in /var/tmp are not deleted when trying to verify a backup which fails. Another attempt then can lead to no longer having enough disk space.

---

donob4n:
Tries to fix https://github.com/QubesOS/qubes-issues/issues/4791

I think that the problem could be more complex if the destination pool is on the same device of the temporary directory. 

Also I am not sure if all the data is being extracted before deleting anything, according to my recent experience it seems the case but looking at the code I am not sure about this.

Well at least I think this does not hurt and it prevents to extract a lot of gigabytes for just failing after some hours.

### DIFF
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -1300,6 +1300,19 @@ class BackupRestore(object):
         self.log.debug("Working in temporary dir: %s", self.tmpdir)
         self.log.info("Extracting data: %s to restore", size_to_human(vms_size))
 
+        # check tmpdir has enough free space
+        statvfs = os.statvfs(self.tmpdir)
+        available_space = statvfs.f_frsize * statvfs.f_bavail
+
+        self.log.info("Available space on %s is %s", \
+                self.tmpdir, size_to_human(available_space))
+
+        if available_space * 0.95 < vms_size and \
+                not self.options.ignore_size_limit:
+            raise qubesadmin.exc.QubesException(
+               'Not enough free space on {}'.
+               format(self.tmpdir))
+
         # retrieve backup from the backup stream (either VM, or dom0 file)
         (retrieve_proc, filelist_pipe, error_pipe) = \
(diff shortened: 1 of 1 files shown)


marmarek:
> Also I am not sure if all the data is being extracted before deleting anything

What do you mean deleting?

donob4n:
> What do you mean deleting?

I mean removing unneeded (already restored) temporary files. I am not sure if https://github.com/QubesOS/qubes-core-admin-client/blob/769f8a5ee83c66db2a00fb85cd09d1ef794683a6/qubesadmin/backup/restore.py#L706 it is being executed for some temp files before the whole process ends.

In that case the free space restriction could be lowered, but if the temporary directory and the current pool are on the same device it should be hardened.

Do you known exactly when the temporary files are deleted?

marmarek:
>  it is being executed for some temp files before the whole process ends.

Yes, it's deleted as soon as the content is imported into the volume. Ideally (when decrypt, decompress and import is fast enough) there should be only a need for a single chunk of data (about 100MB). But in practice the opposite is true.

There are some ideas how to solve it without requiring disk space for the whole backup:
https://github.com/QubesOS/qubes-issues/issues/3230#issuecomment-340253679

donob4n:
Well, seeing that the space problem is not a real requirement I am closing this.

I will try to test with the "checkpoint" method.

PR #108, opened 2019-10-19
GuiVM

fepitre:
https://github.com/QubesOS/qubes-issues/issues/5662
https://github.com/QubesOS/qubes-issues/issues/833

### DIFF
--- a/qubesadmin/tools/qvm_start_gui.py
+++ b/qubesadmin/tools/qvm_start_gui.py
@@ -253,7 +253,7 @@ class GUILauncher(object):
         :param monitor_layout: monitor layout configuration
         """
         guivm = getattr(vm, 'guivm', None)
-        if guivm != vm.app.get_local_name():
+        if guivm != vm.app.local_name:
             vm.log.info('GUI connected to {}. Skipping.'.format(guivm))
             return
 
@@ -316,7 +316,7 @@ class GUILauncher(object):
         """Send monitor layout to all (running) VMs"""
         monitor_layout = get_monitor_layout()
         for vm in self.app.domains:
-            if getattr(vm, 'guivm', None) != vm.app.get_local_name():
+            if getattr(vm, 'guivm', None) != vm.app.local_name:
                 continue
             if vm.klass == 'AdminVM':
                 continue
@@ -329,7 +329,7 @@ class GUILauncher(object):
     def on_domain_spawn(self, vm, _event, **kwargs):
         """Handler of 'domain-spawn' event, starts GUI daemon for stubdomain"""
         try:
-            if getattr(vm, 'guivm', None) != vm.app.get_local_name():
+            if getattr(vm, 'guivm', None) != vm.app.local_name:
                 return
             if not vm.features.check_with_template('gui', True):
                 return
@@ -342,7 +342,7 @@ class GUILauncher(object):
     def on_domain_start(self, vm, _event, **kwargs):
         """Handler of 'domain-start' event, starts GUI daemon for actual VM"""
         try:
-            if getattr(vm, 'guivm', None) != vm.app.get_local_name():
+            if getattr(vm, 'guivm', None) != vm.app.local_name:
                 return
             if not vm.features.check_with_template('gui', True):
                 return
@@ -360,7 +360,7 @@ class GUILauncher(object):
         for vm in self.app.domains:
             if vm.klass == 'AdminVM':
                 continue
-            if getattr(vm, 'guivm', None) != vm.app.get_local_name():
+            if getattr(vm, 'guivm', None) != vm.app.local_name:
                 continue
             if not vm.features.check_with_template('gui', True):
                 continue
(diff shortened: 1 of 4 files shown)


qubesadmin/tools/qvm_start_gui.py
```diff
@@ -177,19 +181,22 @@ def common_guid_args(self, vm):
...
+        """Start GUI daemon (qubes-guid) connected directly to a VM
 
         This function is a coroutine.
 
         :param vm: VM for which start GUI daemon
         :param monitor_layout: monitor layout to send; if None, fetch it from
             local X server.
-        '''
+        """
+        if vm.guivm != vm.app.get_local_name():
```
marmarek: I think this needs to be (and already is) handled by a function calling it. But looks harmless.

qubesadmin/tools/qvm_start_gui.py
```diff
@@ -208,14 +215,17 @@ def start_gui_for_vm(self, vm, monitor_layout=None):
...
 
     @asyncio.coroutine
     def start_gui_for_stubdomain(self, vm, force=False):
-        '''Start GUI daemon (qubes-guid) connected to a stubdomain
+        """Start GUI daemon (qubes-guid) connected to a stubdomain
 
         This function is a coroutine.
-        '''
+        """
+        if vm.guivm != vm.app.get_local_name():
```
marmarek: And this also should be in the caller (but is harmless here).

qubesadmin/tools/qvm_start_gui.py
```diff
@@ -293,29 +311,35 @@ def send_monitor_layout(self, vm, layout=None, startup=False):
...
+                                    autostart=False))
         except subprocess.CalledProcessError as e:
             vm.log.warning('Failed to send monitor layout: %s', e.stderr)
 
     def send_monitor_layout_all(self):
-        '''Send monitor layout to all (running) VMs'''
+        """Send monitor layout to all (running) VMs"""
         monitor_layout = get_monitor_layout()
         for vm in self.app.domains:
+            if vm.guivm != vm.app.get_local_name():
```
marmarek: And here too.

qubesadmin/tools/qvm_start_gui.py
```diff
@@ -325,8 +349,10 @@ def on_domain_spawn(self, vm, _event, **kwargs):
             vm.log.warning('Failed to start GUI for %s: %s', vm.name, str(e))
 
     def on_domain_start(self, vm, _event, **kwargs):
-        '''Handler of 'domain-start' event, starts GUI daemon for actual VM'''
+        """Handler of 'domain-start' event, starts GUI daemon for actual VM"""
         try:
+            if vm.guivm != vm.app.get_local_name():
```
marmarek: This is a good place too. And also should handle exception nicely.

qubesadmin/tools/qvm_start_gui.py
```diff
@@ -293,29 +311,35 @@ def send_monitor_layout(self, vm, layout=None, startup=False):
...
                     continue
                 asyncio.ensure_future(self.send_monitor_layout(vm,
-                    monitor_layout))
+                                                               monitor_layout))
 
     def on_domain_spawn(self, vm, _event, **kwargs):
-        '''Handler of 'domain-spawn' event, starts GUI daemon for stubdomain'''
+        """Handler of 'domain-spawn' event, starts GUI daemon for stubdomain"""
         try:
+            if vm.guivm != vm.app.get_local_name():
```
marmarek: This is a good place for such check. But depending on the policy, this may get rejected and I think it should be handled here (by ignoring such VM too). Either add try/except as you do some below, or try `getattr(vm, 'guivm', None)` (in theory should return None if access was refused, but requires testing).

qubesadmin/tools/qvm_start_gui.py
```diff
@@ -268,7 +283,10 @@ def send_monitor_layout(self, vm, layout=None, startup=False):
             local X server.
         :param startup:
         :return: None
-        '''
+        """
+        if vm.guivm != vm.app.get_local_name():
```
marmarek: This is already handled in `send_monitor_layout_all`.

qubesadmin/tools/qvm_start_gui.py
```diff
@@ -239,27 +249,32 @@ def start_gui_for_stubdomain(self, vm, force=False):
...
 
         This function is a coroutine.
 
         :param vm: VM for which GUI daemon should be started
         :param force_stubdom: Force GUI daemon for stubdomain, even if the
             one for target AppVM is running.
-        '''
+        :param monitor_layout: monitor layout configuration
+        """
+        if vm.guivm != vm.app.get_local_name():
```
marmarek: This may be ok to leave (it's reachable directly via command line), but in this case, it should log some message why it's skipping the VM. Otherwise the call `qvm-start-gui abc` will be confusing why it does nothing.

marmarek (changes requested):
Besides comments in code, please separate code style change from functional changes into separate commits.

qubesadmin/tools/qvm_start_gui.py
fepitre: Yes it works.

marmarek (approved):

PR #112, opened 2019-11-03
Improvements to qvm-pool argument parser

### LINKED ISSUE #5407: Improve/rewrite qvm-pool arguments parser
qvm-pool arguments are inconsistent with other tools.

Things to change:
 - use action as a first positional argument (`add`, `info`, `remove` etc) instead of an option (`-a`, `-i` etc)
 - accept multiple pool options in separate `-o` arguments, instead of one `-o` with coma separated values

Consistency with other tools like `qvm-volume`, `qvm-device` etc.

**Additional context**
There need to be some kind of backward compatibility. I would use a tool `qvm-pool-legacy` accepting "old" options and redirect to it automatically when old action (`-a`, `-i`) is detected.

While at it, we should update https://www.qubes-os.org/doc/storage-pools/ (still R3.2).

https://www.qubes-os.org/doc/storage-pools/
https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-pool.html

SaswatPadhi:
This change introduces a complete rewrite of the argument parser for the `qvm-pool` tool.
As suggested by @marmarek in QubesOS/qubes-issues#5407, the goal is to be consistent with other tools such as `qvm-device`, `qvm-volume` etc.

(resolves QubesOS/qubes-issues#5407)

### DIFF
--- a/doc/manpages/qvm-pool.rst
+++ b/doc/manpages/qvm-pool.rst
@@ -1,11 +1,16 @@
 .. program:: qvm-pool
 
-:program:`qvm-pool` -- manage pools
+:program:`qvm-pool` -- Manage pools
 ===================================
 
 Synopsis
 --------
-:command:`qvm-pool` [-h] [--verbose] [--quiet] [--help-drivers] [-o options] [-l | -i *NAME* | -a *NAME* *DRIVER* | -r *NAME*]
+| :command:`qvm-pool` {add,a} [*options*] <*pool_name*> <*driver*>
+| :command:`qvm-pool` {drivers,d} [*options*]
+| :command:`qvm-pool` {info,i} [*options*] <*pool_name*>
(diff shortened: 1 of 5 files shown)


marmarek (changes requested):
Good work! Generally it looks good.

But this breaks any scripts/documentation written for R4.0, and we try to keep backward compatibility within major versions (4.x). Keeping both old and the new parser in the same tool would be awful. But I think it would be enough to to put the old tool into a separate file (`qvm_pool_legacy.py`?) and call it when the old syntax is detected (any old option for action, like `-a`).

SaswatPadhi:
Thank you, @marmarek. That makes sense. I should be able to implement the change and push it by end of today.

marmarek (approved):

PR #117, opened 2019-11-17
Fixed overeager assert

marmarta:
Pool usage details can actually be empty.


marmarek (approved):
This is ok, but I'll leave it open for now, as it's a logical place to move changes from https://github.com/QubesOS/qubes-desktop-linux-manager/pull/84

PR #122, opened 2020-01-14
Add --wait documentation for qvm-shutdown

tetrahedras:
Resolves qubes-issue #5561:
https://github.com/QubesOS/qubes-issues/issues/5561

qvm-shutdown's --wait option should be fully documented, explaining that it can be used to shutdown VMs in the correct sequence automatically.


tetrahedras:
I'm not really sure it makes sense to set up a GPG key for a *documentation* change...

marmarek (approved):

PR #124, opened 2020-01-15
Use admin.vm.CurrentState in client

### LINKED ISSUE #5166: qvm-ls: "memory" field/column does not output
**Qubes OS version**
R4.01 - current-testing

qvm-ls

**Brief summary**
The field/column of "memory" does not output when requested via the --fields parameter. However, qvm-prefs does show "memory" as a setting for the VMs.

**To Reproduce**
Open dom0 terminal
$ qvm-ls --fields name,memory,maxmem

**Expected behavior**
Three columns of output, with the middle column showing the memory setting

**Actual behavior**
Three columns of output, with the middle column always showing "-"

pwmarcz:
Corresponding PR in admin: QubesOS/qubes-core-admin#307.
Fixes QubesOS/qubes-issues#5166.

### DIFF
--- a/qubesadmin/tests/app.py
+++ b/qubesadmin/tests/app.py
@@ -916,11 +916,11 @@ class TC_30_QubesRemote(unittest.TestCase):
 
     @mock.patch('os.isatty', lambda fd: fd == 2)
     def test_014_run_service_no_autostart1(self):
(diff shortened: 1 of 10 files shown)


marmarek:
Pylint complains about missing docstring (you can copy from core-admin).

marmarek (approved):

PR #126, opened 2020-01-16
qvm-remove: remove misleading handling of QubesVMInUseError

pwmarcz:
Currently qvm-remove handles QubesVMInUseError by assuming that the
remove failed due to dependencies between VMs, and tries to recover
these. However, qubesd does not actually check these dependencies,
and throws QubesVMInUseError for different reasons (such as
"installed by package manager"), providing an explanatory message
which we disregard.

Instead, we should just display the message, as we do in the GUI.

See also QubesOS/qubes-core-admin#308 for which this was needed.


pwmarcz:
As discussed, I will not be removing any messages. I will make another PR for this because the description is going to be different.

PR #128, opened 2020-01-20
Import data using ImportWithSize call, not manual resizing

pwmarcz:
See the API call in QubesOS/qubes-core-admin#309.

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -1,4 +1,4 @@
-# -*- encoding: utf8 -*-
+# -*- encoding: utf-8 -*-
 #
 # The Qubes OS Project, http://www.qubes-os.org
 #
@@ -25,6 +25,7 @@ Main Qubes() class and related classes.
 import os
 import shlex
 import socket
+import shutil
 import subprocess
 import sys
 
@@ -37,8 +38,6 @@ import qubesadmin.storage
 import qubesadmin.utils
 import qubesadmin.vm
 import qubesadmin.config
-
-BUF_SIZE = 4096
 
 
 class VMCollection(object):
@@ -480,7 +479,8 @@ class QubesBase(qubesadmin.base.PropertyHolder):
         """
         Execute Admin API method.
 
-        Only one of `payload` and `payload_stream` can be specified.
+        If `payload` and `payload_stream` are both specified, they will be sent
+        in that order.
 
         :param dest: Destination VM name
         :param method: Full API method name ('admin...')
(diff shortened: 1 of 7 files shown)


pwmarcz:
Some benchmarks. I tried importing a 2GB image for a VM (non-sparse, containing a file full of random data). The command was `time qvm-volume import test-import:private data`.

The real time is a bit noisy but user/sys are more consistent:

|version|real|user|sys|
|-|-|-|-|
|current|49.3|1.9|6.2|
|current|62.0|2.0|5.2|
|new (default buf size, 64K)|57.9|5.6|13.5|
|new (default buf size, 64K)|54.3|5.7|14.2|
|new (buf size 1M)|55.0|2.3|9.4|
|new (buf size 1M)|70.6|2.3|10.0|

marmarek:
Can you try also from a VM? That's slightly different code path.

pwmarcz:
Here are the numbers when running from a VM:

|version|real|user|sys|
|-|-|-|-|
|current|105.5|2.7|7.1|
|current|86.9|2.7|6.5|
|new (default buf size, 64K)|80.7|5.1|11.9|
|new (default buf size, 64K)|127.4|5.0|12.1|
|new (buf size 1M)|98.1|3.6|11.3|
|new (buf size 1M)|95.5|3.7|12.6|

marmarek:
Since multiple processes, some in different VM are involved user and sys times are meaningless for overall performance (more important is the actual throughput, not CPU time used to achieve it).
Can you try to get more consistent results (real time)? IDK, maybe clear disk caches before the call (`echo 3 > /proc/sys/vm/drop_caches`)?

pwmarcz:
Here are the results for 3 runs each, from a VM, dropping the cache just before the import, and a relatively quiet system (I did not do anything on the machine during):

|version|real|
|-|-|
|current|80|
|current|94|
|current|100|
|new (64K)|86|
|new (64K)|108|
|new (64K)|102|
|new (1M)|70|
|new (1M)|106|
|new (1M)|103|

pwmarcz:
And here is what happened when I tried it in dom0 and even shut down noisy programs like Firefox:

|version|real|
|-|-|
|current|63|
|current|102|
|current|103|
|new (64K)|60|
|new (64K)|79|
|new (64K)|97|
|new (1M)|66|
|new (1M)|93|
|new (1M)|86|

marmarek:
It's interesting that the first call consistently is faster.
Anyway, 1) slowdown if any isn't big b) I don't see conclusive difference between 64K and 1M buffer, so I'd go with the default.

qubesadmin/app.py
```diff
@@ -658,20 +681,19 @@ def qubesd_call(self, dest, method, arg=None, payload=None,
...
-        if payload and payload_stream:
-            raise ValueError(
-                'Only one of payload and payload_stream can be used')
         service_name = method
         if arg is not None:
             service_name += '+' + arg
         p = subprocess.Popen([qubesadmin.config.QREXEC_CLIENT_VM,
                               dest, service_name],
-                             stdin=(payload_stream or subprocess.PIPE),
+                             stdin=subprocess.PIPE,
```
marmarek: Shouldn't this still connect `payload_stream` directly if no `payload` is set?

marmarek (changes requested):
And as in the other PR - please squash fixes of commits in the same PR

qubesadmin/app.py
pwmarcz: Okay. That does make the code more complicated (3 possible paths, not 2), but runs a bit faster and potentially avoids the deadlock in that case. Refactored.

marmarek (approved):

PR #130, opened 2020-01-21
Added force shutdown option to vm.shutdown

### LINKED ISSUE #5591: Regression: qvm-shutdown --wait --all displays a lot of warnings
**Qubes OS version**
4

qvm-shutdown

**Brief summary**
After the most recent dom0 update (stable repo) `qvm-shutdown --all --wait` started to display a lot of warnings about one VM depending on another. It didn't do that in the past. It still appears to be doing its job though.

**To Reproduce**
`qvm-shutdown --wait --all`

**Expected behavior**
All VMs shut down.

**Actual behavior**
All VMs shut down. A lot of red warnings from `notify-send` appear.

#5561, #4572

marmarta:
Furthermore makes qvm-shutdown --all use that option to force
shutdown and avoid unnecessary errors.

requires https://github.com/QubesOS/qubes-core-admin/pull/312
fixes QubesOS/qubes-issues#5591

### DIFF
--- a/qubesadmin/tools/qvm_shutdown.py
+++ b/qubesadmin/tools/qvm_shutdown.py
@@ -54,6 +54,8 @@ parser.add_argument('--timeout',
 def main(args=None, app=None):  # pylint: disable=missing-docstring
     args = parser.parse_args(args, app=app)
 
+    force = bool(args.all_domains)
+
     if have_events:
         loop = asyncio.get_event_loop()
     remaining_domains = args.domains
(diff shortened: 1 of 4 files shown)


marmarek:
It looks like this also fixes https://github.com/QubesOS/qubes-issues/issues/4572

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -54,6 +54,8 @@
 def main(args=None, app=None):  # pylint: disable=missing-docstring
     args = parser.parse_args(args, app=app)
 
+    force = bool(args.all_domains)
```
marmarek: Having this option supported in the backend, it would be useful to also allow using it directly via --force argument, not only implied by --all.

marmarek (changes requested):

marmarek (approved):

PR #131, opened 2020-01-21
Move qvm-console from qubes-core-agent-linux

aidecoe:
See QubesOS/qubes-core-agent-linux#191 where qvm-console is removed from qubes-core-agent-linux.


marmarek (approved):
I'll merge it together with https://github.com/QubesOS/qubes-core-agent-linux/pull/191

PR #136, opened 2020-03-24
Cloning a vm now clones persistent PCI device assignments

### LINKED ISSUE #4992: After cloning or renaming a qube, the devices are missing from the new qube
**Qubes OS version**

Qubes release 4.0 (R4.0)
Although I thought it was 4.0.1 specifically

The Qubes Manager - Copying/Renaming Templates

**Brief summary**
Cloning qubes and Renaming qubes do not include any "devices" that were part of the originating qube

**To Reproduce**
1. Install Qubes
2. Boot Qubes
3. Run Qube Manager
4. Right click a VM containing at least one item in the "devices" tab of Qube Settings (sys-net is a good example)
5. Select Qube Settings
6. Choose "Clone qube" or "Rename"
7. Enter a name for the VM
8. Right click on the qube with the name from step 7
9. Select Qube Settings
10. Select "Devices" Tab

**Expected behavior**
Expected to see the devices from the device in step 4

**Actual behavior**
Devices is empty

**Solutions you've tried**
Adding the device manually, it worked

This is pretty similar but it is different 
https://github.com/QubesOS/qubes-issues/issues/4978

marmarta:
fixes QubesOS/qubes-issues#4992

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -38,6 +38,7 @@ import qubesadmin.storage
 import qubesadmin.utils
 import qubesadmin.vm
 import qubesadmin.config
+import qubesadmin.devices
 
 
 class VMCollection(object):
@@ -316,7 +317,9 @@ class QubesBase(qubesadmin.base.PropertyHolder):
         return self.domains[name]
 
     def clone_vm(self, src_vm, new_name, new_cls=None, pool=None, pools=None,
-                 ignore_errors=False, ignore_volumes=None):
+                 ignore_errors=False, ignore_volumes=None,
+                 ignore_devices=False):
+        # pylint: disable=too-many-statements
         """Clone Virtual Machine
 
         Example usage with custom storage pools:
@@ -337,6 +340,7 @@ class QubesBase(qubesadmin.base.PropertyHolder):
             logged, or abort the whole operation?
         :param list ignore_volumes: do not clone volumes on this list,
             like 'private' or 'root'
+        :param bool ignore_devices: if True, do not copy device assignments
 
         :return new VM object
         """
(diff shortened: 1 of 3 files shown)


qubesadmin/app.py
```diff
@@ -316,7 +317,8 @@ def add_new_vm(self, cls, name, label, template=None, pool=None,
         return self.domains[name]
 
     def clone_vm(self, src_vm, new_name, new_cls=None, pool=None, pools=None,
-                 ignore_errors=False, ignore_volumes=None):
+                 ignore_errors=False, ignore_volumes=None, ignore_devices=True):
```
marmarek: Please make the default False. It is a change in behavior, but I think the previous one was a bug.

qubesadmin/app.py
```diff
@@ -337,6 +339,7 @@ def clone_vm(self, src_vm, new_name, new_cls=None, pool=None, pools=None,
             logged, or abort the whole operation?
         :param list ignore_volumes: do not clone volumes on this list,
             like 'private' or 'root'
+        :param bool ignore_devices: if False, do not copy device assignments
```
marmarek: This is inverted.

qubesadmin/app.py
```diff
@@ -472,6 +475,24 @@ def clone_vm(self, src_vm, new_name, new_cls=None, pool=None, pools=None,
...
+                for devclass in src_vm.devices:
+                    for assignment in src_vm.devices[devclass].assignments(
+                            persistent=True):
+                        new_assignment = qubesadmin.devices.DeviceAssignment(
+                            backend_domain=assignment.backend_domain,
+                            ident=assignment.ident,
+                            options=assignment.options,
+                            persistent=assignment.persistent,
+                            frontend_domain=dst_vm,
+                            devclass=assignment.devclass)
```
marmarek: You don't need to set frontend_domain and devclass - those will be handled automatically. But it isn't a problem either (there is an assert in `attach()` that prevents invalid value).

marmarek (changes requested):

marmarek (approved):

PR #138, opened 2020-04-27
[WIP]Add vm.icon property

marmarta:
This is a property for handling vm icons that change depending on
vm type.
Depends on https://github.com/QubesOS/qubes-artwork/pull/17

references QubesOS/qubes-issues#5767


marmarek (changes requested):
Please define new properties in core-admin repo, not here (use `@qubes.stateless_property` decorator there to have it exposed via Admin API). core-admin-client is supposed to be just a wrapper around Admin API, not carry any logic.

PR #140, opened 2020-05-29
Connect to PA in stubdom if audio-model enabled

easydozen:
and run pacat in low latency mode by default

QubesOS/qubes-issues#2624


marmarek (changes requested):
1. pylint complains about too long line (see travis output).
2. Duplication of getting xid for pacat IMO warrant a function for that. I know it's just one (long) line, but if it ever be modified to take some other properties into account, it will be easier to do it in one place.

marmarek (approved):

marmarek:
This looks good now. I'll leave those three PRs open for now, to include them in the next test-run on openQA.
We don't have much audio tests, and especially not for HVM without qubes packages installed, but lets check at least if it doesn't break audio on Linux.

PR #143, opened 2020-06-24
Handle KDE with specific arg/desktop file

fepitre:


etc/qvm-start-daemon-kde.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Name=Qubes Guid/Pacat
+Comment=Starts GUI/AUDIO daemon for Qubes VMs in KDE
+Icon=qubes
+Exec=qvm-start-daemon --all --watch --kde
+Terminal=false
+Type=Application
+OnlyShowIn=KDE
```
marmarek: This needs to end with a semicolon: https://developer.gnome.org/menu-spec/#desktop-entry-extensions-examples
And also there is missing newline.

etc/qvm-start-daemon.desktop
```diff
@@ -5,3 +5,4 @@ Icon=qubes
 Exec=qvm-start-daemon --all --watch
 Terminal=false
 Type=Application
+NotShowIn=KDE
```
marmarek: Same here.

marmarek (changes requested):

marmarek (approved):

PR #144, opened 2020-06-25
Generate qubes-guid options based on features

pwmarcz:
Allow configuring options per VM or globally per GuiVM. The
qvm-start-daemon program reads the options from VM features, and
generates a configuration file for qubes-guid.

Requires QubesOS/qubes-gui-daemon#47 (customizing the configuration
file).

### DIFF
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -1,4 +1,4 @@
-# -*- encoding: utf8 -*-
+# -*- encoding: utf-8 -*-
 #
 # The Qubes OS Project, http://www.qubes-os.org
 #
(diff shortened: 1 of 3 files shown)


qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -262,14 +365,32 @@ def common_guid_args(self, vm):
         if vm.features.check_with_template('rpc-clipboard', False):
             guid_cmd.extend(['-Q'])
 
+        guivm = self.app.domains[self.app.local_name]
```
marmarek: This should be rather `vm.guivm`. In practice it is the same value (gui-daemon is started only if vm.guivm is local), but it will be clearer this way.

marmarek (changes requested):
The temporary config should be also removed when not needed anymore. Add a `domain-stopped` event handler for that?

qubesadmin/tools/qvm_start_daemon.py
pwmarcz: Done.

pwmarcz:
>The temporary config should be also removed when not needed anymore. Add a domain-stopped event handler for that?

Done.

marmarek (approved):

PR #145, opened 2020-07-01
Support for qvm-template

WillyPillow:
Discussions: <https://groups.google.com/g/qubes-devel/c/PyJogqT1TUg>

Design doc: <https://gist.github.com/WillyPillow/b8a643ddbd9235a97bc187e6e44b16e4>

Major changes:

- Place appmenu whitelists in `qvm-features`
- Read `qvm-feature` and `qvm-pref` flags from `template.conf`. See document above for details
- Allow non-split root.img tars (may be removed in the future)
- Introduce `qvm-template` tool


qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+        for i in range(len(table[0])):
+            widths.append(max(len(s[i]) for s in table))
+        for row in sorted(table):
+            cols = ['{key:{width}s}'.format(
+                key=row[i], width=widths[i]) for i in range(len(row))]
+            print(' '.join(cols))
+
+def do_list(args, app):
+    # TODO: Check local template name actually matches to account for renames
+    # TODO: Also display repo like `dnf list`
```
marmarek: And also template state (for example if currently installed).

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+        for vm in app.domains:
+            if 'template-install-date' in vm.features:
+                version_str = build_version_str((
+                    vm.features['template-epoch'],
+                    vm.features['template-version'],
+                    vm.features['template-release']))
+                tpl_list.append((vm.name, version_str))
+
+    if args.available or args.all:
+        query_res = qrexec_repoquery(args, app)
```
marmarek: You can call `qrexec_repoquery` just once and then reuse result in each case below.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+            sys.stdout.buffer.write(c)
+            sys.stdout.flush()
+        if proc.wait() != 0:
+            return False
+        return True
+
+def build_version_str(evr):
+    return '%s:%s-%s' % evr
+
+def pretty_print_table(table):
```
marmarek: There is already `qubesadmin.tools.print_table()` for this.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+            stdout=f, encoding=None)
+        payload = qrexec_payload(args, app, spec)
+        proc.stdin.write(payload.encode('UTF-8'))
+        proc.stdin.close()
+        while True:
+            c = proc.stderr.read(1)
+            if not c:
+                break
+            # Write raw byte w/o decoding
+            sys.stdout.buffer.write(c)
```
marmarek: This can be bad - terminal escape sequences can be dangerous. At lease filter out ESC character (either drop, or replace with something innocent), but to be on the safe side, I would do the same with anything outside of ASCII (so char code >128).

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+
+def qrexec_download(args, app, spec, path):
+    with open(path, 'wb') as f:
+        proc = qrexec_popen(args, app, 'qubes.TemplateDownload',
+            stdout=f, encoding=None)
+        payload = qrexec_payload(args, app, spec)
+        proc.stdin.write(payload.encode('UTF-8'))
+        proc.stdin.close()
+        while True:
+            c = proc.stderr.read(1)
```
marmarek: Why doing it byte-by-byte?

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+            tpl.features['template-version'] = \
+                package_hdr[rpm.RPMTAG_VERSION]
+            tpl.features['template-release'] = \
+                package_hdr[rpm.RPMTAG_RELEASE]
+            tpl.features['template-install-date'] = \
+                str(datetime.datetime.today())
+            tpl.features['template-name'] = name
+            # TODO: Store source repo
+
+def qrexec_popen(args, app, service, stdout=subprocess.PIPE, encoding='UTF-8'):
```
marmarek: There is already `app.domains[...vmname...].run_service()` doing very similar thing.

Also, avoid decoding untrusted data (coming from another VM), at least not before some sanity checks, even if it's only UTF-8. Either use the simplest possible encoding (ascii), or drop automatic decoding entirely (preferred) - and then only decode parts that are actually needed (but still preferable with simple ascii).

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+def qrexec_payload(args, app, spec):
+    payload = ''
+    for r in args.enablerepo if args.enablerepo else []:
+        payload += '--enablerepo=%s\n' % r
+    for r in args.disablerepo if args.disablerepo else []:
+        payload += '--disablerepo=%s\n' % r
+    for r in args.repoid if args.repoid else []:
+        payload += '--repoid=%s\n' % r
+    payload += '--releasever=%s\n' % args.releasever
+    payload += spec + '\n'
```
marmarek, line 457: All of the above could use a check if no unexpected characters are included (spaces, newlines). Better report a clear error early, than some confusing message when the command was misinterpreted.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+                ' operations.)') % name, file=sys.stderr)
+            del dl_list_copy[name]
+        else:
+            version_str = build_version_str(ver)
+            target_file = \
+                '%s%s-%s.rpm' % (PACKAGE_NAME_PREFIX, name, version_str)
+            rpm_list.append(os.path.join(args.cachedir, target_file))
+    dl_list = dl_list_copy
+
+    download(args, app, path_override=args.cachedir, dl_list=dl_list)
```
marmarek: To avoid accidental use of the file before verification, consider downloading to a differently-named file (like, append `.unverified` suffix or similar) and only rename after verification.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+                print(('Template \'%s\' already installed, skipping...'
+                    ' (You may want to use the {reinstall,upgrade,downgrade}'
+                    ' operations.)') % name, file=sys.stderr)
+                continue
+
+            print('Installing template \'%s\'...' % name, file=sys.stderr)
+            extract_rpm(name, rpmfile, target)
+            cmdline = [
+                'qvm-template-postprocess',
+                '--keep-source',
```
marmarek: Why `--keep-source` (besides debugging)?

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,409 @@
...
+
+def get_package_hdr(path, transaction_set=None):
+    if transaction_set is None:
+        transaction_set = rpm.TransactionSet()
+    with open(path, 'rb') as f:
+        hdr = transaction_set.hdrFromFdno(f)
+        return hdr
+
+def extract_rpm(name, path, target):
+    with open(path, 'rb') as in_file:
```
marmarek: Why? You don't seem to use `in_file` here.

marmarek (changes requested):
Besides remarks below, take a look also at pylint output: https://travis-ci.com/github/QubesOS/qubes-core-admin-client/jobs/358989660#L350

Tests will need adjustments too, but that can be later step (when the actual tool will be done).

qubesadmin/tools/qvm_template.py
WillyPillow: Oops, that was indeed something I did for debugging. As TemporaryDirectory removes the directory, I did not notice this while testing it afterwards.

qubesadmin/tools/qvm_template.py
WillyPillow: This seems to be the remains of the rpm2archive version. Will remove it later.

qubesadmin/tools/qvm_template.py
WillyPillow: Ahh, I initially did notice the function in the docs, but did not realize that it accepts a kwargs argument (as I needed to redirect the output to a file).

As for decoding, the only thing that needs to be decoded should be the output of TemplateSearch, which *should* indeed only require ASCII.

qubesadmin/tools/qvm_template.py
WillyPillow: The reason for this and the byte-by-byte reads below are because I'm currently reusing curl's progress indicator. That being said, this is indeed an issue for security. Maybe I should either implement my own progress indicator or pull in pv/tqdm as a dependency?

qubesadmin/tools/qvm_template.py
WillyPillow: Shouldn't whitespace characters (besides newlines) work as intended as long as the quoting is done correctly in the qrexec service?

qubesadmin/tools/qvm_template.py
WillyPillow: Implemented a simple progress bar for the time being.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,407 @@
...
+        while proc.poll() == None:
+            width = shutil.get_terminal_size((80, 20)).columns
+            pct = '%5.f%% ' % math.floor(f.tell() / dlsize * 100)
+            bar_len = width - len(pct) - 2
+            num_hash = math.floor(f.tell() / dlsize * bar_len)
+            num_space = bar_len - num_hash
+            # Clear previous bar
+            print(u'\u001b[1000D', end='', file=sys.stderr)
+            print(pct + '[' + ('#' * num_hash) + (' ' * num_space) + ']',
+                end='', file=sys.stderr)
```
marmarek: Take a look at tqdm python module for progress bar. New (not released yet) pdf converter uses it: https://github.com/QubesOS/qubes-app-linux-pdf-converter/blob/master/qubespdfconverter/client.py

marmarek (changes requested):

qubesadmin/tools/qvm_template.py
marmarek: Yes, you're right.

qubesadmin/tools/qvm_template.py
WillyPillow: Ahh, so tqdm is actually available :) Mentioned it above but was not sure if it's okay to pull in a dependency for this.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,802 @@
...
+    stdout, stderr = proc.communicate(payload.encode('UTF-8'))
+    stdout = stdout.decode('ASCII')
+    if proc.wait() != 0:
+        for line in stderr.decode('ASCII').rstrip().split('\n'):
+            print('[Qrexec] %s' % line, file=sys.stderr)
+        raise ConnectionError("qrexec call 'qubes.TemplateSearch' failed.")
+    name_re = re.compile(r'^[A-Za-z0-9._+\-]*$')
+    evr_re = re.compile(r'^[A-Za-z0-9._+~]*$')
+    date_re = re.compile(r'^\d+-\d+-\d+ \d+:\d+$')
+    licence_re = re.compile(r'^[A-Za-z0-9._+\-()]*$')
```
marmarek: `-` must be the last in the [] set, otherwise it is interpreted as a chars range.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,802 @@
...
+            buildtime = datetime.datetime.strptime(buildtime, '%Y-%m-%d %H:%M')
+            # XXX: Perhaps whitelist licenses directly?
+            if not re.fullmatch(licence_re, licence):
+                raise ValueError
+            # Check name actually matches spec
+            if not is_match_spec(name, epoch, version, release, spec):
+                continue
+
+            result.append((name, epoch, version, release, reponame, dlsize,
+                buildtime, licence, url, summary, description))
```
marmarek: Perhaps use `collections.namedtuple()`? Will ease handling this many items.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,802 @@
...
+        _ = install_time # unused
+        #pylint: disable=unused-variable
+        name, epoch, version, release, reponame, dlsize, \
+            buildtime, licence, url, summary, description = data
+        version_str = build_version_str((epoch, version, release))
+        tpl_list.append((status, name, version_str, reponame))
+
+    def append_info(data, status, install_time=None):
+        name, epoch, version, release, reponame, dlsize, \
+            buildtime, licence, url, summary, description = data
```
marmarek: With namedtuple you could access `data.name` etc below, without this unpacking.

marmarek (changes requested):

qubesadmin/tools/qvm_template.py
WillyPillow: Both in my testing and from regex101, the backslash before the `-` seem to work as intended?

marmarek: Ah, indeed, I haven't noticed backslash.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,831 @@
...
+
+def query_local_evr(vm):
+    return (
+        vm.features['template-epoch'],
+        vm.features['template-version'],
+        vm.features['template-release'])
+
+def is_managed_template(vm):
+    return 'template-name' in vm.features \
+        and vm.name == vm.features['template-name']
```
marmarek: This can be simplified as `vm.features.get('template-name', None) == vm.name`. A not so obvious side effect is that, it avoids listing vm features, which may save on required Admin API permissions.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,831 @@
...
+    if spec == '---':
+        parser.error("Malformed template name: argument should not be '---'.")
+
+    def check_newline(string, name):
+        if '\n' in string:
+            parser.error(f"Malformed {name}:" +
+                " argument should not contain '\\n'.")
+
+    payload = ''
+    for repo in args.enablerepo if args.enablerepo else []:
```
marmarek: You can add `default=[]` to the `parser.add_argument()` call to simplify this (and other instances below).

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,831 @@
...
+    payload += spec + '\n'
+    payload += '---\n'
+    for path in args.repo_files:
+        with open(path, 'r') as fd:
+            payload += fd.read() + '\n'
+    return payload
+
+def qrexec_repoquery(args, app, spec='*', refresh=False):
+    proc = qrexec_popen(args, app, 'qubes.TemplateSearch')
+    payload = qrexec_payload(args, app, spec, refresh)
```
marmarek, line 483: Better first prepare payload and only then open the qrexec call - this way if the `qrexec_payload` raises an exception, you saves the qrexec call.

marmarek (changes requested):

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,1269 @@
...
+            if reponame != '@commandline':
+                path = rpmfile + UNVERIFIED_SUFFIX
+            else:
+                path = rpmfile
+
+            package_hdr = verify_rpm(path, args.nogpgcheck, transaction_set)
+            if not package_hdr:
+                parser.error('Package \'%s\' verification failed.' % rpmfile)
+
+            if reponame != '@commandline':
```
marmarek: `path != rpmfile` would be clearer.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,1269 @@
...
+                os.rename(path, rpmfile)
+
+            package_name = package_hdr[rpm.RPMTAG_NAME]
+            if not package_name.startswith(PACKAGE_NAME_PREFIX):
+                parser.error(
+                    'Illegal package name for package \'%s\'.' % rpmfile)
```
marmarek: Shouldn't this be before rename?

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,1269 @@
...
+                override_tpls.append(name)
+
+            print('This will override changes made in the following VMs:',
+                file=sys.stderr)
+            for name in override_tpls:
+                print('  %s' % name, file=sys.stderr)
+            confirm = ''
+            while confirm != 'y':
+                confirm = input('Are you sure? [y/N] ').lower()
+                if confirm == 'n':
```
marmarek: This definitely should be `confirm != 'y'`.

marmarek (changes requested):

qubesadmin/tools/qvm_template.py
WillyPillow: The idea was to ask again if the input is neither 'y' or 'n', hence the while loop. AFAIK some CLI tools do this.

(I'm also okay with making it ask only once for simplicity though.)

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,1387 @@
...
+        remove_dummy = purge
+        # Create dummy template; handle name collisions
+        orig_dummy = dummy
+        cnt = 1
+        while dummy in app.domains \
+                and not app.domains[dummy].features.get('template-dummy', 0):
+            dummy = '%s-%d' % (orig_dummy, cnt)
+            cnt += 1
+        if dummy not in app.domains:
+            dummy_vm = app.add_new_vm('TemplateVM', dummy, 'red')
```
marmarek, line 1386: Above you look at `template-dummy` feature, shouldn't you also add it here?

marmarek (changes requested):

marmarek:
Generally the changes looks good now. Take a look at test failure (travis logs) - you need to adjust few tests for new menu-items feature.

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,1396 @@
...
+UNVERIFIED_SUFFIX = '.unverified'
+LOCK_FILE = '/var/tmp/qvm-template.lck'
+
+def qubes_release() -> str:
+    """Return the Qubes release."""
+    if os.path.exists('/usr/share/qubes/marker-vm'):
+        with open('/usr/share/qubes/marker-vm', 'r') as fd:
+            # Get last line (in the format `x.x`)
+            return fd.readlines()[-1].strip()
+    return subprocess.check_output(['lsb_release', '-sr'],
```
marmarta: Either this is missing a dependency (there's no lsb_release in dom0 by default), or maybe you could just directly read /etc/os-release ? it seems a bit overkill to use a separate process and program to parse a simple file - especially as redhat-lsb-core has several otherwise useless dependencies.

WillyPillow: Replaced with reading /etc/os-release.

(Did not realize that there's no lsb_release by default -- upon inspection it seemed to be pulled in by `tlp` on my box.)

marmarek:
> The diff coverage is `2.36%`.

This could use some tests...

WillyPillow:
> > The diff coverage is `2.36%`.
> 
> This could use some tests...

Yes, I'm working on it right now :)

Other than that, CI seems to pass now.

That being said, I also had to ignore the modules `rpm` and `dnf` for `pylint` as they aren't on PyPI. Is there a better way of approaching this?
(For `rpm`, it should possible to install it as a system package. OTOH, `dnf` doesn't seem to be packaged on Ubuntu.)

WillyPillow:
Side note: I'm unsure why `policy/qubesos/code-signing` gives `Unable to verify (no valid key found)`.

[My (sub)key](http://hkps.pool.sks-keyservers.net/pks/lookup?op=vindex&fingerprint=on&search=0x1C16C70EE7C31C84) does seem to be on `hkps.pool.sks-keyservers.net`.

IIRC, the SKS keyservers are kind of unstable recently. Don't know if it's because of that. (Running `--recv-keys` on my end also seems to fail most of the time.)

marmarek:
> IIRC, the SKS keyservers are kind of unstable recently. Don't know if it's because of that.

Yes, that's exactly it...


> That being said, I also had to ignore the modules `rpm` and `dnf` for `pylint` as they aren't on PyPI. Is there a better way of approaching this?

You'll face this issue for tests too. As a workaround you can create stubs in test-packages directory. For tests, you'd most likely mock them anyway.

WillyPillow:
CI is failing because of timezones. Will fix it later.

WillyPillow:
It appears that `datetime.fromisoformat` does not exist in Python 3.6...

Considering <https://groups.google.com/g/qubes-devel/c/f7YcmR6Irus/m/dEgLk92nAgAJ>, should I, e.g., store instead the unix timestamp in feature tags, or just give up support for 3.6?

(On further reflection, it's probably simpler to stick to a fixed date format like `YYYY-MM-DD HH:mm:ss`. Also easier to define precisely in the docs compared to the current solution.)

marmarek:
> (On further reflection, it's probably simpler to stick to a fixed date format like `YYYY-MM-DD HH:mm:ss`. Also easier to define precisely in the docs compared to the current solution.)

Yes, this is fine. But please either include also timezone info in the format, or always store in UTC.

marmarek:
The qvm-template tool itself I think looks pretty good already. But take a look at the [test coverage report](https://codecov.io/gh/QubesOS/qubes-core-admin-client/pull/145/diff?src=pr&el=tree#diff-cXViZXNhZG1pbi90b29scy9xdm1fdGVtcGxhdGUucHk=) - some functions are missing tests. The trivial ones are that important, but for example get_dl_list or search could use some tests.

WillyPillow:
> Yes, this is fine. But please either include also timezone info in the format, or always store in UTC.

Yes, I think the times are all in UTC now.

(The times returned by qubes.TemplateSearch are also UTC, cf. <https://github.com/rpm-software-management/dnf/blob/master/dnf/cli/commands/repoquery.py#L711>.)

> The qvm-template tool itself I think looks pretty good already. But take a look at the [test coverage report](https://codecov.io/gh/QubesOS/qubes-core-admin-client/pull/145/diff?src=pr&el=tree#diff-cXViZXNhZG1pbi90b29scy9xdm1fdGVtcGxhdGUucHk=) - some functions are missing tests. The trivial ones are that important, but for example get_dl_list or search could use some tests.

I see. I was planning to do it in a bottom-up order, hence the current set of tests. Regardless, the basic functions should be mostly done, and right now I'm working on the more complicated operations like the ones you mentioned.

marmarek:
I reviewed the updated version and I see you also found some minor bugs while writing tests :) 
Looking at test coverage, those are non-trivial parts not covered with tests:
 - "download" function (especially adding/dropping .rpm suffix)
 - install with override_existing=True
 - "remove" function
 - qvm-template-postprocess - processing template.conf

marmarek:
Let me know when you think it is ready for merge and also remove "WIP" tag from the title.

marmarek:
PipelineRetry

WillyPillow:
Thanks for the commits! Apologies for the delay -- honestly I just took up too much work this semester, got pretty burnt out in the end, and had to push things back a bit :(

qubesadmin/tools/qvm_template.py
```diff
@@ -0,0 +1,1546 @@
...
+        remove_dummy = purge
+        # Create dummy template; handle name collisions
+        orig_dummy = dummy
+        cnt = 1
+        while dummy in app.domains \
+                and app.domains[dummy].features.get(
+                    'template-dummy', '0') != '1':
+            dummy = '%s-%d' % (orig_dummy, cnt)
+            cnt += 1
+        if dummy not in app.domains:
```
WillyPillow, line 1382: Is there a preferred way to mock constructs like this?

Existing tests that create VMs seem to just put the created VM in `admin.vm.List`, which interferes with the `if` here.

WillyPillow (commented):
Soon after posting this I realized that I already _had_ something similar (`add_new_vm_side_effect`). That was pretty silly...
Edit: Accidentally deleted the comment that this was referring to when I was trying to close the review :(

marmarek:
You can assign a list of results to `self.expected_calls[...]` - they will be returned in order for subsequent calls.

marmarek (approved):
The code looks good now. There is a missing test for repolist, but it isn't very complex function and also not a critical functionality, so I'm ok with merging it as is.

Also I've been testing this whole feature manually for some time already and it works fine!

WillyPillow:
Thanks. IIRC, the main reason I skipped the tests for `repolist` is that the dnf dependency isn't available everywhere and writing a stub for it doesn't seem trivial.

Again, I really appreciate the help along the way!

PR #147, opened 2020-07-05
Removed obsolete references to vm.icon_path

marmarta:
references QubesOS/qubes-issues#5934
needs https://github.com/QubesOS/qubes-core-admin/pull/355


marmarek:
While VM icon symlink will be gone in Qubes 4.1, it is present in backups created with older versions. Lets keep it in tests too, to verify if backup restoring code correctly handle older backups.

PR #148, opened 2020-07-09
Added better __eq__ method to Label class

### LINKED ISSUE #5944: Class Label in core-admin-client has no __eq__ method
**Qubes OS version**
R4.1

Class Label in core-admin-client has no __eq__ method, while the corresponding class in core-admin does.

marmarta:
Labels with the same name should not be distinguishable.

fixes QubesOS/qubes-issues#5944

### DIFF
--- a/qubesadmin/label.py
+++ b/qubesadmin/label.py
@@ -75,3 +75,8 @@ class Label(object):
 
     def __str__(self):
         return self._name
+
+    def __eq__(self, other):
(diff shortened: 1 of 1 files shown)


marmarta:
@marmarek , do we really need tests for this __eq__ method copy-pasted from core-admin?

marmarek:
Not really.

marmarek (approved):

PR #149, opened 2020-07-13
Add admin.vm.volume.Clear call (QubesOS/qubes-issues#5946)

### LINKED ISSUE #5946: Function "reset_private_image" in qvm_template_postprocess fails with "not enough data"
**Qubes OS version**

R4.1

`qvm-template-postprocess`

**Brief summary**

 [`reset_private_img`](https://github.com/QubesOS/qubes-core-admin-client/blob/master/qubesadmin/tools/qvm_template_postprocess.py#L125) raises the following error:

```
qubesadmin.exc.QubesException: Data import failed: not enough data (copied 0 bytes, expected 2147483648 bytes)
```

**To Reproduce**

Reinstall an already-installed template RPM.

**Expected behavior**

Private volume is cleared.

**Actual behavior**

The aforementioned error occurs.

**Additional context**

Ran into this issue while testing the the {reinstall,{down,up}grade} operations of `qvm-template`  (QubesOS/qubes-core-admin-client#145).

Upon further inspection, it seems that the `admin.vm.volume.Import` call requires the payload stream to be of the same size as the volume, causing the error. (Commit that introduced the check: https://github.com/QubesOS/qubes-core-admin/commit/e9b97e42b1dffe7b1d7317d73fcfa912f724d016)

**Solutions you've tried**

I tried replacing the import call with resizing the volume to zero then back to the old size. However, the relevant qrexec call does not seem to support the shrinking of volumes.

It seems sufficient to simply treat an input length of zero as a special case in `admin.vm.volume.Import`, though I am not sure if this is an ideal solution.

- #5192
- https://github.com/QubesOS/qubes-core-admin/commit/e9b97e42b1dffe7b1d7317d73fcfa912f724d016

None.

WillyPillow:
Resolves QubesOS/qubes-issues#5946.

Qrexec call implementation: https://github.com/QubesOS/qubes-core-admin/pull/359

### DIFF
--- a/qubesadmin/storage.py
+++ b/qubesadmin/storage.py
@@ -236,6 +236,10 @@ class Volume(object):
             'ImportWithSize', payload=size_line.encode(),
             payload_stream=stream)
 
+    def clear_data(self):
+        ''' Clear existing volume content. '''
+        self._qubesd_call('Clear')
+
     def clone(self, source):
         ''' Clone data from sane volume of another VM.
 
--- a/qubesadmin/tools/qvm_template_postprocess.py
+++ b/qubesadmin/tools/qvm_template_postprocess.py
@@ -124,8 +124,7 @@ def import_root_img(vm, source_dir):
 
 def reset_private_img(vm):
     '''Clear private volume'''
-    with open('/dev/null', 'rb') as null:
-        vm.volumes['private'].import_data(stream=null)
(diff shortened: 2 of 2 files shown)


marmarek (approved):

marmarek:
This needs also an update to relevant test, will do when merging.

PR #152, opened 2020-07-21
Add a lightdm session for sys-gui

pwmarcz:
Starts just sys-gui VM and qvm-start-daemon in a special mode.

See QubesOS/qubes-issues#5662.

This seems to need `allow_fullscreen`. I'd rather not add it to `sys-gui` because then it's becomes problematic to test it when running a normal session. Added an override: `qvm-start-daemon --allow-fullscreen`.

Remaining issues:

* Selecting "Logout" in `sys-gui` seems to shut down the Xephyr window but not the VM. Should it work that way? Should we detect that the GUI agent inside has died and then exit dom0 session, or rather make the VM shut down?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,11 @@ install:
 	install -d $(DESTDIR)/etc/xdg/autostart
 	install -m 0644 etc/qvm-start-daemon.desktop $(DESTDIR)/etc/xdg/autostart/
 	install -m 0644 etc/qvm-start-daemon-kde.desktop $(DESTDIR)/etc/xdg/autostart/
+	install -d $(DESTDIR)/usr/share/xsessions/
+	install -m 0644 xsessions/sys-gui.desktop $(DESTDIR)/usr/share/xsessions/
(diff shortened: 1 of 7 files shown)


pwmarcz:
>This seems to need allow_fullscreen. I'd rather not add it to sys-gui because then it's becomes problematic to test it when running a normal session. Added an override: qvm-start-daemon --allow-fullscreen.

After discussing: remove --allow-fullscreen and the permission for `sys-gui` globally in Salt configuration.

pwmarcz:
Salt should create a xsession file alongside the GuiVM, so that if you create several GuiVMs, it will create several session files.

pwmarcz:
* qvm-start-daemon should exit when GUI daemon exits (and run it in foreground)
* GUI daemon should have a `--no-restart` option

pwmarcz:
It might be better to modify qubes-guid's `restart_guid` function.

pwmarcz:
That will need a new branch, I'll close this one.

PR #161, opened 2020-09-08
Added supported-service keyboard-layout to allow for better working Qube Manager

marmarta:
references QubesOS/qubes-issues#6030

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -25,6 +25,9 @@ override_dh_auto_install:
 	install -d debian/qubes-core-admin-client/usr/share/man/man1
 	install -m 0644 build/man/* \
 		debian/qubes-core-admin-client/usr/share/man/man1/
+	install -d debian/qubes-core-admin-client/etc/qubes/post-install.d
+	install -m 0755 scripts/30-keyboard-layout-service.sh \
+		debian/qubes-core-admin-client/etc/qubes/post-install.d/30-keyboard-layout-service.sh
 
 override_dh_clean:
 	dh_clean -O--buildsystem=pybuild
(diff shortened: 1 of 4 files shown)


scripts/30-keyboard-layout-service.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+qvm-features-request supported-service.keyboard-layout=1
```
marmarek: `supported-service.` prefix is reserved for things controlled with `qvm-service` (aka `service.`-prefixed features).
Normally announcing supported features requires appropriate part in core-admin (like [here](https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/ext/core_features.py#L28-L64)). Since this isn't the first time we need a simple feature support to announce that isn't strictly `qvm-service`, perhaps we need a generic mechanism for this? Maybe similar thing with `supported-feature.` prefix?

marmarek (changes requested):

scripts/30-keyboard-layout-service.sh
marmarta: Sure, that sounds perfectly fine.

debian/rules
```diff
@@ -25,6 +25,9 @@ override_dh_auto_install:
 	install -d debian/qubes-core-admin-client/usr/share/man/man1
 	install -m 0644 build/man/* \
 		debian/qubes-core-admin-client/usr/share/man/man1/
+	install -d $(DESTDIR)/etc/qubes/post-install.d
+	install -m 0755 scripts/30-keyboard-layout-service.sh \
+		$(DESTDIR)/etc/qubes/post-install.d/30-keyboard-layout-service.sh
```
marmarek: See other rules (and also travis log), this should place files in `debian/qubes-core-admin-client`, not `$(DESTDIR)`. Or set `DESTDIR` at the beginning of this file, like it's done in other packages.

marmarek (changes requested):

marmarek (approved):

PR #172, opened 2021-06-27
Fix guivm

fepitre:


doc/manpages/qvm-service.rst
```diff
@@ -137,9 +137,32 @@ disable-dns-server
     Enabling this service will result in an empty /etc/resolv.conf.
     The functionality is implemented in /usr/lib/qubes/setup-ip.
 
+lightdm:
+    Default: enabled in GPU and VNC GuiVM
```
marmarek: Default value is rather about what happens if this service is not set (neither enabled nor disabled). So, in all those cases the default is "disabled"

doc/manpages/qvm-service.rst
```diff
@@ -137,9 +137,32 @@ disable-dns-server
...
+
+    Start `lightdm` on boot which prevent `qubes-gui-agent` to start.
```
marmarek: Those are rather two separate effects of this service: Start `lightdm` and avoid starting `qubes-gui-agent`.

marmarek (changes requested):

marmarek (approved):

PR #175, opened 2021-07-18
Use fcntl locking for the PID file

### LINKED ISSUE #6780: qvm-start-daemon does not relaunch on X restart
### Qubes OS version

R4.1

### Which [testing](https://www.qubes-os.org/doc/testing/) repositories are you using, if any?

All are enabled in dom0 (where the bug is)

### Affected component(s) or functionality

GUI and audio daemon lifecycle management

-----
### Brief summary

`qvm-start-daemon` uses a lock file (`$XDG_RUNTIME_DIR/qvm-start-daemon.pid`) to guard against multiple instances running.  However, an X restart (such as by `systemctl restart lightdm.service`) does not delete the lock file, so `qvm-start-daemon` fails to start the second time around.

### How reproducible

100%

### Steps to reproduce

1. Restart X with `systemctl restart lightdm.service`
2. Wait for the GUI daemon to come back up.

### Expected behavior

The GUI daemon comes back up normally.

### Actual behavior

The GUI daemon does not come back up.

### Solutions you've tried

Manually deleting the file helps.  A better solution would be to overwrite the file unless another process has a lock on it.

### Additional context

-----
### Relevant [documentation](https://www.qubes-os.org/doc/) you've consulted

None

### Related, [non-duplicate](https://www.qubes-os.org/doc/issue-tracking/#new-issues-should-not-be-duplicates-of-existing-issues) issues

#4729

DemiMarie:
This handles the case where qvm-start-daemon gets killed and leaves a
stale PID file.  Previously, qvm-start-daemon would refuse to start.
Now, it successfully locks the file and starts normally.

Fixes QubesOS/qubes-issues#6780.

### DIFF
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -23,6 +23,7 @@
 import os
 import signal
 import subprocess
+import fcntl
 import asyncio
 import re
 import functools
@@ -30,7 +31,6 @@ import sys
 import xcffib
 import xcffib.xproto  # pylint: disable=unused-import
 
-import daemon.pidfile
 import qubesadmin
 import qubesadmin.events
 import qubesadmin.exc
@@ -750,7 +750,24 @@ def main(args=None):
         kde=args.kde)
 
     if args.watch:
-        with daemon.pidfile.TimeoutPIDLockFile(args.pidfile):
+        fd = os.open(args.pidfile,
+                     os.O_RDWR | os.O_CREAT | os.O_CLOEXEC,
+                     0o600)
+        with os.fdopen(fd, 'r+') as lock_f:
+            try:
+                fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except BlockingIOError:
+                try:
+                    pid = int(lock_f.read().strip())
+                except ValueError:
+                    pid = 'unknown'
+                print('Another GUI daemon process (with PID {}) is already '
+                      'running'.format(pid),
+                      file=sys.stderr)
+                sys.exit(1)
+            print(os.getpid(), file=lock_f)
+            lock_f.flush()
+            lock_f.truncate()
             loop = asyncio.get_event_loop()
             # pylint: disable=no-member
             events = qubesadmin.events.EventsDispatcher(args.app)


qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -750,7 +750,14 @@ def main(args=None):
         kde=args.kde)
 
     if args.watch:
-        with daemon.pidfile.TimeoutPIDLockFile(args.pidfile):
+        fd = os.open(args.pidfile,
+                     os.O_WRONLY | os.O_CREAT | os.O_CLOEXEC,
+                     0o600)
+        with os.fdopen(fd, 'w') as lock_f:
+            fcntl.flock(fd, fcntl.LOCK_EX)
```
marmarek: `fcntl.LOCK_EX | fcntl.LOCK_NB`, and catch the possible exception to convert into appropriate error message?

marmarek (changes requested):

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -750,7 +750,24 @@ def main(args=None):
...
+        fd = os.open(args.pidfile,
+                     os.O_WRONLY | os.O_CREAT | os.O_CLOEXEC,
+                     0o600)
+        with os.fdopen(fd, 'rw') as lock_f:
+            try:
+                fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except BlockingIOError:
+                try:
+                    pid = int(lock_f.read().strip())
+                except Exception:
```
marmarek: `ValueError`? IMO better not silently ignore `IOError` here, as it may suggest something really wrong happening.

DemiMarie: Ah, good point.

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -750,7 +750,24 @@ def main(args=None):
         kde=args.kde)
 
     if args.watch:
-        with daemon.pidfile.TimeoutPIDLockFile(args.pidfile):
+        fd = os.open(args.pidfile,
+                     os.O_WRONLY | os.O_CREAT | os.O_CLOEXEC,
+                     0o600)
+        with os.fdopen(fd, 'rw') as lock_f:
```
marmarek: `rw` is invalid. I think you need `r+` or something like this.

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -750,7 +750,24 @@ def main(args=None):
         kde=args.kde)
 
     if args.watch:
-        with daemon.pidfile.TimeoutPIDLockFile(args.pidfile):
+        fd = os.open(args.pidfile,
+                     os.O_WRONLY | os.O_CREAT | os.O_CLOEXEC,
```
marmarek: When you use `O_WRONLY` here, reading the PID below fails. `O_RDWR`?

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -750,7 +750,24 @@ def main(args=None):
         kde=args.kde)
 
     if args.watch:
-        with daemon.pidfile.TimeoutPIDLockFile(args.pidfile):
+        fd = os.open(args.pidfile,
+                     os.O_O_RDWR | os.O_CREAT | os.O_CLOEXEC,
```
marmarek:
```suggestion
                     os.O_RDWR | os.O_CREAT | os.O_CLOEXEC,
```

marmarek (approved):

PR #182, opened 2021-09-09
Better error reporting if backup file not found

### LINKED ISSUE #3446: qvm-backup-restore indicates header error when passing inexisting file
#### Qubes OS version:

Q4RC3

` $ qvm-backup-restore -d existingvm /some/invalid/path` 

### Expected behavior:

Indicates an error such as "file does not exist" and no prompt of the password

### Actual behavior:

Indicates an error regarding the invalid header of the file and prompts for a password

### General notes:

Not really a bug but _very_ confusing

marmarta:
qvm-backup-restore will now show a File Not Found message
if specified backup file is not found.

fixes QubesOS/qubes-issues#3446

### DIFF
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -1195,15 +1195,16 @@ class BackupRestore(object):
         retrieve_proc_returncode = retrieve_proc.wait()
         if retrieve_proc in self.processes_to_kill_on_cancel:
             self.processes_to_kill_on_cancel.remove(retrieve_proc)
-        extract_stderr = error_pipe.read(MAX_STDERR_BYTES)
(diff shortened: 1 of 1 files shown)


rustybird:
This checks if a backup file exists in dom0, and will always fail with `-d`/`--dest-vm` (e.g. in the ticket).

Looks like the header retrieval error message should include stderr from the `qubes.Restore` RPC script. Although this stderr currently isn't very helpful anyway in case of a missing file:

    Starting Restorecopy
    Arguments: mybackupfile
    Paths: backup-header backup-header.hmac
    Checking if arguments is matching a command
    Invalid command mybackupfile

PR #195, opened 2021-11-10
fix stubdomain gui starting when 'gui' is empty

### LINKED ISSUE #5739: Win10 VM won't start properly after installing qubes windows tools
**Qubes OS version**
4.0.1

StandaloneVM start

**Brief summary**
After installing qubes-windows-tools on a Windows 10 VM, I used it for a while, everything worked. Now when I go to launch it again, instead of having simply "qubes settings" and "start" like before, I have all the windows apps to start. If I click on one of them, the VM starts but does not show any screen, it just shows up as started. There's also not a "start" option anymore

**To Reproduce**
_Steps to reproduce the behavior:_
Install qubes-windows-tools
shutdown
try to start vm again

jevank:
'gui' feature advertised with ~None~ empty which is None, but 'not None' is True here

This fixes QubesOS/qubes-issues#5739

### DIFF
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -601,7 +601,8 @@ class DAEMONLauncher:
         try:
             if getattr(vm, 'guivm', None) != vm.app.local_name:
                 return
-            if not vm.features.check_with_template('gui', True):
+            if not vm.features.check_with_template('gui', True) and \
+                    not vm.features.check_with_template('gui-emulated', True):
                 return
             if vm.virt_mode == 'hvm' and \
                     kwargs.get('start_guid', 'True') == 'True':


jevank:
Perhaps description is inaccurate. Here should be positive 'gui' condition check, not negative.

marmarek:
Setting `gui` feature to false (empty) is intended to disable any kind of GUI for the VM, both gui-agent inside VM and stubdomain one. Maybe you meant `gui-emulated` instead? Take a look at description at https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-features.html#gui

I think the issue may be mishandling `gui-emulated` here (it looks like if `gui=`, then it ignores `gui-emulated=1`).

jevank:
Empty is no gui at all (this is new for me), zero means emulated gui?

So qwt should advertise zero instead of empty?

marmarek:
There are two settings: `gui` and `gui-emulated`. The intended case is:
 - `gui=` and `gui-emulated=` -> no GUI at all
 - `gui=` (or missing) and `gui-emulated=1` -> just stubdomain GUI
 - `gui=1` and `gui-emulated=` (or missing) -> just GUI agent from within the VM
 - `gui=1` and `gui-emulated=1` - both
 - no `gui`, or `gui-emulated` set at all (which is different than empty) - same as `gui-emulated=1`

jevank:
Got it, thank you.

> I think the issue may be mishandling gui-emulated here (it looks like if gui=, the it ignores gui-emulated=1).

So, that is correct. I will fix the PR.

marmarek:
pylint complains:
```
************* Module qubesadmin.tools.qvm_start_daemon
qubesadmin/tools/qvm_start_daemon.py:605:0: C0301: Line too long (82/80) (line-too-long)
```

marmarek (approved):

PR #196, opened 2021-11-22
Use `%_pkgverify_level all` when checking package signatures

DemiMarie:
This is what DNF uses, and is what RPM requires users to use for best
security.  The current behavior is believed to not be exploitable, as
packages that could exploit it cannot be created without a trusted
secret key.


DemiMarie:
Looks like a Debian mirror issue.  Best option I can think of is to retry failed update installs.

PR #203, opened 2021-12-18
qvm-run: Add command block to example commands

siliconninja:
Makes it so the online Sphinx docs don't change them to en-dashes by mistake, e.g. https://dev.qubes-os.org/projects/core-admin-client/en/stable/manpages/qvm-run.html

I didn't see a coding style convention, so I thought it would be appropriate to add this to help fix the web docs.

Additionally, I have a question. I wasn't able to get the first example command working with `--` (the second one works fine). I am on Qubes R4.0.4. I have tried a couple of variations of the `ls` command provided. Any suggestions?

In my terminal in dom0, I get:
```bash
$ qvm-run --pass-io personal -- ls -a
usage: qvm-run [--verbose] [--quiet] [--help] [--user USER] [--autostart]
...
qvm-run: error: unrecognized arguments: -a
```
I have also tried:
`$ qvm-run --pass-io personal ls -- -a`: `qvm-run: error: unrecognized arguments: -a`
`$ qvm-run --pass-io personal -- ls -- -a`: `qvm-run: error: unrecognized arguments: -- -a`

However, `qvm-run --pass-io personal -- ls` works.

[I read about the double-dash online and it seems to depend on the command being run. It seems this isn't a problem for `ls` though.](https://serverfault.com/a/388035)

### DIFF
--- a/doc/manpages/qvm-run.rst
+++ b/doc/manpages/qvm-run.rst
@@ -16,14 +16,14 @@ Execution mode
 When executing a command, it is recommended to pass the arguments directly to
 `qvm-run`, for example:
 
-    qvm-run --pass-io personal -- ls -a
+    ``qvm-run --pass-io personal -- ls -a``
 
 (Note the `--` used to ensure that `-a` is not treated as option of `qvm-run`).
 
 If no arguments are specified, *COMMAND* will be interpreted as a shell
 command, for example:
 
-    qvm-run --pass-io personal 'ls -a'
+    ``qvm-run --pass-io personal 'ls -a'``
 
 This is more flexible but also less safe, because care must be taken to quote
 or escape special characters. Use the :option:`--no-shell` option to ensure


siliconninja:
Re-signing the branch commits with qubes key, sorry about that

siliconninja:
It seems that according to the documentation, [I should submit my key to `pool.sks-keyservers.net`](https://www.qubes-os.org/doc/code-signing/#upload-the-key). However, [that service is no longer in use](https://stackoverflow.com/a/68328830) (visiting the website now gives an SSL error).

I have sent it to both `keys.openpgp.org` and `pgp.mit.edu` (which should send it to a bunch of other servers, from what I've heard about PGP), but `policy/qubesos/code-signing` still fails.

rustybird:
> I wasn't able to get the first example command working with `--` (the second one works fine). I am on Qubes R4.0.4.

The first form is a new feature in R4.1.

siliconninja:
I see, thanks for answering!

I recently found in the [documentation style guide](https://www.qubes-os.org/doc/documentation-style-guide/#writing-command-line-examples) and in a [recent commit using the style guide for shell commands](https://github.com/QubesOS/qubes-doc/commit/29ed6e6e9b442e598ecf8475fd5863051dfaf4fd#diff-135ef8d05daf178d31bb470951f69b592b4a121843fa54e8abfeebd379d26a97R224) that I should be using multi-line code blocks and `shell_session`.

I will update the PR in my next commit, and then I think it will be ready for review.

siliconninja:
That documentation is only for markdown documents, not for manpages. Sorry for my confusion! I think it's ready for review as-is.

marmarek (approved):

PR #205, opened 2022-03-05
Use python's standard script creation

iacore:
learned from https://github.com/QubesOS/qubes-desktop-linux-menu/blob/master/setup.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ pkgs/
 .coverage
 __pycache__
 debian/changelog.*
+qubesadmin.egg-info/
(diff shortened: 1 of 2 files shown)


marmarek:
It is very much intentionally done the way it is. See https://github.com/QubesOS/qubes-core-admin-client/commit/b731ef388571c30e1d2383bdf6b287ed7c6dbe42

iacore:
What if I don't want to install this is `/usr/bin`, but as a local package?

marmarek:
`self.root` (`--root` argument) is respected. If you really want, you can adjust the generator to respect other settings. But do not revert to the old method.

PR #214, opened 2022-05-21
Fix misparsing of literal `--` by `qvm-run`

DemiMarie:
`qvm-run` misparsed a COMMAND argument that is literally `--`, causing
it to crash with an uncaught exception.  Additionally, the first `--` in
the arguments is swallowed, leading to option injection.

### DIFF
--- a/qubesadmin/tests/tools/qvm_run.py
+++ b/qubesadmin/tests/tools/qvm_run.py
@@ -769,7 +769,7 @@ class TC_00_qvm_run(qubesadmin.tests.QubesTestCase):
              'vmexec', None)] = \
             b'0\x001'
         ret = qubesadmin.tools.qvm_run.main(
-            ['--no-gui', '--dispvm', 'test-vm', 'command', '--', 'arg'],
+            ['--no-gui', '--dispvm=test-vm', 'command', '--', 'arg'],
             app=self.app)
         self.assertEqual(ret, 0)
         self.assertEqual(self.app.service_calls, [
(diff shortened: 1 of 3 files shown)


qubesadmin/tests/tools/qvm_run.py
```diff
@@ -643,3 +643,179 @@ def test_021_paused_vm(self):
...
+        ])
+        self.assertAllCalled()
+
+    def test_027_no_shell_dispvm(self):
+        self.app.expected_calls[
+            ('$dispvm:test-vm', 'admin.vm.feature.CheckWithTemplate',
+             'vmexec', None)] = \
+            b'0\x001'
+        ret = qubesadmin.tools.qvm_run.main(
+            ['--no-gui', '--dispvm', 'test-vm', 'command', '--', 'arg'],
```
marmarek: Written this way, `test-vm` should be really interpreted as a command (and no explicit base VM for DispVM - use default). If you want explicit base VM, then use `--dispvm=test-vm`. I'm not sure if argparse can make this distinction...

marmarek (changes requested):

marmarek:
PipelineRetry

qubesadmin/tests/tools/qvm_run.py
DemiMarie: Ugh, you are correct.  I should report these bugs to Python upstream, but honestly I think argparse doesn’t provide sufficient low-level control for what we need.  I would also like to raise a deprecation warning if no `--` is used, and I don’t think argparse supports that either.  I suggest switching to getopt or just parsing options manually.

For now, I changed these tests to not trigger the bug, and added a new test marked as “expected failure”.

qubesadmin/tools/qvm_run.py
```diff
@@ -223,6 +232,15 @@ def run_command_single(args, vm):
...
     args = parser.parse_args(args, app=app)
+    # pylint: disable=unidiomatic-typecheck
+    if type(args.cmd) is not str:
+        # pylint: disable=unidiomatic-typecheck
+        if type(args.cmd) is list and not args.cmd:
+            # Work around an argparse bug: if COMMAND is literally "--", it is
+            # misparsed as an empty list!
+            args.cmd = '--'
+        else:
+            raise AssertionError('args.cmd misparsed somehow?  (this is a bug)')
```
DemiMarie, line 243: Of course these lines are not covered by tests.  They are supposed to be unreachable, after all.

Is there a way to indicate this to the coverage code?

marmarek (approved):

PR #215, opened 2022-05-21
Enable restoring volumes with size smaller than default

a-barinov:


a-barinov:
This commit is harmless standalone as backend will not downsize volumes without additional commits.

marmarek:
Thanks for the contribution!
You need to update tests - see CI log at https://gitlab.com/QubesOS/qubes-core-admin-client/-/jobs/2488973220

a-barinov:
I miss he days when patches were submitted by email... Anyway, [this PR](https://github.com/QubesOS/qubes-core-admin/pull/479) needs to come in first, then I'll do a new PR for restore.py code change

marmarek:
> Anyway, [this PR](https://github.com/QubesOS/qubes-core-admin/pull/479) needs to come in first,

This isn't strictly true. `import_with_size()` allows importing volume even if given size is smaller than current volume size - there is no need to call resize separately. What you need to do, is to adjust `_handle_volume_size` to save the size somewhere, to be used as `import_with_size()` argument, instead of calling resize. Alternatively, refactor `_handle_volume_size` and `_handle_volume_data` into one function with two parameters, and adjust all places where it is called.

PR #220, opened 2022-08-24
Update manpage for qvm-firewall

### LINKED ISSUE #7074: qvm-firewall: '--before' option not documented on manpage

`qvm-firewall` can only add rules at the end.

### The solution you'd like
`qvm-firewall` can insert rules in the middle as well. This could either be implemented as `qvm-firewall add --rule-no=` similar to `qvm-firewall del --rule-no=` or as `qvm-firewall insert [index] [rule]`.

`qvm-firewall clear` to fully clear a firewall and cause a deny-all would also be nice. (The currently existing `qvm-firewall reset` creates the default allow-all rule.)

Another nice feature would be to fully write a firewall from stdin, e.g. `qvm-firewall read < rules.txt`.

The last two proposed features can however rather easily be implemented in e.g. `bash` if an insert primitive is available, but not otherwise.

All of these implementations should _not_ cause some intermittent time frame during which the firewall allows something that wasn't allowed before.

Not needing to do a lot of `qvm-firewall del` operations and `qvm-firewall add` operations to insert a rule in a specific place.

unman:
Include `--before` option in `qvm-firewall add`
Closes QubesOS/qubes-issues#7074

### DIFF
--- a/doc/manpages/qvm-firewall.rst
+++ b/doc/manpages/qvm-firewall.rst
@@ -102,4 +102,4 @@ Authors
 
 | For complete author list see: https://github.com/QubesOS/qubes-core-admin-client.git
 
-.. vim: ts=3 sw=3 et tw=80*RULE_NUMBER*
+.. vim: ts=3 sw=3 et tw=80
--- a/doc/manpages/qvm-firewall.rst
+++ b/doc/manpages/qvm-firewall.rst
@@ -6,7 +6,7 @@
 Synopsis
 --------
 
-:command:`qvm-firewall` [-h] [--verbose] [--quiet] [--reload] *VMNAME* add *RULE*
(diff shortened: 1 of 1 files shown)


doc/manpages/qvm-firewall.rst
```diff
@@ -102,4 +102,4 @@ Authors
 
 | For complete author list see: https://github.com/QubesOS/qubes-core-admin-client.git
 
-.. vim: ts=3 sw=3 et tw=80
+.. vim: ts=3 sw=3 et tw=80*RULE_NUMBER*
```
marmarek: paste issue?

marmarek (changes requested):

marmarek (approved):

PR #224, opened 2022-11-04
backup/restore: hardcode MD5 algo for format v2/v3 decryption

rustybird:


marmarek:
Please update tests too, to generate backup as it used to be with the old openssl.
(tests currently aren't started automatically, but I hope the issue will be resolved soon)

rustybird:
Done. Missed it because I didn't set `ENABLE_SLOW_TESTS=1`

marmarek (approved):

PR #235, opened 2023-03-07
Document that qvm-volume revert can revert to a specific revision

Rudd-O:


Rudd-O:
Uh the codesigning thing says "no valid key found" even though the commit is verified according to Github.

Rudd-O:
image

Rudd-O:
Sometimes Github is just an asshole.

marmarek:
Those 4 commits also want squashing into one.

Rudd-O:
Will do right away.  Thanks.

marmarek (approved):

PR #237, opened 2023-03-23
Monitor and update keyboard layout only if the qube asked to be a guivm

### LINKED ISSUE #8109: qvm-start-daemon - audiovm - keyboard layout

When using a audiovm, the system will complain that the audiovm is not allowed to override the keyboard layout. 
Modifying the keyboard layout is normally only related to guivm, not audiovm. 

### The solution you'd like

I believe the issue is here:

https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/tools/qvm_start_daemon.py#L800
https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/tools/qvm_start_daemon.py#L792

The same actions are executed for both guivm and audiovm. 
This is not a behavior wanted for the audiovm. 
Maybe could add a check like that https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/tools/qvm_start_daemon.py#L741 to only execute thoses actions if the VM ask for guivm 

No useless warning, so not thinking that something don't work

neowutran:
Before this patch, qvm_start_daemon try to read and update keyboard layout regardless if the qube is a audiovm or guivm. 
With this patch it read and update keyboard layout only if the qube is a guivm.

Should fix https://github.com/QubesOS/qubes-issues/issues/8109

### DIFF
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -895,8 +895,15 @@ def main():
             lock_f.flush()
             lock_f.truncate()
             loop = asyncio.get_event_loop()
-            # pylint: disable=no-member
-            events = qubesadmin.events.EventsDispatcher(args.app)
+
+            if "guivm" in enabled_services:
+                # pylint: disable=no-member
+                events = qubesadmin.events.EventsDispatcher(args.app)
+            else:
+                # pylint: disable=no-member
+                events = qubesadmin.events.EventsDispatcher(args.app, enable_cache=False)
+
+            
             # pylint: enable=no-member
             launcher.register_events(events)
 


qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -785,19 +785,20 @@ def main(args=None):
...
-            conn = xcffib.connect()
-            x_watcher = XWatcher(conn, args.app)
-            x_fd = conn.get_file_descriptor()
-            loop.add_reader(x_fd, x_watcher.event_reader,
-                            events_listener.cancel)
-            x_watcher.update_keyboard_layout()
+            if "guivm" in enabled_services:
+                for signame in ('SIGINT', 'SIGTERM'):
+                    loop.add_signal_handler(getattr(signal, signame),
+                                            events_listener.cancel)  # pylint: disable=no-member
```
marmarek: This should be done whenever events_listener is started, not only in guivm.

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -785,19 +785,20 @@ def main(args=None):
...
+
+                loop.add_signal_handler(signal.SIGHUP,
+                                        launcher.send_monitor_layout_all)
```
marmarek: This one indeed is relevant to guivm only, but the signal handler can be dropped only when `watch-screen-layout-changes` (in gui-daemon repo) won't be started outside of guivm either. Otherwise it will send SIGHUP, which (without signal handler) will terminate qvm-start-daemon.

marmarek (changes requested):

qubesadmin/tools/qvm_start_daemon.py
neowutran: Thanks, should be ok with the new commit

qubesadmin/tools/qvm_start_daemon.py
neowutran: Thanks, moved it out of the guivm check. Should be ok

neowutran:
I restricted a bit more the permission in my audio qube setup. ( the users comments in the qubes forum all use the "admin.vm.properties.GetAll" permission. Probably better to avoid using this one ? ). 

I ended up with a some of others permissions warning coming from dom0. ( Tries to access the "guivm" property and tries to access the "GetAll" property ). 

I ended up modifying "qvm_start_daemon" to add services check and to not use the caching system of "EventsDispatcher": 
https://github.com/neowutran/qubes-core-admin-client/commit/1ae008d9f483085fb93b4c271be40ab85ec19e78

From my tests the warnings disappear as I wanted but I am not sure if it is the best way of doing that. 

I can update the goal & code of this pull request or create a new pull request

marmarek:
Disabling caching if it's only audiovm is probably okay, as it access much less properties. For guivm, the caching (and using GetAll) avoids several "Get" calls when starting a single qube. Your approach looks fine I think.

marmarek:
> * gui_keyboard_layout: [Failed](https://openqa.qubes-os.org/tests/70342#step/gui_keyboard_layout/47) (test died)
>       `# Test died: command 'test "$(cd ~user;ls e1*)" = "$(qvm-run -p wor...`

This failure does mean the keyboard layout setting got broken, so something went wrong in this PR.

neowutran:
Some additionnal modifications: 
- add more checks for currently enabled service in the vm. Don't try to ask RPC permission for not configured service to avoid triggering useless warning. 
- For the special case of dom0,  assign it all the possible service (variable "only_if_service_enabled" )

Notes:
It doesn't remove all the useless warning, only most of them. More work / more modifications are required to try to remove all the useless warnings

neowutran:
Can I help / improve something for this pull request ?

neowutran:
A lot of things have been modified since my last comment. I force pushed a new patch proposal: For things that are not GUIVM, don't enable the cache for the eventsdispatcher, so the audiovm won't be required to have the GetAll privilege. 

It was part of the original push request here. 
Now the name of the push request is quite far away from the content of the patch, so I could close this one and open a new pull request if needed

marmarek:
GetAll isn't really required, it's just an optimization. Better to add a rule that denies it with notify=no, to avoid annoying notification.

neowutran:
thanks a lot ! 
Wasn't aware of this option and it indeed seems to solve this issue. 
I close this pull request

PR #250, opened 2023-08-05
Filter-out keyboard layout switching options

### LINKED ISSUE #8035: Change of keyboard layouts with Capslock/Shift+Capslock (`shift_caps_switch` in `setxkbmap`) breaks all the time
### Qubes OS release
R4.1.1.

### Brief summary
Switching 2 keyboard layouts with Capslock for the first layout and Shift+Capslock for the second one (`shift_caps_switch` group in `setxkbmap`) is not working properly and breaks many times a day.

### Steps to reproduce
- During system installation add a second layout and set layout switch hotkey to Capslock for first layout and Shift+Capslock for the second one (`shift_caps_switch` group in `setxkbmap`)
- Use the system normally.

Note: Also settings this mode after installation with a single English layout has the same results.

### Expected behavior
Layout switch works.

### Actual behavior
After some time layout switch stops working completely. Very often and very annoying.
The way to temporary fix the issue is to press capslock and shift+capslock several times in some order, to exit some kind of wrong state of Qubes OS keyboard layout scripts. But soon the layout switching gets broken again.

### More
R4.0 worked fine, so this bug is a 100% **regression**.

marmarek:
When switching layout with a key combo, do not let VM to switch the
layout again (once as instructed by qvm-start-daemon, and then on its
own due to the key combo).

Fixes QubesOS/qubes-issues#8035

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,6 +16,7 @@ checks:pylint:
   - sudo dnf install -y python3-rpm
   - pip3 install --quiet -r ci/requirements.txt
   script:
+  - export PATH="$HOME/.local/bin:$PATH"
   - PYTHONPATH=test-packages pylint --extension-pkg-whitelist=rpm,lxml qubesadmin
   stage: checks
   tags:
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -179,6 +179,8 @@ class KeyboardLayout:
         self.languages = split_string[2].decode().split(',')
         self.variants = split_string[3].decode().split(',')
         self.options = split_string[4].decode()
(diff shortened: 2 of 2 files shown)


jamke:
@marmarek I see an issue in the code in case the layout option does not end in `_toggle`.
I provided a PR about it here QubesOS/qubes-core-admin-client/pull/255

The current problem shortly: Currently options like shift_caps_switch, lctrl_rctrl_switch and other first-last layout toggles are broken because they are not be filtered out.

The provided PR maybe does it better.
Note, that solution was not tested by me!
It is just an idea how it is possible to improve fc08f4e.

Docs: The list of options for keyboard layout switching is present in 'grp' chapter of /usr/share/X11/xkb/rules.
Note: I am not sure that "while pressed" shortcuts will work or if it even works now.

PR #252, opened 2023-08-07
Fix markup issue in qvm-create.rst: *NAME*=*VALUE*

jamke:
The used technique is described in reStructuredText Markup Specification, chapter Character-Level Inline Markup: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#character-level-inline-markup-1 
Note, that according to docs this solution should be used only when absolutely necessary.  I think it is exactly our case, because, it seems, there is no other proper way to separate equal sign out of emphasis of the NAME and VALUE. 
Github preview of `qvm-create.rst` file is now correct, too. 
Alternative solutions would be to use `*NAME=VALUE*` or NAME=VALUE instead, but the result will not be completely right.

### DIFF
--- a/doc/manpages/qvm-create.rst
+++ b/doc/manpages/qvm-create.rst
@@ -6,7 +6,7 @@
 Synopsis
 --------
 
-:command:`qvm-create` [-h] [--verbose] [--quiet] [--force-root] [--class *CLS*] [--property *NAME*=*VALUE*] [--pool *POOL_NAME:VOLUME_NAME*] [--template *VALUE*] --label *VALUE* [--root-copy-from *FILENAME* | --root-move-from *FILENAME*] *VMNAME*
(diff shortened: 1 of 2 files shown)


jamke:
@marmarek CI pipeline seems to be broken:

```
/usr/bin/bash: line 133: pylint: command not found
Cleaning up project directory and file based variables 00:00
ERROR: Job failed: exit code 1
```

marmarek:
PipelineRetry

jamke:
The same small change was not changing coverage 10 hours ago. Nor it should.
Please help me to make this trivial patch to be finally accepted. Should I git pull and submit this PR again?

marmarek (approved):

marmarek:
No need to change anything, test coverage change may happen due to other changes in the repository in the meantime, but that doesn't really matter for a documentation change.

jamke:
Once again sorry, I used github GUI to create the PR, and it did not create a patch branch. I did not noticed that and github started to automatically add my additional commits from my repo to this PR. That was not intended.
Instead of reverting I will recreate this trivial PR and also next ones properly with separated brunches on my end. Sorry again.

PR #254, opened 2023-08-08
Fix: qvm-ls outputs to stderr for --help-* calls

### LINKED ISSUE #8408: `qvm-ls --help-formats` and `qvm-ls --help-columns` output to `stderr`

### Qubes OS release
R4.1

### Brief summary
`qvm-ls --help-formats` and `qvm-ls --help-columns` outputs to `stderr` instead of `stdout`.
I think output goes to `sys.stderr` by mistake because it is obviously not considered to be an error to call `qvm-ls --help-formats`.

### Steps to reproduce
In `dom0`:
```
qvm-ls --help-formats 2>/dev/null | grep full
```

### Expected behavior
```
  full     NAME,STATE,CLASS,LABEL,QID,XID,UUID
```

### Actual behavior
(nothing)

jamke:
The problem is that ArgumentParser has no ability to provide output file for exit() calls. In the ArgumentParser documentation [1] there is even no information about output target file (stderr or stdout) for ArgumentParser.exit() functions, but from the source code of argparse [2] it is obvious that it always outputs message to stderr which is not desired in our case. [1]: https://docs.python.org/3/library/argparse.html#exiting-methods [2]: https://github.com/python/cpython/blob/3.11/Lib/argparse.py#L2612

I have tested this patch on my system, it works as expected.

Fixes QubesOS/qubes-issues#8408

### DIFF
--- a/qubesadmin/tools/qvm_ls.py
+++ b/qubesadmin/tools/qvm_ls.py
@@ -529,7 +529,8 @@ class _HelpColumnsAction(argparse.Action):
             for column in sorted(Column.columns.values()))
         text += '\n\nAdditionally any VM property may be used as a column, ' \
                 'see qvm-prefs --help-properties for available values'
-        parser.exit(message=text + '\n')
+        print(text + '\n')
+        sys.exit(0)
 
 
 class _HelpFormatsAction(argparse.Action):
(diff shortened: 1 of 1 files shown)


marmarek:
PipelineRetry

marmarek (approved):

PR #257, opened 2023-08-17
Make qvm-run --nogui default if DISPLAY is not set

marmarek:
Using qvm-run outside of GUI session is currently quite annoying,
because one needs to set --nogui each time. Otherwise the tool waits for
GUI session to initialize.
Solve this by making qvm-run --nogui default if DISPLAY is not set.

Fixes QubesOS/qubes-issues#8431


marmarek:
PipelineRetry

PR #258, opened 2023-09-07
Minor fix misspelling in docstring

gonzalo-bulnes:
## Current

A _docstring_ that describes `qvm-features` in the `qvm-tags` file.

## Expected

A _docstring_ that describes `qvm-tags`. :slightly_smiling_face:


gonzalo-bulnes:
:eyes: I don't think I can retry the builds that have failed, but I'm pretty confident it's not the change set of this PR that made them fail. (Only because it is a very tiny documentation change.)

marmarek (approved):

PR #267, opened 2023-12-11
Add a mock Qubes object with improved contents

marmarta:
Also several convenience methods and pre-populated Qubes setups.

### DIFF
--- a/qubesadmin/tests/mock_app.py
+++ b/qubesadmin/tests/mock_app.py
@@ -1,0 +1,686 @@
+# -*- encoding: utf8 -*-
+#
+# The Qubes OS Project, http://www.qubes-os.org
+#
+# Copyright (C) 2023 Marta Marczykowska-Górecka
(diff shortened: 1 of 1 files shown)


marmarta:
PipelineRetryFailed

qubesadmin/tests/mock_app.py
```diff
@@ -0,0 +1,698 @@
...
+        """The complete response string."""
+        return b"0\x00" + str(self).encode()
+
+    def __str__(self):
+        if self.volume_type == 'kernel':
+            result = "pool=linux-kernel\n"
+            result += 'vid=1.1\n'
+            result += 'size=0\n'
+            result += 'usage=0\n'
+            result += "rw=True\n"
```
marmarek: That's the other way around: here rw=False, and otherwise rw=True.

marmarek (changes requested):

marmarek (approved):

PR #269, opened 2023-12-22
Add symlinks to qvm-device

### LINKED ISSUE #8786: qubes-core-admin-client missing /usr/bin/qvm-block?

### Qubes OS release
4.1

### qubes-core-admin-client version:
  Installed: 4.1.27-1+deb11u1
  Candidate: 4.1.27-1+deb11u1
  Version table:
 *** 4.1.27-1+deb11u1 500
        500 https://deb.qubes-os.org/r4.1/vm bullseye/main amd64 Packages
        100 /var/lib/dpkg/status

### Brief summary
Not sure if this is intended, but I believe that /usr/bin/qvm-block is supposed to be a symbolic link to  /usr/bin/qvm-device .  If it is supposed to be there, then since qvm-device was installed by qubes-core-admin-client, I'm guessing that qubes-core-admin-client was supposed to be the one to add the link.

### Steps to reproduce
sudo apt install qubes-core-admin-client
ls -l /usr/bin/qvm-block

### Expected behavior
see symbolic link to /usr/bin/qvm-device

### Actual behavior
no such file

DemiMarie:
These were only in dom0, which is inconsistent.

Fixes: QubesOS/qubes-issues#8786

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ install:
 	install -m 0644 etc/qvm-start-daemon-kde.desktop $(DESTDIR)/etc/xdg/autostart/
 	install -d $(DESTDIR)/usr/bin
 	ln -sf qvm-start-daemon $(DESTDIR)/usr/bin/qvm-start-gui
+	for i in block usb pci; do ln -sf qvm-device $(DESTDIR)/usr/bin/"qvm-$$i"; done
 	install -m 0755 scripts/qubes-guivm-session $(DESTDIR)/usr/bin/
 	install -d $(DESTDIR)/etc/qubes/post-install.d
 	install -m 0755 scripts/30-keyboard-layout-service.sh \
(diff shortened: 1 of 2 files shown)


marmarek:
For this to work, there needs to be a similar PR removing qvm-device symlinks from core-admin, otherwise packages will conflict.

DemiMarie:
What about package upgrades?

marmarek:
It will require appropriate dependencies for clean upgrades. I'll add them after merging.

marmarek (changes requested):
Man page needs symlinks too.

marmarek:
> It will require appropriate dependencies for clean upgrades. I'll add them after merging.

Interestingly, rpm is happy with two packages owning the same file - I guess it works because the file is exactly the same.

marmarek (approved):

PR #271, opened 2024-01-26
Improve mock API

marmarta:
add service.updates-proxy-setup feature to TemplateVMs.


marmarek (changes requested):
It isn't visible this way in real system, so shouldn't be in mockup either. The `updates-proxy-setup` service is implicite enabled within the template, not from dom0.

PR #275, opened 2024-02-06
minor cleanup of rpm2cpio usage leftover in var name

slayoo:
In https://github.com/QubesOS/qubes-core-admin-client/commit/2ac29639a3382ea3d518d87b4df5e74e87ec0053, a call to `rpm2cpio` was replaced with a call to `rpm2archive`. However, one variable is still named `rpm2cpio` around that location. This PR just renames that var to avoid misleading the reader.

Well, nothing worth a separate PR - just wanted to confirm that I'm looking into https://github.com/QubesOS/qubes-issues/issues/8876

### DIFF
--- a/qubesadmin/tools/qvm_template.py
+++ b/qubesadmin/tools/qvm_template.py
@@ -743,13 +743,13 @@ def extract_rpm(name: str, path: str, target: str) -> bool:
     with open(path, 'rb') as pkg_f:
         with subprocess.Popen(['rpm2archive', "-"],
                 stdin=pkg_f,
-                stdout=subprocess.PIPE) as rpm2cpio:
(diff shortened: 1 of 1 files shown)


marmarek (commented):
This looks okay, but IIUC you want to push more stuff to this PR, right?

slayoo:
> This looks okay, but IIUC you want to push more stuff to this PR, right?

Yes, this code is part of `extract_rpm()` which will likely get an additional parameter or have its logic changed for https://github.com/QubesOS/qubes-issues/issues/8876

But this tiny change belongs more to the earlier commit, so can be also merged as is.

slayoo:
Thanks!

PR #277, opened 2024-02-13
Enable starting DispVMs from admin API

marmarta:
Add a start() method to DispVMs that allows the user to start one. Makes it easier to create a dispvm that's not tied to a single process.

### DIFF
--- a/qubesadmin/vm/__init__.py
+++ b/qubesadmin/vm/__init__.py
@@ -428,6 +428,20 @@ class DispVMWrapper(QubesVM):
             except qubesadmin.exc.QubesVMNotRunningError:
                 pass
 
+    def start(self):
+        """Start this DispVM"""
+        # create dispvm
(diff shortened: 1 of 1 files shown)


marmarta:
PipelineRetryFailed

qubesadmin/vm/__init__.py
```diff
@@ -428,6 +428,21 @@ def cleanup(self):
             except qubesadmin.exc.QubesVMNotRunningError:
                 pass
 
+    def start(self):
+        """Start this DispVM"""
+        if self.app.qubesd_connection_type == 'socket':
```
marmarek: I don't think this condition is necessary in `start()` method. It's used in `run_service()` because when running outside of dom0 the DispVM will be created automatically on a service call. But it isn't the case for just starting a DispVM.

marmarek (changes requested):

PR #279, opened 2024-03-04
printing list object as one string for list-like properties

### LINKED ISSUE #9007: `qvm-ls -O=tags` produces unreadable output
### Qubes OS release

`Q4.2` and `Q4.1`.

### Brief summary

`qvm-ls -O=tags` produces unreadable output instead of a list of all tags for all VMs. 

### Steps to reproduce

Just run `qvm-ls -O=tags`.

### Expected behavior

List of tags:

```
NAME                       TAGS
browser-dvm                tags tags tags
debian-12-minimal          tags tags tags
```

### Actual behavior

Something close to being a list of tags:

```
NAME                       TAGS
browser-dvm                <qubesadmin.tags.Tags object at 0x7c0202b0aa10>
debian-12-minimal          <qubesadmin.tags.Tags object at 0x7c0202b0add0>
```

ZYousef:
Fixes https://github.com/QubesOS/qubes-issues/issues/9007 print tags list as one string instead of object.

### DIFF
--- a/qubesadmin/tools/qvm_ls.py
+++ b/qubesadmin/tools/qvm_ls.py
@@ -107,6 +107,8 @@ class Column(object):
                 ret = vm
                 for attrseg in self._attr.split('.'):
                     ret = getattr(ret, attrseg)
+                    if hasattr(ret, '__iter__') and not isinstance(ret, str):
+                        ret = ' '.join(map(str, ret))
             elif isinstance(self._attr, collections.abc.Callable):
                 ret = self._attr(vm)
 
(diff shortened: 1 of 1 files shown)


qubesadmin/tools/qvm_ls.py
```diff
@@ -107,6 +107,8 @@ def format(self, vm):
                 ret = vm
                 for attrseg in self._attr.split('.'):
                     ret = getattr(ret, attrseg)
+                    if hasattr(ret, '__iter__') and not isinstance(ret, str):
+                        ret = ' '.join(map(str, ret))
```
marmarek: better separate with coma (which also cannot be part of the tag name), otherwise columns may be ambiguous.

marmarek (changes requested):

qubesadmin/tools/qvm_ls.py
marmarek: I meant just coma, to avoid spaces within the column content...

ZYousef: sorry about that, I just amended the previous commit to replace space with a comma

marmarek (approved):

PR #281, opened 2024-03-26
Reformat using black and add lint to CI

fepitre:
The command `black -l 80 qubesadmin` has been performed together with adding lint job to `.gitlab-ci.yml`.


fepitre:
Please see second commit that makes `pylint` happy. There is minor manual format adjustments.

PR #283, opened 2024-03-30
Support for dynamic audiovm switching

fepitre:
https://github.com/QubesOS/qubes-issues/issues/8975

### DIFF
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -362,7 +362,8 @@ def get_monitor_layout():
 class DAEMONLauncher:
     """Launch GUI/AUDIO daemon for VMs"""
 
-    def __init__(self, app: qubesadmin.app.QubesBase, vm_names=None, kde=False):
+    def __init__(self, app: qubesadmin.app.QubesBase, enabled_services,
+                 vm_names=None, kde=False):
         """ Initialize DAEMONLauncher.
 
         :param app: :py:class:`qubesadmin.Qubes` instance
@@ -370,6 +371,7 @@ class DAEMONLauncher:
         :param kde: add KDE-specific arguments for guid
         """
         self.app = app
+        self.enabled_services = enabled_services
         self.started_processes = {}
         self.vm_names = vm_names
         self.kde = kde
@@ -377,10 +379,6 @@ class DAEMONLauncher:
         # cache XID values when the VM was still running -
         # for cleanup purpose
         self.xid_cache = {}
-        for qube in self.app.domains:
-            if qube.name == "dom0":
-                continue
-            self.xid_cache[qube] = qube.xid, qube.stubdom_xid
 
     async def send_monitor_layout(self, vm, layout=None, startup=False):
         """Send monitor layout to a given VM
@@ -699,16 +697,20 @@ class DAEMONLauncher:
 
             power_state = vm.get_power_state()
             if power_state == 'Running':
-                asyncio.ensure_future(
-                    self.start_gui(vm, monitor_layout=monitor_layout))
-                asyncio.ensure_future(self.start_audio(vm))
+                if "guivm" in self.enabled_services:
+                    asyncio.ensure_future(
+                        self.start_gui(vm, monitor_layout=monitor_layout)
+                    )
+                if "audiovm" in self.enabled_services:
+                    asyncio.ensure_future(self.start_audio(vm))
                 self.xid_cache[vm.name] = vm.xid, vm.stubdom_xid
             elif power_state == 'Transient':
                 # it is still starting, we'll get 'domain-start'
                 # event when fully started
-                if vm.virt_mode == 'hvm':
+                if vm.virt_mode == 'hvm' and "guivm" in self.enabled_services:
                     asyncio.ensure_future(
-                        self.start_gui_for_stubdomain(vm))
+                        self.start_gui_for_stubdomain(vm)
+                    )
 
     def on_domain_stopped(self, vm, _event, **_kwargs):
         """Handler of 'domain-stopped' event, cleans up"""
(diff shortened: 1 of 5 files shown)


marmarek:
This made the default `qvm-start-daemon --all --watch` not start in dom0 (it exits immediately).
When I use `--force`, it starts, but then it tries to start audio processes several times for each VM (I stopped counting at 10).

fepitre:
> it

I've not tested in dom0, I guess this is because I removed `/etc/qubes-release`..

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -709,59 +824,56 @@ def is_watched(self, vm):
...
-    only_if_service_enabled = ['guivm', 'audiovm']
-    enabled_services = [service for service in only_if_service_enabled if
-                        os.path.exists('/var/run/qubes-service/%s' % service)]
-    if not enabled_services and '--force' not in sys.argv and \
-            not os.path.exists('/etc/qubes-release'):
-        print(parser.format_help())
+    if args.force or os.path.exists("/etc/qubes-release"):
+        enabled_services = only_if_service_enabled
+
+    if not enabled_services:
```
marmarek, line 828: It turned out checking this before `parser.parse_args()` was important - otherwise all qubes attempt to do `admin.vm.List` call now.

marmarek (changes requested):

marmarek:
One more thing I noticed - the audiovm tries to access "guivm" property (I get a message about policy deny), even though only "audiovm" service is enabled. I haven't looked yet where it tries to access it.

marmarek:
This should have "fixes https://github.com/QubesOS/qubes-issues/issues/8966" tag in "cleanup pacat process" commit message

marmarek:
And maybe "fixes https://github.com/QubesOS/qubes-issues/issues/8975" in some other too?

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -699,6 +814,10 @@ def register_events(self, events):
                            self.on_connection_established)
         events.add_handler('domain-stopped', self.on_domain_stopped)
 
+        for event in ["property-set:audiovm", "property-pre-set:audiovm",
+                      "property-pre-del:audiovm"]:
```
marmarek, line 803: Why all three? especially, it should be only `property-set`, not `property-pre-set`, but then probably also `property-del` (or more intuitively `property-reset`), not `property-pre-del`.

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -709,59 +828,56 @@ def is_watched(self, vm):
...
 
     if args.watch:
-        fd = os.open(args.pidfile,
-                     os.O_RDWR | os.O_CREAT | os.O_CLOEXEC,
-                     0o600)
+        # Start qubes running before we started qvm-start-daemon
+        for qube in launcher.app.domains:
+            if qube.name == "dom0":
+                continue
+            launcher.on_connection_established(
```
marmarek: This is wrong, it should be called only when really receiving `connection-established` event, otherwise you will miss anything that happens before you start receiving events.

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -813,23 +937,36 @@ def main(args=None):
...
+                        )
+                    )
         if tasks:
             loop.run_until_complete(asyncio.wait(tasks))
         loop.stop()
         loop.run_forever()
         loop.close()
 
+    # Cleanup any started daemon
+    launcher.cleanup()
```
marmarek: This sound like a bad idea, `qvm-start-daemon` is responsible for starting those daemons, but restarting `qvm-start-daemon` (for any reason) shouldn't kill all the audio/gui daemons...

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -653,16 +702,20 @@ def on_connection_established(self, _subject, _event, **_kwargs):
...
+                        self.start_gui(vm, monitor_layout=monitor_layout)
+                    )
+                if "audiovm" in self.enabled_services:
+                    asyncio.ensure_future(self.start_audio(vm))
                 self.xid_cache[vm.name] = vm.xid, vm.stubdom_xid
             elif power_state == 'Transient':
                 # it is still starting, we'll get 'domain-start'
                 # event when fully started
-                if vm.virt_mode == 'hvm':
+                if vm.virt_mode == 'hvm' and "guivm" in self.enabled_services:
```
marmarek: Put the enabled services check first, to avoid accessing virt_mode property if not needed.

marmarek (changes requested):
The "qvm-start-daemon: put parser into main func" commit is basically reverted later, can you squash those two together?

qubesadmin/tools/qvm_start_daemon.py
marmarek: (based on direct chat) this is because changing audiovm changes also tags, which affects qrexec policy - after unsetting audiovm, old audiovm won't have access to VM's properties; and similarly, when setting new, it's accessible only after tags are updated. This could use a comment, or an explanation in the commit message.

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -678,19 +731,68 @@ def on_domain_stopped(self, vm, _event, **_kwargs):
...
+        if vm.name not in self.xid_cache:
+            try:
+                self.xid_cache[vm.name] = vm.xid, vm.stubdom_xid
+            except qubesadmin.exc.QubesDaemonAccessError as e:
+                log.error("vm.name: failed to determine XID: %s", str(e))
+                return
+        xid, stubdom_xid = self.xid_cache[vm.name]
+        oldvalue = _kwargs.get("oldvalue", None)
+        newvalue = _kwargs.get("newvalue", None)
+        if oldvalue == self.app.local_name and newvalue == "None":
```
marmarek: Shouldn't it be `newvalue != self.app.local_name` ? Right now when I change audiovm from dom0 to sys-audio, the pacat process in dom0 remains running. Or is it intentional for more reliable switching back?

marmarek: But also, this means if I first switch from sys-audio to dom0 and then shutdown the qube, its pacat remains in sys-audio (it's terminated in dom0 only).

fepitre: > Shouldn't it be `newvalue != self.app.local_name` ? Right now when I change audiovm from dom0 to sys-audio, the pacat process in dom0 remains running. Or is it intentional for more reliable switching back?

No it's not intentional, I think it was at the very early work for this to handle all the cases in the same function without rethinking the whole after. I'll fix that indeed.

PR #287, opened 2024-05-03
qvm-start-daemon: fix TypeError in message

fepitre:


marmarek:
Not really an issue with this PR, but I think the fix belongs here:
1. `qvm-start testvm`
2. `qvm-prefs testvm audiovm none`
3. `qvm-prefs --default testvm audiovm`

At this point `qvm-prefs testvm audiovm` shows dom0 again now (as that's default on this system), but pacat isn't running there. I think qvm-start-daemon misses handling `property-reset:audiovm` event - but note this event doesn't have `newvalue` parameter (but it does have `oldvalue`), you need to read the value from the property. But I do see `property-pre-del:audiovm`, which is `property-pre-reset:audiovm`, so maybe it's just about this `newvalue` parameter.

While at it, I think pre- and non-pre- events wants splitting to do just cleanup in the pre- event and just starting in non-pre events. Otherwise at least dom0 instance (but others, if qrexec policy allows) will try to do things twice.

marmarek (approved):

PR #295, opened 2024-06-23
qvm-start-daemon: expect gui-window-background-color, gui-default-window-background-color

### LINKED ISSUE #9304: Allow the gui-daemon background color to be configurable

This is an outgrowth of this forum thread:

https://forum.qubes-os.org/t/suppressing-the-white-flash/26829

I've filed this as an enhancement but it's a mitigation of a usability bug that affects some proportion of users.

**Context:**

When VM window content is not immediately available to dom0's X, but something must be displayed in the dom0 passthrough window for that VM window, the `gui-daemon` falls back to filling this unavailable content with a hardcoded background color.

Examples:
1. A domU window is resized on dom0 and hasn't yet been redrawn by the VM.
2. The user switches to another workspace which triggers redraws of domU windows placed on that workspace and the passthrough windows are displayed before the redrawn domU content is available. (I am supposing here this is the event flow.)

This display of the hardcoded background color is usually quite transient, milliseconds, but is briefly visible to the user.

**Problem:**

The background color is hardcoded to white. If the user is running a dark color theme on their desktop, a window's brief flash of bright white is visually uncomfortable, especially if it happens frequently. Especially especially if the window is large or full-screen.

In my own workflow I switch between workspaces very frequently, which means quite frequent white flashes.

### The solution you'd like

Allow the `gui-daemon` background color to be black instead of white. I patched my `gui-daemon` to use black and for me it's been a complete mitigation of this usability issue. Users who run a bright desktop theme probably would prefer the current white, though. Therefore the background color should probably be configurable.

marmarek outlines the shape of that work [here](https://forum.qubes-os.org/t/suppressing-the-white-flash/26829/10):
> I guess it has to do with “background pixel”:
> 
> https://github.com/QubesOS/qubes-gui-daemon/blob/main/gui-daemon/xside.c#L332
> 
> If somebody wants to play with it, it should be quite easy to change it and see if it helps. For example there is similar `BlackPixel` function. Making it configurable should also be possible, but it’s going to be a bit more work (add config option to gui-daemon, fetch it from the config file there, and then add support for setting this option via features in core-admin-client).
> 

My instinct is that an option of black or white is good enough; allowing the color to be an arbitrary hue perhaps grants this issue more attention than it deserves. Also, the implementation is simpler. But I defer to others.

The majority of the work of the implementation will be the plumbing to propagate the color choice to `gui-daemon/xside.c`.

Better usability/comfort for users like me who use dark styling.

Euwiiwueir:
Look for these features as settable by `qvm-features`:
- `gui-window-background-color`
- `gui-default-window-background-color`

Translate them into the `window_background_color` setting in the configuration file generated for `qubes-guid`.

See also: QubesOS/qubes-gui-daemon/pull/144
Fixes QubesOS/qubes-issues#9304

### DIFF
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -52,6 +52,7 @@ GUI_DAEMON_OPTIONS = [
     ('secure_paste_sequence', 'str'),
     ('windows_count_limit', 'int'),
     ('trayicon_mode', 'str'),
+    ('window_background_color', 'str'),
     ('startup_timeout', 'int'),
 ]
 


andrewdavidwong:
Suggestion: If you intend for this PR to fix/close issue https://github.com/QubesOS/qubes-issues/issues/9304 and would like it to be linked to that issue automatically, use:

```
Fixes QubesOS/qubes-issues#9304
```

instead of

```
For: QubesOS/qubes-issues#9304
```

in the PR and/or commit message.

`Fixes` is a special keyword that GitHub will recognize. For more information, see:

- https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue
- https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

marmarek (approved):

PR #297, opened 2024-07-10
Improve qvm-check mechanism on invalid domains

### LINKED ISSUE #3496: qvm-check's output and retcode kind of buggy
#### Qubes OS version:

R4.0 rc3

---

I'm scripting some stuff in dom0 and need to check if a VM exists ; qvm-check seems to be the way to go rather than grepping the output of qvm-ls.

Let's assume that a VM named "work" exists and that a VM named "work1" doesn't.

Case 1: checking if VM "work" is running: works as expected
```
[user@dom0 ~]$ qvm-check work
VM work exists
[user@dom0 ~]$ echo $?
0
```

Case 2: adding the --quiet arg: the output isn't suppressed
```
[user@dom0 ~]$ qvm-check --quiet work
VM work exists
```

Case 3: finding out what the retcode for a wrong syntax is (to compare with case 4 below)
```
[user@dom0 ~]$ qvm-check --blahblah
[...]
[user@dom0 ~]$ echo $?
2
```

Case 4: checking if a non existent VM exists: usage() is displayed and usage()'s retcode is returned despite the syntax being right; one would expect the retcode "1"
```
[user@dom0 ~]$ qvm-check work1
usage: qvm-check [-h] [--verbose] [--quiet] [--all] [--exclude EXCLUDE]
                 [--running] [--paused] [--template]
                 [VMNAME [VMNAME ...]]
qvm-check: error: no such domain: 'work1'
[user@dom0 ~]$ echo $?
2
```

I looked at /usr/lib/python3.5/site-packages/qubesadmin/tests/tools/qvm_check.py to try to fix it and send a PR, but it was a bit too cryptic for me.

Very minor issue BTW.

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/3496

### DIFF
--- a/qubesadmin/tools/qvm_check.py
+++ b/qubesadmin/tools/qvm_check.py
@@ -35,7 +35,7 @@ class QvmCheckArgumentParser(qubesadmin.tools.QubesArgumentParser):
 
     def error(self, message):
         if message.startswith('no such domain: '):
-            self._invalid_domains.append(message[17:])
+            self._invalid_domains.append(message[17:-1])
         else:
             super().error(message)
 
--- a/qubesadmin/tools/qvm_check.py
+++ b/qubesadmin/tools/qvm_check.py
@@ -27,26 +27,49 @@ import sys
 import qubesadmin.tools
 import qubesadmin.vm
 
-class QvmCheckArgumentParser(qubesadmin.tools.QubesArgumentParser):
-    """Collecting error message(s) on invalid domain(s) instead of aborting"""
-    def __init__(self, description, vmname_nargs):
-        super().__init__(description=description, vmname_nargs=vmname_nargs)
-        self._invalid_domains = []
+class QvmCheckVmNameAction(qubesadmin.tools.VmNameAction):
+    """ Action for parsing one or multiple valid/invalid VMNAMEs """
 
-    def error(self, message):
-        if message.startswith('no such domain: '):
-            self._invalid_domains.append(message[17:-1])
+    def __init__(self, option_strings, nargs=1, dest='vmnames', help=None,
+                 **kwargs):
+        # pylint: disable=redefined-builtin
+        super().__init__(option_strings, dest=dest, help=help,
+                                           nargs=nargs, **kwargs)
+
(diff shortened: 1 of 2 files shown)


alimirjamali:
> Attention: Patch coverage is `65.00000%` with `7 lines` in your changes missing coverage. Please review.

Ok. review bot. I will write the requested Unit Tests.

qubesadmin/tools/qvm_check.py
```diff
@@ -20,15 +20,33 @@
...
-                                              vmname_nargs='+')
+class QvmCheckArgumentParser(qubesadmin.tools.QubesArgumentParser):
+    """Collecting error message(s) on invalid domain(s) instead of aborting"""
+    def __init__(self, description, vmname_nargs):
+        super().__init__(description=description, vmname_nargs=vmname_nargs)
+        self._invalid_domains = []
+
+    def error(self, message):
+        if message.startswith('no such domain: '):
+            self._invalid_domains.append(message[17:-1])
```
marmarek: I don't like this approach for two reasons:
- matching error message by string is fragile
- `error` method normally terminates the program, changing this behavior is asking for troubles

Better extend QubesArgumentParser with this functionality (an option to collect invalid domains instead of exiting?), instead of catching the error after the fact.

marmarek (changes requested):

qubesadmin/tools/qvm_check.py
alimirjamali: > Better extend QubesArgumentParser with this functionality (an option to collect invalid domains instead of exiting?), instead of catching the error after the fact.

It could be done. But the patch will be longer. As two classes have to be extended. I will work on it.

alimirjamali: It is done. I am waiting for CI/CD bot to re-evaluate it to see if any additional unit test is necessary.

alimirjamali:
Making pylint happy(er). The original `VmNameGroup` had the same `# pylint: disable=redefined-builtin` directive.

marmarek (approved):

PR #301, opened 2024-07-25
Improve qvm-ls options to avoid parsing

### LINKED ISSUE #8614: Improve qvm-ls options to avoid parsing

Currently, it is lacking some things that `qvm-check` has such as qube class (--template), state ( --paused, --running).

I don't want to duplicate the tools but I believe the following options fits better on qvm-ls:
* --class DispVM|StandaloneVM|AppVM|TemplateVM
* --state Running|Paused
* --no-tags TAG
* --service SERVICE=True/on/False/off
* --features FEATURE=value

This list can be improved, there are countless options.

Currently developers have to parse this information by using `--no-spinner --raw-data` and grep/awk or parse via another language for the qubes class and it's state.
To avoid printing a qube that has a tag, currently I have to `qvm-ls --tags TAG` and then get this output and separate it in multiple `--exclude` options: `qvm-ls --all --exclude QUBE1 --exclude QUBE2 --exclude QUBE3`.

### The solution you'd like

I would like for more use of `--raw-list` with extra option instead of parsing `--raw-data`.
Any disadvantages besides making the `qubesadmin/tools/qvm_ls.py` bigger?

Easy way to get a list of qubes with certain characteristics without parsing.

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8614

Filtering, Formatting & Sorting ArgumentParser Groups
Sorting options based on output columns
A new pref(erence) format
Dedicated filter options for class, label, template, netvm preferences
Dedicated filter options for internal,servicevm & updates-available
Generic filtering options for preferences and features

### DIFF
--- a/doc/manpages/qvm-ls.rst
+++ b/doc/manpages/qvm-ls.rst
@@ -8,20 +8,27 @@ Synopsis
 
 :command:`qvm-ls` [-h] [--verbose] [--quiet] [--help-columns] [--help-formats] [--format *FORMAT* | --fields *FIELD*,...] [--tags *TAG* [*TAG* ...]] [--running] [--paused] [--halted]
 
-Options
--------
+Positional arguments
+--------------------
+
+.. option:: VMNAME
+
+   Zero or more domain names
+
+General options
+---------------
 
 .. option:: --help, -h
 
    Show help message and exit
 
-.. option:: --help-columns
+.. option:: --verbose, -v
 
-   List all available columns with short descriptions and exit.
+   Increase verbosity.
 
-.. option:: --help-formats
+.. option:: --quiet, -q
 
-   List all available formats with their definitions and exit.
+   Decrease verbosity.
 
 .. option:: --all
 
@@ -31,6 +38,17 @@ Options
 
    Exclude the qube from --all. You need to use --all option explicitly to use
    --exclude.
+
+.. option:: --spinner
+
+   Have a spinner spinning while the spinning mainloop spins new table cells.
+
+.. option:: --no-spinner
+
+   No spinner today.
+
+Formatting options
+------------------
 
 .. option:: --format=FORMAT, -o FORMAT
 
@@ -43,14 +61,10 @@ Options
    :option:`--format`. All columns along with short descriptions can be listed
    with :option:`--help-columns`.
 
-.. option:: --tags TAG ...
+.. option:: --tree, -t
 
-   Shows only VMs having specific tag(s).
-
-.. option:: --running, --paused, --halted
-
-   Shows only VMs matching the specified power state(s). When none of these
-   options is used (default), all VMs are shown.
+   List domains as a network tree. Domains are sorted as they are connected to
+   their netvms. Names are indented relative to the number of connected netvms.
 
 .. option:: --raw-data
 
@@ -61,11 +75,6 @@ Options
 
    Give plain list of VM names, without header or separator. Useful in scripts.
    Same as --raw-data --fields=name
-
-.. option:: --tree, -t
-
-   List domains as a network tree. Domains are sorted as they are connected to
-   their netvms. Names are indented relative to the number of connected netvms.
 
 .. option:: --disk, -d
 
@@ -79,21 +88,83 @@ Options
 
    Same as --format=kernel, for compatibility with Qubes 3.x
 
-.. option:: --verbose, -v
+.. option:: --help-columns
 
-   Increase verbosity.
+   List all available columns with short descriptions and exit.
 
-.. option:: --quiet, -q
+.. option:: --help-formats
 
-   Decrease verbosity.
+   List all available formats with their definitions and exit.
 
-.. option:: --spinner
+Filtering options
+-----------------
 
-   Have a spinner spinning while the spinning mainloop spins new table cells.
+.. option:: --class CLASS ...
 
-.. option:: --no-spinner
+   Show only qubes of specific class(es)
 
-   No spinner today.
+.. option:: --label LABEL ...
+
+   Show only qubes with specific label(s)
+
+.. option:: --tags TAG ...
+
+   Shows only VMs having specific tag(s).
+
+.. option:: --exclude-tags TAG ...
+
+   Exclude VMs having specific tag(s).
+
+.. option:: --running, --paused, --halted
+
+   Shows only VMs matching the specified power state(s). When none of these
+   options is used (default), all VMs are shown.
+
+.. option:: --template-source TEMPLATE ...
+
+   Filter results to the qubes based on the TEMPLATE(s)
+
+.. option:: --netvm-is NETVM ...
+
+   Filter results to the qubes connecting via NETVM(s)
+
+.. option:: --internal <y|n>
+
+   Show only internal qubes or exclude them from output
+
+.. option:: --servicevm <y|n>
+
+   Show only servicevms or exclude them from output
+
+.. option:: --pending-update
+
+   Filter results to qubes pending for update
+
+.. option:: --features FEATURE=VALUE ...
+
+   Filter results to qubes that match all specified features. Omitted VALUE
+   means None (not set). "" or '' means blank
+
+.. option:: --prefs PREFERENCE=VALUE ...
+
+   Filter results to qubes that match all specified preferences. Omitted VALUE
+   means None (not set). "" or '' means blank
+
+Sorting options
+---------------
+
+.. option:: --sort COLUMN
+
+   Sort based on provided column rather than NAME. Sort key should be in the
+   output columns
+
+.. option:: --reverse
+
+   Reverse sort
+
+.. option:: --ignore-case
+
+   Ignore case distinctions for sorting
 
 Authors
 -------
--- a/doc/manpages/qvm-ls.rst
+++ b/doc/manpages/qvm-ls.rst
@@ -6,12 +6,26 @@
 Synopsis
 --------
 
-:command:`qvm-ls` [-h] [--verbose] [--quiet] [--help-columns] [--help-formats] [--format *FORMAT* | --fields *FIELD*,...] [--tags *TAG* [*TAG* ...]] [--running] [--paused] [--halted]
+:command:`qvm-ls` [--verbose] [--quiet] [--help] [--all]
+                  [--exclude *EXCLUDE*] [--spinner] [--no-spinner]
+                  [--format *FORMAT* | --fields *FIELD* [*FIELD* ...] | --disk | --network | --kernel]
+                  [--tree] [--raw-data] [--raw-list] [--help-formats]
+                  [--help-columns] [--class *CLASS* [*CLASS* ...]]
+                  [--label *LABEL* [*LABEL* ...]] [--tags *TAG* [*TAG* ...]]
+                  [--exclude-tags *TAG* [*TAG* ...]] [--running] [--paused]
+                  [--halted] [--template-source *TEMPLATE* [*TEMPLATE* ...]]
+                  [--netvm-is *NETVM* [*NETVM* ...]] [--internal <y|n>]
+                  [--servicevm <y|n>] [--pending-update]
+                  [--features *FEATURE=VALUE* [*FEATURE=VALUE* ...]]
+                  [--prefs *PREFERENCE=VALUE* [*PREFERENCE=VALUE* ...]]
+                  [--sort *COLUMN*] [--reverse] [--ignore-case]
+                  [VMNAME ...]
+
 
 Positional arguments
 --------------------
 
-.. option:: VMNAME
+.. option:: VMNAME ...
 
    Zero or more domain names
 
--- a/doc/manpages/qvm-ls.rst
+++ b/doc/manpages/qvm-ls.rst
@@ -1,6 +1,6 @@
 .. program:: qvm-ls
 
-:program:`qvm-ls` -- List VMs and various information about them
+:program:`qvm-ls` -- List qubes and various information about them
 ================================================================
 
 Synopsis
@@ -77,7 +77,7 @@ Formatting options
 
 .. option:: --tree, -t
 
-   List domains as a network tree. Domains are sorted as they are connected to
+   List qubes as a network tree. Qubes are sorted as they are connected to
    their netvms. Names are indented relative to the number of connected netvms.
 
 .. option:: --raw-data
@@ -123,16 +123,16 @@ Filtering options
 
 .. option:: --tags TAG ...
 
-   Shows only VMs having specific tag(s).
+   Shows only qubes having specific tag(s).
 
 .. option:: --exclude-tags TAG ...
 
-   Exclude VMs having specific tag(s).
+   Exclude qubes having specific tag(s).
 
 .. option:: --running, --paused, --halted
 
-   Shows only VMs matching the specified power state(s). When none of these
-   options is used (default), all VMs are shown.
+   Shows only qubes matching the specified power state(s). When none of these
+   options is used (default), all qubes are shown.
 
 .. option:: --template-source TEMPLATE ...
 
@@ -157,19 +157,19 @@ Filtering options
 .. option:: --features FEATURE=VALUE ...
 
    Filter results to qubes that match all specified features. Omitted VALUE
-   means None (not set). "" or '' means blank
+   means None (unset). Empty value means "" or '' (blank)
 
 .. option:: --prefs PREFERENCE=VALUE ...
 
    Filter results to qubes that match all specified preferences. Omitted VALUE
-   means None (not set). "" or '' means blank
+   means None (unset). Empty value means "" or '' (blank)
 
 Sorting options
 ---------------
 
 .. option:: --sort COLUMN
 
-   Sort based on provided column rather than NAME. Sort key should be in the
+   Sort based on provided column rather than NAME. Sort key must be in the
    output columns
 
 .. option:: --reverse
@@ -178,7 +178,7 @@ Sorting options
 
 .. option:: --ignore-case
 
-   Ignore case distinctions for sorting
+   Ignore case distinctions when sorting
 
 Authors
 -------
--- a/qubesadmin/tests/tools/qvm_ls.py
+++ b/qubesadmin/tests/tools/qvm_ls.py
@@ -234,6 +234,14 @@ class TC_70_Tags(qubesadmin.tests.QubesTestCase):
         self.assertEqual(stdout.getvalue(),
             'NAME  STATE  CLASS  LABEL  TEMPLATE  NETVM\n')
 
+    def test_100_exclude_tag(self):
+        with qubesadmin.tests.tools.StdoutBuffer() as stdout:
+            qubesadmin.tools.qvm_ls.main(['--exclude-tags', 'not-my'], \
+                    app=self.app)
+        self.assertEqual(stdout.getvalue(),
+            'NAME  STATE    CLASS   LABEL  TEMPLATE  NETVM\n'
+            'dom0  Running  TestVM  black  -         -\n')
+
 
 class TC_80_Power_state_filters(qubesadmin.tests.QubesTestCase):
     def setUp(self):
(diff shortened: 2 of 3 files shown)


qubesadmin/tools/qvm_ls.py
```diff
@@ -330,7 +330,10 @@ def calc_used(vm, volume_name):
     return '{}%'.format(usage * 100 // size)
 
 
-# todo maxmem
+# Show hyphen if maxmmem is zero (for HVMs)
+Column('MAXMEM',
+    attr=(lambda vm: vm.maxmem if vm.virt_mode != "hvm" else '-'),
```
ben-grande: 1. Why use a symbol `-` rather than `0` itself?
2. It doesn't _Show hyphen if maxmmem is zero (for HVMs)_, but it does show hyphen if `virt_mode` is `hvm`.

qubesadmin/tools/qvm_ls.py
```diff
@@ -496,6 +511,8 @@ def write_table(self, stream=sys.stdout):
     'kernel': ('name', 'state', 'class', 'template', 'kernel', 'kernelopts'),
     'full': ('name', 'state', 'class', 'label', 'qid', 'xid', 'uuid'),
 #   'perf': ('name', 'state', 'cpu', 'memory'),
```
ben-grande, line 528: Makes sense to delete this comment as you added a substitute in the line below.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
                 wrapper.fill(', '.join(sorted(formats.keys()))),
                 wrapper.fill(', '.join(sorted(sorted(Column.columns.keys()))))))
 
-    parser.add_argument('--help-columns', action=_HelpColumnsAction)
-    parser.add_argument('--help-formats', action=_HelpFormatsAction)
+    parser_format = parser.add_argument_group(title='formatting options')
+    parser_format_group = parser_format.add_mutually_exclusive_group()
```
ben-grande: This name is better than `parser_formats`. Maybe use `parser_exclusive_format`. Nitpick.

alimirjamali:
Making pylint happy(er)

I am waiting for the CI/CD bot to advise on the required unit tests.
The Sphinx error could be resolved by updating documentation if the main change is approved. 
```
18:14:05,477 [executor:docker:b34f76404a9f] output: DEBUG: Sphinx error:
18:14:05,478 [executor:docker:b34f76404a9f] output: DEBUG: Undocumented arguments for command 'qvm-ls': '--based-on, --class, --conn-netvm, --features, --ignore-case, --internal, --label, --no-tags, --pending-update, --prefs, --reverse, --servicevm, --sort'
```

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
...
+    parser_filter.add_argument('--no-tags', nargs='+', metavar='TAG',
+        help='exclude VMs having specific tag(s)')
+
+    for pwstate in DOMAIN_POWER_STATES:
+        parser_filter.add_argument('--{}'.format(pwstate), action='store_true',
+        help='show {} VMs'.format(pwstate))
+
+    parser_filter.add_argument('--based-on', nargs='+',
+        metavar='TEMPLATE', action='store',
+        help='filter results to the AppVMs based on the TEMPLATE. '
```
ben-grande: DispVMs are not AppVMs have an AppVM as template, therefore don't use _AppVMs based on_, but use _qubes based on_.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
...
+        help='show only VMs having specific tag(s)')
+
+    parser_filter.add_argument('--no-tags', nargs='+', metavar='TAG',
+        help='exclude VMs having specific tag(s)')
+
+    for pwstate in DOMAIN_POWER_STATES:
+        parser_filter.add_argument('--{}'.format(pwstate), action='store_true',
+        help='show {} VMs'.format(pwstate))
+
+    parser_filter.add_argument('--based-on', nargs='+',
```
ben-grande: Name should use qubes nomenclature, `--template-source` maybe? Unsure, possibly a nitpick in this case, will let others step in.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
...
+        metavar='TEMPLATE', action='store',
+        help='filter results to the AppVMs based on the TEMPLATE. '
+             '"" means None')
+
+    parser_filter.add_argument('--conn-netvm', nargs='+',
+        metavar='NETVM', action='store',
+        help='filter results to the VMs connecting via NETVM')
```
ben-grande: Prefer. `qubes` instead of `VMs`.

From the [Qubes OS glossary](https://www.qubes-os.org/doc/glossary/#qube):

> A secure compartment in Qubes OS. Currently, qubes are implemented as Xen VMs, but Qubes OS is independent of its underlying compartmentalization technology. VMs could be replaced with a different technology, and qubes would still be called “qubes.”

I notice that there are existing csae of `VM` word being used, but please don't follow it.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
...
+    for pwstate in DOMAIN_POWER_STATES:
+        parser_filter.add_argument('--{}'.format(pwstate), action='store_true',
+        help='show {} VMs'.format(pwstate))
+
+    parser_filter.add_argument('--based-on', nargs='+',
+        metavar='TEMPLATE', action='store',
+        help='filter results to the AppVMs based on the TEMPLATE. '
+             '"" means None')
+
+    parser_filter.add_argument('--conn-netvm', nargs='+',
```
ben-grande: Please avoid abbreviations unless widely know. I know `conn` stands for connection, but not everybody may know.

It is from [Qubes OS website style guide](https://www.qubes-os.org/doc/website-style-guide/#general-practices), but I believe it applies to documentation also.

> Use descriptive variable names. Never use one or two letter variable names. Avoid obscure abbreviations and made-up words.

A better name would be `--netvm-is` to not conflict with a possible future option `--netvm`. Not that my name `--netvm-is` is great... but out of ideas now.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
...
+        metavar='NETVM', action='store',
+        help='filter results to the VMs connecting via NETVM')
+
+    parser_filter.add_argument('--internal', metavar='<y|n|both>',
+        default='both', action='store', choices=['y', 'yes', 'n', 'no', 'both'],
+        help='show only internal VMs or option to hide them. '
+             'default is showing both regular & internal VMs')
```
ben-grande: If default is showing both, why have a `both` option?

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
...
+
+    parser_filter.add_argument('--servicevm', metavar='<y|n|both>',
+        default='both', action='store', choices=['y', 'yes', 'n', 'no', 'both'],
```
ben-grande: Same for `both` here.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +611,115 @@ def get_parser():
...
+        help='filter results to VMs pending for update')
+
+    parser_filter.add_argument('--features', nargs='+', metavar='FEATURE=VALUE',
+        action='store',
+        help='filter results to VMs with all of specified features. '
+            'omitted VALUE means None. "" means blank')
+
+    parser_filter.add_argument('--prefs', nargs='+', metavar='PREFERENCE=VALUE',
+        action='store',
+        help='filter results to VMs with all of specified preferences. '
```
ben-grande: _to qubes that matches all specified_

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +782,109 @@ def main(args=None, app=None):
...
+        # filter only VMs to specific class(es)
+        domains = [d for d in domains if d.klass in args.klass]
+
+    if args.label:
+        # filter only VMs with specific label(s)
+        domains_labeled = []
+        spinner.show('Filtering based on labels...')
+        for dom in domains:
+            if dom.label.name in args.label:
+                domains_labeled.append(dom)
```
ben-grande, line 808: Why outside of list comprehension like in `args.klass`? Was this for the spinner updates?

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +782,109 @@ def main(args=None, app=None):
...
+        domains = [dom for dom in domains
+                   if not set(dom.tags).intersection(set(args.no_tags))]
+
+    if args.based_on:
+        # Filter only VMs based on specific TemplateVM
+        child_domains = []
+        spinner.show('Filtering results to VMs based on their template...')
+        for dom in domains:
+            if getattr(dom, 'template', '') in args.based_on:
+                child_domains.append(dom)
```
ben-grande: Why outside of list comprehension like in args.klass? Was this for the spinner updates?

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +782,109 @@ def main(args=None, app=None):
...
+        domains = child_domains
+        spinner.hide()
+
+    if args.conn_netvm:
+        # Filter only VMs connecting with specific netvm
+        domains_connecting = []
+        spinner.show('Filtering results to VMs based on their netvm...')
+        for dom in domains:
+            if getattr(dom, 'netvm', '') in args.conn_netvm:
+                domains_connecting.append(dom)
```
ben-grande: Why outside of list comprehension like in args.klass? Was this for the spinner updates?

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +782,109 @@ def main(args=None, app=None):
...
+    if args.servicevm in ['y', 'yes']:
+        domains = [d for d in domains if d.features.get('servicevm', None)
+                   in ['1', 'true', 'True']]
+    elif args.servicevm in ['n', 'no']:
+        domains = [d for d in domains if not d.features.get('servicevm', None)
+                   in ['1', 'true', 'True']]
+
+    if args.pending_update:
+        domains = [d for d in domains if
+                   d.features.get('updates-available', None)]
```
ben-grande, line 861: That is very nice, simple and nice to have.

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +786,109 @@ def main(args=None, app=None):
...
+                value = ''
+            domains = [d for d in domains if d.features.get(key, None) == value]
+
+    if args.prefs:
+        # Filter only VMs with specified preferences
+        for pref in args.prefs:
+            try:
+                key, value = pref.split('=', 1)
+            except ValueError:
+                parser.error("Invalid argument --prefs {}".format(pref))
```
ben-grande: Add a colon to separate the message from the argument: `Invalid argument: --prefs {}`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +786,109 @@ def main(args=None, app=None):
...
+            if not key:
+                parser.error("Invalid argument --prefs {}".format(pref))
```
ben-grande: Add a colon to separate the message from the argument: `Invalid argument: --prefs {}`

alimirjamali:
Making pylint even happier (complain on assignment of lambda to variable)

The CI/CD build failure logs are very cluttered and hard to understand. If it is because of sphinx complain, should be fixable by updating the documents (if the overall changes and new options are acceptable).

qubesadmin/tools/qvm_ls.py
```diff
@@ -479,6 +486,18 @@ def write_table(self, stream=sys.stdout):
                         table_data.append(self.get_row(vm))
                     except qubesadmin.exc.QubesVMNotFoundError:
                         continue
+                if self.sort_order in self.get_head():
+                    sort_index = self.get_head().index(self.sort_order)
```
ben-grande: Store `get_head()` output to a variable to avoid repeated calls that will always have the same value.

qubesadmin/tools/qvm_ls.py
```diff
@@ -479,6 +486,18 @@ def write_table(self, stream=sys.stdout):
...
+                    if self.ignore_case:
+                        table_data[1:] = \
+                            sorted(table_data[1:],
+                                   key=(lambda row: row[sort_index].upper()),
+                                   reverse=self.reverse_sort)
+                    else:
+                        table_data[1:] = \
+                            sorted(table_data[1:],
+                                   key=(lambda row: row[sort_index]),
+                                   reverse=self.reverse_sort)
```
ben-grande: Check for `ignore_case` inside the lambda, is this more readable?
```python3
table_data[1:] = sorted(table_data[1:],
    key=(lambda row: row[sort_index].upper()) if self.ignore_case else row[sort_index],
    reverse=self.reverse_sort)
```

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +615,115 @@ def get_parser():
...
+        help='show only VMs of specific class(es)')
+
+    parser_filter.add_argument('--label', nargs='+', metavar='LABEL',
+        action='store',
+        help='show only VMs with specific label(s)')
+
+    parser_filter.add_argument('--tags', nargs='+', metavar='TAG',
+        help='show only VMs having specific tag(s)')
+
+    parser_filter.add_argument('--no-tags', nargs='+', metavar='TAG',
```
ben-grande: I understand many unix tools has the negating option prefix `--no-something`, but it doesn't mean it will print qubes that have no tags, but qubes that don't have a certain tag. For this reason and because `qvm_ls` already has the `--exclude` option, please change `--no-tags` to `--exclude-tags`.

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +786,109 @@ def main(args=None, app=None):
...
+        for dom in domains:
+            if getattr(dom, 'template', '') in args.based_on:
+                child_domains.append(dom)
+            spinner.update()
+        domains = child_domains
+        spinner.hide()
+
+    if args.conn_netvm:
+        # Filter only VMs connecting with specific netvm
+        domains_connecting = []
```
ben-grande: The `domains_connecting` is not a good variable name, there can be multiple types of qube connections such as audiovm, netvm, guivm. Please use a name that has the connection name such as `domains_child_of_netvm` or something similar that has `netvm` in the name.

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +786,109 @@ def main(args=None, app=None):
...
+                key, value = pref.split('=', 1)
+            except ValueError:
+                parser.error("Invalid argument --prefs {}".format(pref))
+            if not key:
+                parser.error("Invalid argument --prefs {}".format(pref))
+            if value == '':
+                value = None
+            elif value in ['\'\'', '""']:
+                value = ''
+            domains = [d for d in domains if str(getattr(d, key, None))==value]
```
ben-grande: Missing space around test: `) == value`.

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +786,109 @@ def main(args=None, app=None):
...
+                key, value = feature.split('=', 1)
+            except ValueError:
+                parser.error("Invalid argument --features {}".format(feature))
+            if not key:
+                parser.error("Invalid argument --features {}".format(feature))
+            if value == '':
+                value = None
+            elif value in ['\'\'', '""']:
+                value = ''
+            domains = [d for d in domains if d.features.get(key, None) == value]
```
ben-grande, line 876: Have you benchmarked this? Same thing about `args.prefs` later on. It may be slow, but decreasingly slow as the `domains` supposedly gets smaller.

Validating the arguments in the for loop seems okay, but the features.get() is ran for every feature instead of once per feature key pair.

If the runtime is not bad, then may be okay to leave as it is.

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +786,109 @@ def main(args=None, app=None):
...
                    if set(dom.tags).intersection(set(args.tags))]
 
+    if args.no_tags:
+        # exclude VMs having at least one of the specified tags
+        domains = [dom for dom in domains
+                   if not set(dom.tags).intersection(set(args.no_tags))]
+
+    if args.based_on:
+        # Filter only VMs based on specific TemplateVM
+        child_domains = []
```
ben-grande: Please use a name that reference the it is child domain of a template. The function `tree_append_child()` uses `childs` variable for network children. Use something similar to `domains_child_of_template`. Maybe child is redundant in this name.

Also you use `child_domains` here and `domains_connecting` later. Use a consistent naming scheme, such as beginning always with `domains_` and referencing the connection type later.

ben-grande (changes requested):
Good work!

qubesadmin/tools/qvm_ls.py
alimirjamali: > 1. Why use a symbol `-` rather than `0` itself?

I also considered 'N/A'. I personally thought hyphen or N/A should imply as not-applicable as maxmem does not make sense for HVMs. There are similar cases for the other fields.

qubesadmin/tools/qvm_ls.py
alimirjamali: > Makes sense to delete this comment as you added a substitute in the line below.

They are different. One is perf(ormance). The other is pref(erence). A little bit confusion I guess. Maybe one should be made longer.

qubesadmin/tools/qvm_ls.py
alimirjamali: > This name is better than `parser_formats`. Maybe use `parser_exclusive_format`. Nitpick.

Makes very sense. I will apply it after receiving other feedback from others (maybe Marek).

qubesadmin/tools/qvm_ls.py
alimirjamali: Theoretically DispVMs are not directly based on a TemplateVM. But rather based on an AppVM which is transformed to be a DispVMTemplate. But Yes. You are right. I will change it _qubes based on_.

qubesadmin/tools/qvm_ls.py
alimirjamali: > Was this for the spinner updates?

Yes. Exactly. If qvm-ls is run outside of dom0, the policy checks comes into play and it could become slow. And it could make it slow. I  considered moving it to Table class. But there might be cases that a field is required for filtering but not in output.

qubesadmin/tools/qvm_ls.py
alimirjamali: Same as  above

qubesadmin/tools/qvm_ls.py
alimirjamali: same as above

qubesadmin/tools/qvm_ls.py
alimirjamali: > Missing space around test: `) == value`.

I tried to cheat here and avoid breaking the line above 80 characters. But yes, I should do it.

qubesadmin/tools/qvm_ls.py
alimirjamali: > Have you benchmarked this?

They are almost instant in dom0 on this HP Elitebook 820 G1 from Windows 7 era. But noticeable in other domains (as policy checks come into play). Not painful. Very much usable. Maybe I should consider adding the spinner for features and preferences filtering as well.

qubesadmin/tools/qvm_ls.py
marmarek: maxmem very much does make sense for HVM. It doesn't make sense for qubes with dynamic memory management disabled, which happen to be the case for the only HVMs in the default configuration.

alimirjamali:
TODo: Numeric sort wherever applicable. ~~Will be implemented after receiving further feedback from others.~~ Implemented.

qubesadmin/tools/qvm_ls.py
alimirjamali: I think it is better to remove maxmem custom implementation from this PR since it is not relevant to the original target of the PR.

qubesadmin/tools/qvm_ls.py
alimirjamali: It is only called again once (and not in a loop). I guess we could pass this

alimirjamali:
Ok. After updating the documentation, we are past CI/CD bot (all Green) and Codecov is advising on the required unit tests. I will write them next.

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+   Show only qubes of specific class(es)
 
-   No spinner today.
+.. option:: --label LABEL ...
+
+   Show only qubes with specific label(s)
+
+.. option:: --tags TAG ...
+
+   Shows only VMs having specific tag(s).
```
ben-grande: `s/VMs/qubes/`

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+
+.. option:: --exclude-tags TAG ...
+
+   Exclude VMs having specific tag(s).
```
ben-grande: `s/VMs/qubes/`

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+
+.. option:: --running, --paused, --halted
+
+   Shows only VMs matching the specified power state(s). When none of these
```
ben-grande: `s/VMs/qubes/`

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+   options is used (default), all VMs are shown.
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tests/tools/qvm_ls.py
```diff
@@ -355,3 +363,347 @@ def test_101_list_selected(self):
...
+
+class TC_100_Sort(qubesadmin.tests.QubesTestCase):
+    def setUp(self):
+        self.app = TestApp()
+        self.app.domains = TestVMCollection(
+            [
+                ('a', TestVM('a', label='red', maxmem='100')),
+                ('B', TestVM('B', label='green', maxmem='1000')),
+                ('c', TestVM('c', label='blue',  maxmem='300')),
+                ('dom0', TestVM('dom0', label='black', maxmem='-'))
```
ben-grande, line 375: Is this maxmem value `-` legacy from your previous implementation of returning `-` if it is `0`? I still prefer integer value.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +626,113 @@ def get_parser():
...
+        metavar='NETVM', action='store',
+        help='filter results to the qubes connecting via NETVM')
+
+    parser_filter.add_argument('--internal', metavar='<y|n>',
+        default='both', action='store', choices=['y', 'yes', 'n', 'no'],
+        help='show only internal qubes or option to hide them')
+
+    parser_filter.add_argument('--servicevm', metavar='<y|n>',
+        default='both', action='store', choices=['y', 'yes', 'n', 'no'],
+        help='show only Service VMs or option to hide them')
```
ben-grande: `s/Service VMs/ServiceVMs/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +626,113 @@ def get_parser():
...
+        help='show {} VMs'.format(pwstate))
+
+    parser_filter.add_argument('--template-source', nargs='+',
+        metavar='TEMPLATE', action='store',
+        help='filter results to the qubes based on the TEMPLATE. '
+             '"" means None')
+
+    parser_filter.add_argument('--netvm-is', nargs='+',
+        metavar='NETVM', action='store',
+        help='filter results to the qubes connecting via NETVM')
```
ben-grande, line 694: The template filter above has:

> filter results to the qubes based on the TEMPLATE.

So for uniformity, I propose:

> filter results to the qubes based on the NETVM.

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +626,113 @@ def get_parser():
...
+    parser_sort = parser.add_argument_group(title='sorting options')
+
+    parser_sort.add_argument('--sort', metavar='COLUMN', action='store',
+        default='NAME', help='sort based on provided column rather than NAME')
+
+    parser_sort.add_argument('--reverse', action='store_true', default=False,
+        help='reverse sort')
+
+    parser_sort.add_argument('--ignore-case', action='store_true',
+        default=False, help='ignore case distinctions for sorting')
```
ben-grande: `s/for sorting/when sorting/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
             vm for vm in args.app.domains if vm.name not in args.exclude
         ]
 
+    if args.klass:
+        # filter only VMs to specific class(es)
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
...
+        domains = [d for d in domains if d.klass in args.klass]
+
+    if args.label:
+        # filter only VMs with specific label(s)
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
...
+        spinner.show('Filtering based on labels...')
+        for dom in domains:
+            if dom.label.name in args.label:
+                domains_labeled.append(dom)
+            spinner.update()
+        domains = domains_labeled
+        spinner.hide()
+
     if args.tags:
         # filter only VMs having at least one of the specified tags
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
...
         domains = [dom for dom in domains
                    if set(dom.tags).intersection(set(args.tags))]
 
+    if args.exclude_tags:
+        # exclude VMs having at least one of the specified tags
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
...
+        domains = [dom for dom in domains
+                   if not set(dom.tags).intersection(set(args.exclude_tags))]
+
+    if args.template_source:
+        # Filter only VMs based on specific TemplateVM
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
...
+        spinner.show('Filtering results to qubes based on their templates...')
+        for dom in domains:
+            if getattr(dom, 'template', '') in args.template_source:
+                domains_template.append(dom)
+            spinner.update()
+        domains = domains_template
+        spinner.hide()
+
+    if args.netvm_is:
+        # Filter only VMs connecting with specific netvm
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
...
+    elif args.servicevm in ['n', 'no']:
+        domains = [d for d in domains if not d.features.get('servicevm', None)
+                   in ['1', 'true', 'True']]
+
+    if args.pending_update:
+        domains = [d for d in domains if
+                   d.features.get('updates-available', None)]
+
+    if args.features:
+        # Filter only VMs with specified features
```
ben-grande: `s/VMs/qubes/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -710,16 +795,110 @@ def main(args=None, app=None):
...
+            if not key:
+                parser.error("Invalid argument: --features {}".format(feature))
+            if value == '':
+                value = None
+            elif value in ['\'\'', '""']:
+                value = ''
+            domains = [d for d in domains if d.features.get(key, None) == value]
+
+    if args.prefs:
+        # Filter only VMs with specified preferences
```
ben-grande: `s/VMs/qubes/`

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+   Show only servicevms or exclude them from output
+
+.. option:: --pending-update
+
+   Filter results to qubes pending for update
+
+.. option:: --features FEATURE=VALUE ...
+
+   Filter results to qubes that match all specified features. Omitted VALUE
+   means None (not set). "" or '' means blank
```
ben-grande: > Omitted VALUE means None (not set). "" or '' means blank

What about:

_Omitted VALUE means None (unset). Empty value means "" or '' (blank)_

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+
+.. option:: --prefs PREFERENCE=VALUE ...
+
+   Filter results to qubes that match all specified preferences. Omitted VALUE
+   means None (not set). "" or '' means blank
```
ben-grande: > Omitted VALUE means None (not set). "" or '' means blank

What about:

_Omitted VALUE means None (unset). Empty value means "" or '' (blank)_

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+
+Sorting options
+---------------
+
+.. option:: --sort COLUMN
+
+   Sort based on provided column rather than NAME. Sort key should be in the
```
ben-grande: `s/should/must/` if it is required.

doc/manpages/qvm-ls.rst
```diff
@@ -79,21 +102,83 @@ Options
...
+   output columns
+
+.. option:: --reverse
+
+   Reverse sort
+
+.. option:: --ignore-case
+
+   Ignore case distinctions for sorting
```
ben-grande: `s/for sorting/when sorting/`

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +626,113 @@ def get_parser():
...
+        help='exclude qubes having specific tag(s)')
+
+    for pwstate in DOMAIN_POWER_STATES:
+        parser_filter.add_argument('--{}'.format(pwstate), action='store_true',
+        help='show {} VMs'.format(pwstate))
+
+    parser_filter.add_argument('--template-source', nargs='+',
+        metavar='TEMPLATE', action='store',
+        help='filter results to the qubes based on the TEMPLATE. '
+             '"" means None')
```
ben-grande, line 690: > Omitted VALUE means None (not set). "" or '' means blank

What about:

_Omitted VALUE means None (unset). Empty value means "" or '' (blank)_

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +626,113 @@ def get_parser():
...
+        default='both', action='store', choices=['y', 'yes', 'n', 'no'],
+        help='show only Service VMs or option to hide them')
+
+    parser_filter.add_argument('--pending-update', action='store_true',
+        help='filter results to qubes pending for update')
+
+    parser_filter.add_argument('--features', nargs='+', metavar='FEATURE=VALUE',
+        action='store',
+        help='filter results to qubes that matches all specified features. '
+            'omitted VALUE means None. "" means blank')
```
ben-grande: > Omitted VALUE means None (not set). "" or '' means blank

What about:

_Omitted VALUE means None (unset). Empty value means "" or '' (blank)_

qubesadmin/tools/qvm_ls.py
```diff
@@ -594,60 +626,113 @@ def get_parser():
...
+
+    parser_filter.add_argument('--prefs', nargs='+', metavar='PREFERENCE=VALUE',
+        action='store',
+        help='filter results to qubes that matches all specified preferences. '
+            'omitted VALUE means None. "" means blank')
```
ben-grande: > Omitted VALUE means None (not set). "" or '' means blank

What about:

_Omitted VALUE means None (unset). Empty value means "" or '' (blank)_

doc/manpages/qvm-ls.rst
```diff
@@ -43,14 +75,10 @@ Options
...
-.. option:: --tags TAG ...
-
-   Shows only VMs having specific tag(s).
-
-.. option:: --running, --paused, --halted
+.. option:: --tree, -t
 
-   Shows only VMs matching the specified power state(s). When none of these
-   options is used (default), all VMs are shown.
+   List domains as a network tree. Domains are sorted as they are connected to
```
ben-grande: `s/domains/qubes/`, `s/Domains/qubes/`

ben-grande (changes requested):
Documentation review.

marmarek (approved):
Looks okay, but just in case lets run it through openqa.

qubesadmin/tests/tools/qvm_ls.py
alimirjamali: No. It is the current existing condition. Not related to my previous (or current) implementation. This is only the unit-test BTW. I had to assure that the numeric sort works the `-`

qubesadmin/tools/qvm_ls.py
alimirjamali: This one is tricky. A template name could not be blank. This is a case that we could have only None.

alimirjamali:
A supplementary commit to implement documentation & commenting improvements based on @ben-grande suggestions

PR #302, opened 2024-07-27
Hide qvm-prefs default value with --hide-default

### LINKED ISSUE #8615: qvm-prefs does not hide defaults if key is specified

### Qubes OS release

R4.1

### Brief summary

When specifying `--hide-default` it conflicts with the key specified, printing incorrect information.

### Steps to reproduce

Try to see if qube netvm is set to default or not (the netvm of this qube is the default):
```sh
qvm-prefs QUBE | grep "^netvm\s\+D\s\+" # exit 0 - prints the whole line
qvm-prefs QUBE | grep "^netvm\s\+D\s\+" | awk '{print $3}' # exit 0 - prints netvm but does not fail on pipes depeding on shell options
qvm-prefs --hide-default QUBE | grep netvm   # exit 1 - prints nothing, expected
qvm-prefs --hide-default QUBE netvm  # exit 0 - prints netvm, should not print
qvm-prefs QUBE netvm # exit 0 - prints netvm, this is okay
```
So... how can one know that the netvm of a qube is using the default value for netvm without parsing the output?

### Expected behavior

```sh
qvm-prefs --hide-default QUBE netvm
```
Should fail if netvm is the default.

### Actual behavior

```sh
qvm-prefs --hide-default QUBE netvm
```
Prints the netvm name, even though it is the default value.

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8615

### DIFF
--- a/qubesadmin/tools/qvm_prefs.py
+++ b/qubesadmin/tools/qvm_prefs.py
@@ -141,9 +141,12 @@ def process_actions(parser, args, target):
         return 0
 
     try:
-        value = getattr(target, args.property)
-        if value is not None:
-            print(str(value))
+        if args.hide_default and target.property_is_default(args.property):
+            return 0
+        else:
+            value = getattr(target, args.property)
+            if value is not None:
+                print(str(value))
     except qubesadmin.exc.QubesNoSuchPropertyError:
         parser.error('no such property: {!r}'.format(args.property))
     except qubesadmin.exc.QubesException as e:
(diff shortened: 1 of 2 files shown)


qubesadmin/tools/qvm_prefs.py
```diff
@@ -141,9 +141,12 @@ def process_actions(parser, args, target):
         return 0
 
     try:
-        value = getattr(target, args.property)
-        if value is not None:
-            print(str(value))
+        if target.property_is_default(args.property) and args.hide_default:
```
marmarek: Better swap order, to save one admin api call in the common case.

marmarek (changes requested):

qubesadmin/tools/qvm_prefs.py
alimirjamali: Done

marmarek (approved):

PR #305, opened 2024-08-21
Make qvm-backup --compress & --no-compress mutual

### LINKED ISSUE #1887: backups: use compression_filter=None for uncompressed backups

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/1887


rustybird:
It's a common CLI idiom though to allow opposing flags, and use the last one that was passed. That way the user can set

    alias mybackup='qvm-backup ... --no-compress ...'

and occasionally override the flag with

    mybackup --compress

alimirjamali:
> It's a common CLI idiom though to allow opposing flags,

Very well said. I believe I have to close this PR.

Deciding on disabling compression if filter is set to `None` or `None` has to happen in `qubes-core-admin` repository. Not here.

PR #306, opened 2024-09-01
Add PRIV-POOL & ROOT-POOL columns to qvm-ls

### LINKED ISSUE #5138: Add POOL as an available column in qvm-ls
Quickly find which LVM pool a VM belongs to.

Add a column to the supported --fields list that qvm-ls supports.

Any user who uses multiple pools should have a quick way to query which VMs are located on which pool.

**Describe alternatives you've considered**
Both `qvm-volume list` and `sudo lvs` can indirectly indicate the pool.

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/5138


alimirjamali:
Reason for force-push: Fixing a minor typo

marmarek (approved):

PR #308, opened 2024-09-15
Support DNF5 UpdateVM for qvm-template

alimirjamali:
Related: https://github.com/QubesOS/qubes-issues/issues/9244

### DIFF
--- a/qubesadmin/tools/qvm_template.py
+++ b/qubesadmin/tools/qvm_template.py
@@ -550,6 +550,9 @@ def qrexec_repoquery(
         # This is because if .strip() is used, the .split() will not work.
         if line == '':
             continue
+        # This is specific to DNF5:
+        if line.endswith('|'):
+            line = line[:-1]
         entry = line.split('|')
         try:
             # If there is an incorrect number of entries, raise an error
@@ -573,9 +576,14 @@ def qrexec_repoquery(
                 raise ValueError
             dlsize = int(dlsize)
             # First verify that the date does not look weird, then parse it
-            if not re.fullmatch(date_re, buildtime):
+            if re.fullmatch(date_re, buildtime):
+                buildtime = datetime.datetime.strptime(buildtime, \
+                        '%Y-%m-%d %H:%M')
+            elif buildtime.isnumeric():
+                # DNF5 provides seconds since epoch
+                buildtime = datetime.datetime.fromtimestamp(int(buildtime))
+            else:
                 raise ValueError
-            buildtime = datetime.datetime.strptime(buildtime, '%Y-%m-%d %H:%M')
             # XXX: Perhaps whitelist licenses directly?
             if not re.fullmatch(licence_re, licence):
                 raise ValueError
--- a/qubesadmin/tests/tools/qvm_template.py
+++ b/qubesadmin/tests/tools/qvm_template.py
@@ -1426,6 +1426,43 @@ qubes-template-fedora-32|1|4.2|20200201|qubes-templates-itl-testing|2048576|2020
         self.assertAllCalled()
 
     @mock.patch('qubesadmin.tools.qvm_template.qrexec_payload')
+    def test_120_b_qrexec_repoquery_success_dnf5(self, mock_payload):
+        args = argparse.Namespace(updatevm='test-vm')
+        mock_payload.return_value = 'str1\nstr2'
+        self.app.expected_calls[('dom0', 'admin.vm.List', None, None)] = \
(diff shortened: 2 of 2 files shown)


alimirjamali:
Making pylint happy(er)

alimirjamali:
It did take some time to write the proper unittest but all checks are green now. This is ready for review

qubesadmin/tools/qvm_template.py
```diff
@@ -539,7 +539,7 @@ def qrexec_repoquery(
         for line in stderr.rstrip().split('\n'):
             print(f"[Qrexec] {line}", file=sys.stderr)
         raise ConnectionError("qrexec call 'qubes.TemplateSearch' failed.")
-    name_re = re.compile(r'\A[A-Za-z0-9._+][A-Za-z0-9._+-]*\Z')
+    name_re = re.compile(r'\A[A-Za-z0-9._+][A-Za-z0-9. _+-]*\Z')
```
marmarek: I haven't checked other places yet, but space is not a valid character in a template name. Is it leading/trailing space? If so, maybe `strip()` call is needed somewhere?

alimirjamali: > I haven't checked other places yet, but space is not a valid character in a template name. Is it leading/trailing space? If so, maybe `strip()` call is needed somewhere?

It is neither leading nor trailing space. The unittest has the actual output data from DNF5. It was:
`Qubes Templates repository`

marmarek: Ok, so this is repository name, not a template name. The `qubes.TemplateSearch ` should have repository id here, not its name (I see variable is named `reponame`, but that's not accurate). So, it either needs an option for dnf to print repo id, or a mapping to translate repo name to repo id. This change belongs to the core-agent-linux PR.

marmarek (changes requested):

qubesadmin/tools/qvm_template.py
alimirjamali: > Ok, so this is repository name, not a template name. The `qubes.TemplateSearch ` should have repository id here, not its name (I see variable is named `reponame`, but that's not accurate). So, it either needs an option for dnf to print repo id, or a mapping to translate repo name to repo id. This change belongs to the core-agent-linux PR.

Ok. I will remove space from the Regex. I am not certain if an option could be found to behave the same data in both DNF4 & DNF5. The [current existing key](https://github.com/QubesOS/qubes-core-agent-linux/blob/main/qubes-rpc/qvm-template-repo-query#L52C76-L52C88) in `qvm-template-repo-query` is `%{reponame}` which behaves different in DNF4 and DNF5.

marmarek: Maybe there is `%{repoid}` or similar that would work on DNF5? (or ideally also with DNF4?) it can use different options for DNF4 and DNF5

alimirjamali: > Maybe there is `%{repoid}` or similar that would work on DNF5? (or ideally also with DNF4?) it can use different options for DNF4 and DNF5

Just checked it. `%{repoid}` is valid for both DNF4 and DNF5. Fix for `core-agent-linux` here:

https://github.com/QubesOS/qubes-core-agent-linux/pull/522

And this PR is also updated

qubesadmin/tools/qvm_template.py
```diff
@@ -573,9 +576,14 @@ def qrexec_repoquery(
                 raise ValueError
             dlsize = int(dlsize)
             # First verify that the date does not look weird, then parse it
-            if not re.fullmatch(date_re, buildtime):
+            if re.fullmatch(date_re, buildtime):
+                buildtime = datetime.datetime.strptime(buildtime, \
+                        '%Y-%m-%d %H:%M')
+            elif buildtime.isnumeric():
+                # DNF5 provides seconds since epoch
+                buildtime = datetime.datetime.fromtimestamp(int(buildtime))
```
marmarek, line 584: Going further - can DNF5 be convinced to the old timestamp format?

alimirjamali: > Going further - can DNF5 be convinced to the old timestamp format?

This appears to be impossible. There is only `%{buildtime}`. Help says: `Display buildtime of the package in Unix time`.

marmarek: I see... I don't like that cross-VM data format is changing here, but since the new one is easier to handle, I'm going to accept it.

marmarek (approved):

PR #311, opened 2024-10-14
Apply "fix: do not crash on pool with no volumes"

### LINKED ISSUE #9504: Do not crash on pool with no volumes

Do not crash on pool with no volumes

### The solution you'd like

```
From 2448b386505af8540f21b0f81275f45776070f8c Mon Sep 17 00:00:00 2001
From: Yaroslav Bolyukin <iam@lach.pw>
Date: Sat, 14 Sep 2024 00:09:19 +0200
Subject: [PATCH 3/3] fix: do not crash on pool with no volumes

---
 qubesadmin/storage.py | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/qubesadmin/storage.py b/qubesadmin/storage.py
index 27f5696..99e8d81 100644
--- a/qubesadmin/storage.py
+++ b/qubesadmin/storage.py
@@ -430,6 +430,8 @@ class Pool(object):
                 'dom0', 'admin.pool.volume.List', self.name, None)
         except qubesadmin.exc.QubesDaemonAccessError:
             raise qubesadmin.exc.QubesPropertyAccessError('volumes')
+        if volumes_data == b'':
+            return
         assert volumes_data.endswith(b'\n')
         volumes_data = volumes_data[:-1].decode('ascii')
         for vid in volumes_data.splitlines():
-- 
2.45.2
```

This is not an issue on QubesOS, because in qubes you have OOTB setup, and pools are never empty.

This is not the case on NixOS as dom0, on Nix you need to install distributions after
https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/storage.py

rustybird:
Submitted by @ethnh
Authored by @CertainLach
Fixes QubesOS/qubes-issues#9504

I don't know what exactly was crashing there. But the patch LGTM, and I've checked that `list(qubesadmin.Qubes().pools['my-empty-pool'].volumes)` now correctly results in `[]` instead of raising `AssertionError`.

### DIFF
--- a/qubesadmin/storage.py
+++ b/qubesadmin/storage.py
@@ -430,6 +430,8 @@ class Pool(object):
                 'dom0', 'admin.pool.volume.List', self.name, None)
         except qubesadmin.exc.QubesDaemonAccessError:
             raise qubesadmin.exc.QubesPropertyAccessError('volumes')
+        if volumes_data == b'':
+            return
         assert volumes_data.endswith(b'\n')
         volumes_data = volumes_data[:-1].decode('ascii')
         for vid in volumes_data.splitlines():


CertainLach (approved):
Qubesd and qubes-qube-manager are not handling this AssertionError, resulting in crash.

I think this error might be specific to nixos-qubes bootstrapping, where the initial system doesn't even have kernels available, but still this AssertionError is not good

rustybird:
Makes sense that the crash was in qubesmanager: That was the only consumer of the buggy method I found too, and limited to kernel pools. (`qvm-volume list --pool=...` calls a different method)

marmarek (approved):

PR #312, opened 2024-10-14
Add UUID column to `qvm-ls`

alimirjamali:
related: https://github.com/QubesOS/qubes-issues/issues/8862


marmarek:
You already can use any VM property as a column, so this change isn't necessary.

PR #313, opened 2024-10-15
qvm-start-daemon: Pass KDE specific guid arguments for stubdomain too

HW42:
Needs backport to 4.2 (actually I only tested it there)


CertainLach (approved):
Works on NixOS-Qubes 4.3

marmarek (approved):

PR #314, opened 2024-10-22
Add UNIX wildcards support

alimirjamali:
Allowing user to select qubes with UNIX like wildcards for multiple qube selection

### DIFF
--- a/qubesadmin/tools/__init__.py
+++ b/qubesadmin/tools/__init__.py
@@ -24,6 +24,7 @@
 from __future__ import print_function
 
 import argparse
+import fnmatch
 import importlib
 import logging
 import os
@@ -37,7 +38,6 @@ import qubesadmin.vm
 
 #: constant returned when some action should be performed on all qubes
 VM_ALL = object()
-
 
 class QubesAction(argparse.Action):
     ''' Interface providing a convinience method to be called, after
@@ -151,6 +151,8 @@ class VmNameAction(QubesAction):
         setattr(namespace, self.dest, values)
 
     def parse_qubes_app(self, parser, namespace):
+        ''' Set ``namespace.domains`` to ``values`` '''
+        # pylint: disable=too-many-nested-blocks
         assert hasattr(namespace, 'app')
         setattr(namespace, 'domains', [])
         app = namespace.app
@@ -175,7 +177,16 @@ class VmNameAction(QubesAction):
                     except KeyError:
                         parser.error('no such domain: {!r}'.format(vm_name))
             else:
-                for vm_name in getattr(namespace, self.dest):
+                destinations = set()
+                for destination in getattr(namespace, self.dest):
+                    if any(wildcard in destination for wildcard in '*?[!]'):
+                        for domain in app.domains:
+                            if fnmatch.fnmatch(domain.name, destination):
+                                destinations.add(domain.name)
+                    else:
+                        destinations.add(destination)
+
+                for vm_name in destinations:
                     try:
                         namespace.domains += [app.domains[vm_name]]
                     except KeyError:
--- a/qubesadmin/tests/tools/qvm_ls.py
+++ b/qubesadmin/tests/tools/qvm_ls.py
@@ -129,6 +129,46 @@ class TC_50_List(qubesadmin.tests.QubesTestCase):
             'dom0     Running  TestVM  -      -         -\n'
             'test-vm  Running  TestVM  -      -         -\n')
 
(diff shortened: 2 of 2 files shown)


alimirjamali:
CI/CD tests are all green. I shall proceed with the unittests if the overall patch is approved.

qubesadmin/tools/__init__.py
```diff
@@ -162,6 +166,12 @@ def parse_qubes_app(self, parser, namespace):
                    vm.name not in namespace.exclude
             ]
         else:
+            def uuid_to_domain(uuid: str):
+                for domain in app.domains:
+                    if domain.uuid == uuid:
+                        return domain
```
marmarek: With https://github.com/QubesOS/qubes-core-admin/pull/574, the `uuid:...` syntax can be used as Admin API call target. This means, it should be relatively easy to get `app.domains["uuid:...."]` working, which is much better option than fetching UUID of all qubes and iterating client-side.
Technical hint: `get_blind()` won't work with just UUID, because the QubesVM object needs a name to be constructed. But that's okay, just skip the blind mode for getting VM via UUID.

qubesadmin/tools/__init__.py
```diff
@@ -171,11 +181,25 @@ def parse_qubes_app(self, parser, namespace):
...
+                        if re.match(UUID_REGEX, vm_name):
+                            namespace.domains += [uuid_to_domain(vm_name[5:])]
+                        else:
+                            namespace.domains += [app.domains[vm_name]]
                     except KeyError:
                         parser.error('no such domain: {!r}'.format(vm_name))
             else:
-                for vm_name in getattr(namespace, self.dest):
+                destinations=set()
+                for destination in getattr(namespace, self.dest, None):
```
marmarek: `None` is not iterable, so it isn't good for default value.

qubesadmin/tools/__init__.py
```diff
@@ -171,11 +181,25 @@ def parse_qubes_app(self, parser, namespace):
...
+                    if any(wildcard in destination for wildcard in '*![]'):
+                        for domain in app.domains:
+                            if fnmatch.fnmatch(domain.name, destination):
+                                destinations.add(domain.name)
+                    elif re.match(UUID_REGEX, destination):
```
marmarek: With UUID-aware VMCollection, this case won't be necessary

marmarek (changes requested):

qubesadmin/tools/__init__.py
alimirjamali: > This means, it should be relatively easy to get `app.domains["uuid:...."]` working

I had tried this before going to iteration path. It does not work.

> Technical hint: `get_blind()` won't work with just UUID

This actually works. But the QubesVM object is constructed with "uuid:..." for its name rather than the original vmname.

qubesadmin/tools/__init__.py
alimirjamali: overall it is necessary to find out why `qubesadmin.Qubes().domains["uuid:..."]` does not work. Neither `__getitem__` nor `get_blind` could not translate `uuid:...`

qubesadmin/tools/__init__.py
alimirjamali: @marmarek I just looked at `qubesadmin/app.py` on my system and recognized that it is v4.3.6. It is the latest [python3-qubesadmin-4.3.6-1.fc41.noarch.rpm](https://yum.qubes-os.org/r4.3/current-testing/dom0/fc41/rpm/python3-qubesadmin-4.3.6-1.fc41.noarch.rpm) in testing repositories. So the UUID changes I need is not yet released?

marmarek: UUID support is on the backend (core-admin, not core-admin-client). It's in qubes-core-dom0 4.3.9

alimirjamali: > UUID support is on the backend (core-admin, not core-admin-client). It's in qubes-core-dom0 4.3.9

Yes. That is correct. My mistake.

I gave up on adding UUID  support to VMCollection. I readjusted the PR, commit message and comments just for UNIX Wildcard support. It is still useful IMO. Users will be able to `qvm-kill disp*`, `qvm-ls *minimal` and similar tasks.

marmarek:
yes, looks good, please add a test :)

alimirjamali:
> yes, looks good, please add a test :)

It is done. All 4 CI/CD checks are green.

marmarek (approved):

PR #315, opened 2024-10-30
Ignore invalid GUI related features and options

### LINKED ISSUE #7730: GUI-related feature flags not validated

### Qubes OS release

R4.1

### Brief summary

Various GUI-related features are not validated by qubesd.  Once https://github.com/QubesOS/qubes-core-admin-client/pull/221 lands, this will cause qvm-start-daemon to produce an invalid GUI daemon configuration file, preventing the GUI daemon from starting.

### Steps to reproduce

Have https://github.com/QubesOS/qubes-core-admin-client/pull/221 and set `gui-subwindows` feature to something invalid.

### Expected behavior

Feature set call is rejected by qubesd.  Attempts to set unknown `gui-*` features should likely be rejected unconditionally, as these could acquire a meaning in the future.

### Actual behavior

Feature set call is not rejected by qubesd.

alimirjamali:
resolves: https://github.com/QubesOS/qubes-issues/issues/7730

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -28,6 +28,7 @@ Depends:
  qubes-repo-templates,
  scrypt,
  qubes-rpm-oxide (>= 0.2.3),
+ python3-xlib,
  ${python:Depends},
  ${python3:Depends},
  ${misc:Depends},
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -39,22 +39,77 @@ import qubesadmin.tools
 import qubesadmin.vm
 from qubesadmin.tools import xcffibhelpers
 
+# pylint: disable=wrong-import-position,wrong-import-order
+from Xlib import X, XK
+from Xlib.display import Display
+from Xlib.error import DisplayConnectionError
+
 GUI_DAEMON_PATH = '/usr/bin/qubes-guid'
 PACAT_DAEMON_PATH = '/usr/bin/pacat-simple-vchan'
 QUBES_ICON_DIR = '/usr/share/icons/hicolor/128x128/devices'
 
+def validator_key_sequence(sequence: str) -> bool:
+    """ xside.c key sequence validation is not case sensitive and supports more
+    choices than Global Config's limited choices, so we replicate it here """
+    if not isinstance(sequence, str):
+        return False
+    if sequence.lower() in ['none', 'disable']:
+        return True
+    while '-' in sequence:
+        modifier, sequence = sequence.split('-', 1)
+        if not modifier.lower() in \
+                ['shift', 'ctrl', 'alt', 'mod1', 'mod2', 'mod3', 'mod4']:
+            return False
+    try:
+        display = Display()
+        result = display.keysym_to_keycode(XK.string_to_keysym(sequence))
+        display.close()
+        return result != X.NoSymbol
+    except DisplayConnectionError:
+        return False
+
+def validator_trayicon_mode(mode: str) -> bool:
+    """ xside.c tray mode validation is replicated here """
+    if not isinstance(mode, str):
+        return False
+    if mode in ['bg', 'border1', 'border2', 'tint']:
+        return True
+    if mode.startswith('tint'):
+        if mode[4:] in ['+border1', '+border2', '+saturation50', '+whitehack']:
+            return True
+    return False
+
+def validator_color(color: str) -> bool:
+    """ xside.c `parse_color` validation code is replicated here """
+    if not isinstance(color, str):
+        return False
+    if re.match(r"^0[xX](?:[0-9a-fA-F]{3}){1,2}$", color.strip()):
+        # Technically `parse_color` could parse values such as `0x0`
+        # Xlib.xobject.colormap conventions & standards are different.
+        return True
+    try:
+        display = Display()
+        result = display.screen().default_colormap.alloc_named_color(color)
+        display.close()
+        if result is not None:
+            return True
+    except DisplayConnectionError:
+        return False
+    return False
+
 GUI_DAEMON_OPTIONS = [
-    ('allow_fullscreen', 'bool'),
-    ('override_redirect_protection', 'bool'),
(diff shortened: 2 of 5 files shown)


alimirjamali:
I am trying to add `python3-xlib` for CI/CD tests. Isn't it enough to have it in `rpm_spec`? Should it be added to `.gitlab-ci.yml`?

alimirjamali:
@marmarek apologies for pinging you here. I need to know how to include `python-xlib` for pylint and unit tests. Doing this via rpm_spec is not sufficient. Should it be done via `.gitlab-ci.yml` or `ci/requirements`?

marmarek:
> I am trying to add `python3-xlib` for CI/CD tests. Isn't it enough to have it in `rpm_spec`? Should it be added to `.gitlab-ci.yml`?

Unit tests do not use the spec file. For installing system packages add them to .gitlab-ci.yml.

alimirjamali:
I am marking this PR as ready for review. `test_013_common_args_guid_config` fails since `Xlib.display.Display()` can not connect to a valid display which results in `DisplayConnectionError`. Subsequently the validation code can not validate `Ctrl-Alt-Shift-c` copy sequence and it is not added to guid config. this does not happen in production.

marmarek:
> `Xlib.display.Display()` can not connect to a valid display which results in `DisplayConnectionError`

Try adding xvfb-run (see desktop-linux-manager)

qubesadmin/tools/qvm_start_daemon.py
```diff
@@ -39,22 +39,75 @@
 import qubesadmin.vm
 from qubesadmin.tools import xcffibhelpers
 
+# pylint: disable=wrong-import-position,wrong-import-order
+from Xlib import X, XK
+from Xlib.display import Display
+from Xlib.error import DisplayConnectionError
```
marmarek, line 45: Two issues here:
1. I believe nowadays it should be "xcffib" - "xcb" is preferred over "xlib" directly, and its python bindings are in python3-xcffib
2. You forgot to add dependency to the debian package, so qvm-start-daemon fails on all Debian debian tests.

marmarek (changes requested):

qubesadmin/tools/qvm_start_daemon.py
alimirjamali: > 1. I believe nowadays it should be "xcffib" - "xcb" is preferred over "xlib" directly, and its python bindings are in python3-xcffib

I have to study xcffib documentation as it appears to be a drop-in replacement for xcb rather than xlib. It does not provide `alloc_named_color` and `string_to_keysym` functions (or I can not find it).

> 2\. You forgot to add dependency to the debian package, so qvm-start-daemon fails on all Debian debian tests.

OK. I will do it

marmarek: If there are no easy equivalents, then I guess it may stay with python3-xlib, but check first.

marmarek: https://stackoverflow.com/a/50539803
> (Except that it only shows keysym values, not keysym names, which you can find in /usr/include/X11/keysymdef.h or using Xlib's XKeysymToString(), and I don't think there exists an XCB port of that function, but you could write your own based on keysymdef.h.)

debian/control
```diff
@@ -28,6 +28,7 @@ Depends:
  qubes-repo-templates,
  scrypt,
  qubes-rpm-oxide (>= 0.2.3),
+ python3-xlib
```
marmarek: missing coma

marmarek (changes requested):

qubesadmin/tools/qvm_start_daemon.py
alimirjamali: > https://stackoverflow.com/a/50539803

As Geoffrey comments in the Stackoverflow, his suggested code does not show keysym names (only syms & values). And suggests `XKeysymToString` (which I am using it's python-xlib equivalent here).

For the keyboard, It might be possible to revert to [python3-xkbcommon](https://github.com/sde1000/python-xkbcommon) which is Python bindings for libxkbcommon using cffi. And what is used nowadays with [Wayland](https://wayland-book.com/seat/xkb.html) as well.

For the color, I have to look for alternatives

alimirjamali: @marmarek there are positive and negative things about `python3-xkbcommon`. I tested it and it works well. While it is the new modern approach and what is currently used with both X & Wayland, It is only available for Fedora 38 onward. Not Fedora 37. So it is time to decide if this patch is going to come only to Qubes R4.3 or R4.2 as well.

alimirjamali: For color, using `xcffib.xproto.xprotoExtension().AllocNamedColor()` works. But not as good as the Xlib's `alloc_named_color`. Since Xlib supports `#RGB`, `#RRGGBB`, `rgb:RRRRGGGGBBBB`, etc (as well as the named colors).

marmarek: > It is only available for Fedora 38 onward.

It also isn't available in Debian... So, I guess python3-xlib it is.

alimirjamali: > It also isn't available in Debian... So, I guess python3-xlib it is.

During the early stages of writing this PR, I used higher level GTK 4.0 for Keysym name validation ([reference](https://docs.gtk.org/gdk4/func.keyval_from_name.html)). It is still possible. And the ugly approach of `try import xcffib except import xlib` is also possible. Still your decision.

marmarek:
See https://github.com/QubesOS/qubes-issues/issues/9564, you made it too good ;)

alimirjamali:
> See [QubesOS/qubes-issues#9564](https://github.com/QubesOS/qubes-issues/issues/9564), you made it too good ;)

This should be fixed now. I just added this:
```python
    if sequence.lower() in ['none', 'disable']:                                                                                     
        return True 
```

marmarek (approved):

PR #316, opened 2024-11-03
Dedicate `x-` prefix for user's custom features

### LINKED ISSUE #9557: Provide advisory on naming of custom features to avoid possible conflicts

If user sets a feature for their local personal program (or whatever reason), it might be allocated by Qubes OS project in the future and break something. Just like how it is mentioned in  #7730

### The solution you'd like

1. Allocate some prefixes (e.g. `user-`, `custom-` or `dev-`) for user custom features workflow. Never use them it in future developments. Document this.
Optional:
2. Have a database of all known features. If user tries to set an unknown feature without the above prefixes, warn them that it might conflict with future features and they would better prepend the custom prefixes.

Advanced users and 3rd party developers who might need custom features.

alimirjamali:
Allocating `x-*` feature prefix to end users.

resolves: https://github.com/QubesOS/qubes-issues/issues/9557

### DIFF
--- a/doc/manpages/qvm-features.rst
+++ b/doc/manpages/qvm-features.rst
@@ -341,6 +341,12 @@ Boot the HVM qube via UEFI boot, instead of legacy one. Support for this boot
 mode is experimental and may not work in all the cases. It is ignored for
 non-HVM qubes.
 
+End user specific features
+--------------------------
+
+Features with the `x-` prefix are specifically meant for end users and will
+never be used for internal Qubes OS features.
+
 Authors
 -------
 


doc/manpages/qvm-features.rst
```diff
@@ -341,6 +341,12 @@ Boot the HVM qube via UEFI boot, instead of legacy one. Support for this boot
 mode is experimental and may not work in all the cases. It is ignored for
 non-HVM qubes.
 
+Custom and user specific features
+---------------------------------
+
+Features with `x-` prefix is dedicated to end users and is guarantied to never
+be used for Qubes OS future internal features.
```
ben-grande: Features with the `x-` prefix are specifically meant for end users and will never
be used for internal Qubes OS features.

---

Grammar and some rewording to make it easier to understand.

Maybe it should not be `user specific features` but `3rd party` as mentioned in https://github.com/QubesOS/qubes-issues/issues/9557#issuecomment-2479008750

ben-grande (changes requested):
Grammar and wording.

marmarek (approved):

PR #331, opened 2025-01-08
tests: Account for argparse bug fix in Python 3.13

DemiMarie:
Test test_004_enable_opt_mixed is marked as xfail because of an argparse bug, but the bug is fixed in Python 3.13 so the test unexpectedly succeeds.  Remove the @unittest.expectedFailure, skip it on Python<3.13, and add a new tests for older Python that checks for the previous buggy behavior.

### DIFF
--- a/qubesadmin/tests/tools/qvm_service.py
+++ b/qubesadmin/tests/tools/qvm_service.py
@@ -20,6 +20,7 @@
 
 # pylint: disable=missing-docstring
 
+import sys
 import unittest
 
 import qubesadmin.tests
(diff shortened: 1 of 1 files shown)


marmarek:
CI says "syntax error"

DemiMarie:
I did test these changes locally, but I also used `black` to try to format the new code properly, and that required running `git add -p` or similar.  I suspect that while the code I had checked out was fine, the code I committed and pushed was broken.  Should be fixed now.

marmarek (approved):

PR #332, opened 2025-01-20
Add Arch packaging

marmarek:
QubesOS/qubes-issues#9231

### DIFF
--- a/.qubesbuilder
+++ b/.qubesbuilder
@@ -9,3 +9,6 @@ vm:
   deb:
     build:
     - debian
+  archlinux:
+    build:
+    - archlinux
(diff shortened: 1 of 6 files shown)


marmarek:
PipelineRetry

ben-grande:
Marek, could there be a tag/label that would server as an openQArun parameter to run the test in a select template that is not installed by default? I don't think you want to have Archlinux being tested in all OpenQA runs, but for every run that impacts Archlinux, I think you'd very much want that.

---

Edit: Also, I think you can rebase to main, some if not all the commits here were merged.

marmarek:
> Marek, could there be a tag/label that would server as an openQArun parameter to run the test in a select template that is not installed by default?

It exists already, `UPDATE_TEMPLATES=archlinux TEST_TEMPLATES=archlinux`. List of all variables is in https://github.com/QubesOS/openqa-tests-qubesos/blob/main/README, but not all are allowed in openQArun command.

DemiMarie:
Time to go ahead and merge?

marmarek:
No, https://github.com/QubesOS/qubes-issues/issues/9231#issuecomment-2614460397 still applies

PR #333, opened 2025-01-28
Add top-like monitoring tool

ben-grande:
Add top-like monitoring tool

Inspiration and code from qui-domains and visuals from xentop.

- Colors entries according to state
- Footer with clock to show last time the screen was refreshed
- Allows scrolling domain list when it is longer than the screen height,
  with a less-type hint indicating how many lines to scroll
- To scroll, readline style (Emacs) navigation is supported, with some
  Vi keys also supported
- Inexpensive calls using "admin.vm.Stats"

Requires:

- https://github.com/QubesOS/qubes-core-admin/pull/827
- https://github.com/QubesOS/qubes-core-admin/pull/894

This PR is based on top of:

- https://github.com/QubesOS/qubes-core-admin-client/pull/480

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -61,10 +61,10 @@ method-rgx=([a-z_][a-z0-9_]{2,60}|assert([A-Z][a-z0-9]*)*)$
 attr-rgx=([a-z_][a-z0-9_]{2,40}|maxDiff)$
 
 # Regular expression which should only match correct argument names
-argument-rgx=[a-z_][a-z0-9_]{2,30}$
+argument-rgx=[a-z_][a-z0-9_]{2,40}$
 
 # Regular expression which should only match correct variable names
-variable-rgx=[a-z_][a-z0-9_]{2,30}$
+variable-rgx=[a-z_][a-z0-9_]{2,40}$
 
 # Regular expression which should only match correct list comprehension /
 # generator expression variable names
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -401,6 +401,8 @@ man_pages = [
         u'Manage tags on a qube', _man_pages_author, 1),
     ('manpages/qvm-template', 'qvm-template',
         u'Manage templates', _man_pages_author, 1),
+    ('manpages/qvm-top', 'qvm-top',
+        u'Top-like monitoring tool', _man_pages_author, 1),
     ('manpages/qvm-unpause', 'qvm-unpause',
         u'Pause a qube', _man_pages_author, 1),
     ('manpages/qvm-volume', 'qvm-volume',
--- a/doc/manpages/qvm-top.rst
+++ b/doc/manpages/qvm-top.rst
@@ -1,0 +1,298 @@
+.. program:: qvm-top
+
+:program:`qvm-top` -- top-like monitoring tool
+==============================================
+
+Synopsis
+--------
+
+:command:`qvm-top` [--verbose] [--quiet] [--help] [--version] [--all] [--exclude *EXCLUDE*] [--show-halted] [--filter *FILTER*,...] [--columns *COLUMN*,... | --format *FORMAT*] [--help-columns] [--help-formats] [--sort-column *COLUMN*] [--reverse] [--no-color] [*VMNAME* ...]
+
+Top-like info for Qubes. Defaults is to show all non-halted qubes.
+
+Positional arguments
+--------------------
+
+.. option:: VMNAME ...
+
+   Zero or more domain names. If none spcified, the default is to show all.
+
+General Options
+---------------
+
+.. option:: --verbose, -v
+
+   Increase verbosity.
+
+.. option:: --quiet, -q
+
+   Decrease verbosity.
+
+.. option:: --help, -h
+
+   Show this help message and exit.
+
+.. option:: --version
+
+   Show program's version number and exit.
+
+.. option:: --all
+
+   Perform the action on all qubes
+
+.. option:: --exclude EXCLUDE
+
+   Exclude the qube from ``--all``.
+
+.. option:: --show-halted, -S
+
+   Don't hide halted qubes.
+
+.. option:: --filter, -f FILTER,...
+
+   Filter domains name matching each fixed string separated by comma.
+
+.. option:: --no-color
+
+   Do not colorize the screen. If this option is not provided, but environment
+   variable `NO_COLOR` is set, the screen will not be colorized.
+
+Formatting Options
+------------------
+
+.. option:: --columns, -C COLUMN,...
+
+   Show only specified columns.
+
+.. option:: --format, -F FORMAT
+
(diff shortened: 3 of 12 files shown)


marmarek:
> is a blocking operation as well as `proc.stdout.read(2)`

Do you mean read() hangs for you? Maybe the service waits for something before starting to output any data (doesn't look like it based on the other PR...).

ben-grande:
Not the service by default, I prepended sleep to awk to test if a DomU
could hang operation on Dom0 and if this is a valid concern.

marmarek:
Based on discussion in the other PR, this should be converted to use stat events (and events extended to include necessary info). Converting to draft until then.

ben-grande:
LLM helped on `draw_table()`. It's almost stable. It works nicely on R4.2, but not on R4.3. I don't know yet why builds are failing on manpages.

ben-grande:
One thing I don't plan to fix is the memory reporting being wrong, as only qmemman has the knowledge about memory of each qube, and getting information from it is not currently nice with it's current API (synchonous), and I don't want to delay reporting statistics because qmemman is busy for seconds at a time.

- https://github.com/qubesos/qubes-issues/issues/10857

ben-grande:
Maybe it's not necessary, reading past conversations, something that I didn't understand one year ago (qmemman, admin.vm.Stats), I think I will settle for meminfo: https://github.com/QubesOS/qubes-core-agent-linux/pull/550#issuecomment-2619166703.

ben-grande:
qvm-top

`xentop` has `cpu_time` for each `vcpu`, but that may be too much. Xentop has actually a lot of things that can be show about virtual blocks also, but I don't like that much info. Customizing delay is not important to us, the screen refreshed in a loop and the delay is pretty insignificant and is not causing issues to navigate up and down. One think I like is that xentop has sort order, but I can't really see which column is currently being sorted, it's supposed to be bold. Anyway, I like how qvm-top is right now and don't plan to improve further (unless I think about this again, tomorrow...).

ben-grande:
And one more bug, for dom0, it should use `cpu_usage_raw` instead of `cpu_usage`. It seems that the value gets much closer to what xentop reports about Domain-0.

ben-grande:
And yes, `vm-stats` from my core-admin PR is reporting stubdomains, but as the client VMCollection doesn't know about stubdomains, it's never considered. Maybe I could workaround this limitation.

ben-grande:
new-qvm-top

- stubdom reporting, can't show memory usage for it, but can show CPU stats
- NA is better than lying how much memory the qube is using
- I don't know why Xen adds one mebibyte to `maxmem_kb` (`domain_getinfo`) but not on `static-max` xenstore key.

marmarek:
> I don't know why Xen adds one mebibyte to `maxmem_kb` (`domain_getinfo`) but not on `static-max` xenstore key.

I'm not sure either, but I guess `domain_getinfo` includes some internal overhead in Xen (EPT? maybe something else too?), but `static-max` is just what OS can really see.

Anyway, I don't like exposing stubdomains in a high level tool, it should be counted as part of the target qube; if anything, maybe some separate column for some internal overhead, but not separate row.
If somebody really want to look under the hood, they have `xentop` and other low level tools.

ben-grande:
> Anyway, I don't like exposing stubdomains in a high level tool, it should be counted as part of the target qube; if anything, maybe some separate column for some internal overhead, but not separate row.

Ok, will change the separate row to either:

- sum to domain stats: less clutter than separate column, not a problem for memory as it doesn't change, but CPU stats does
- separate column: I don't like having many extra columns..., although it would be truthful, it greatly lowers readability.

> If somebody really want to look under the hood, they have `xentop` and other low level tools.

Xentop doesn't work from domU (maybe there is a Xen policy for this) and won't show correct information if not running under Xen. The idea is for interface qubes to be able to analyze stats without requiring to run a terminal in dom0. It also doesn't report the correct memory information when qmemman is active. So this is the reason why I am pushing for greater observability. See sub-issues of: https://github.com/qubesos/qubes-issues/issues/2134.

ben-grande:
newnew-qvm-top

I think it is better to show aggregated values rather than use extra columns that will be zero most of the time except for HVMs.

ben-grande:
I have added a ton of things recently. You can do everything from the keyboard or mouse. You can filter the view, sort by column (or invert), select one or more domains. For the latter, it does nothing currently besides highlighting, the goal is to have an htop like kill column, where you select the kill signal, in our case, it would be "shutdown, kill, (re)start". If there is anything you'd like to see, please let me know. There will be a video tomorrow, if QVC doesn't burn my computer.

In the far future, this might become the qube manager for headless connections, but for now, I will restrict just for statistics due to time.

ben-grande:
- Tested with RemoteVM
- Reduced default view to just a few columns
- Improved no color view
- Actions available, from niceties such as terminal or console to controlling qube state
- Reordered columns as `total`, `used`, `internal` 
- New swap column
- Qube names are colored according to their label if the terminal supports, else it will fallback to something close enough

I think I am finished on the client, but will have to retest it after writing and fixing tests of https://github.com/QubesOS/qubes-core-admin/pull/827.

---

For the description of each column, see the manual page or the `help=` of the `Column`.

**Minimal view** (55 columns):

top-min

**Default view** (79 columns):

top-default

**Maximum view** (203 columns):

top-max

The `min` view is what `qui-domains` should show.

The `default` view is what you get without specifying any arguments.

If you noticed, I didn't include `maxmem` property in the `default` or `min`, as I think is `MAU` -> "memory_assigned_usable", which is what is available in the domain, is more relevant to show available to the domain and `MU/MAU` to show memory usage (excluding swap, separate column `MUS`).

The `max` view is all you can get. Great for debugging the application and see if values add up when they should.

---

**Actions**:

top-action

top-preload

top-action-common

Notice that when selecting both domains, you can only get the common actions.

ben-grande:
PR rebased on top of:

- https://github.com/QubesOS/qubes-core-admin-client/pull/480

ben-grande:
This code is working with latest commit of https://github.com/QubesOS/qubes-core-admin/pull/827.

`-F max`

latest-qvm-top

`-F max --thin-columns`

latest-short-qvm-top

`-F max --thin-columns --memory-unit GiB.0`

gib-qvm-top

ben-grande:
> `-F max --thin-columns --memory-unit GiB.0`
> gib-qvm-top

The headers were wrong on the image above, see for example CPUsec, it is trying to convert it to the memory unit, that is wrong. It is fixed now:

`-F max --thin-columns --memory-unit GiB.0`

neww-top

`-F max --thin-columns --memory-unit GiB.1`

newww-top

ben-grande:
Add qube uptime. AdminVM doesn't set it, so it should be added. Display time in the format `HH:MM:SS`, where only `HH` can grow beyond two digits. Or find a better format to display days when it goes over twenty four hours.

PR #334, opened 2025-02-16
Add free-form text to qube for notes, comments, ...

### LINKED ISSUE #899: Allow users to add free-form text to qubes (for descriptions, notes, comments, remarks, reminders, etc.)
**Reported by Oleg Artemiev  grey olli on 12 Sep 2014 08:14 UTC**
It could be usefull to have a text description for a VM in its
properties to document changes/difference.
https://groups.google.com/d/topic/qubes-users/uYCcOqIeF5k/discussion

Migrated-From: https://wiki.qubes-os.org/ticket/899

alimirjamali:
qubesadmin part of adding free-form text to each qube for comments, notes, descriptions, remarks, reminders, etc.

fixes: https://github.com/QubesOS/qubes-issues/issues/899

### DIFF
--- a/qubesadmin/vm/__init__.py
+++ b/qubesadmin/vm/__init__.py
@@ -424,6 +424,20 @@ class QubesVM(qubesadmin.base.PropertyHolder):
             self._klass = super().klass
         return self._klass
 
+    def get_notes(self) -> str:
+        ''' Get qube notes '''
+        response = self.qubesd_call(self._method_dest, 'admin.vm.notes.Get')
+        return response.decode()
+
+    def set_notes(self, notes: str):
+        ''' Set qube notes '''
+        self.qubesd_call(
+            self._method_dest,
+            'admin.vm.notes.Set',
+            payload=str(notes).encode(encoding='utf-8')
+        )
+
+
 class DispVMWrapper(QubesVM):
     '''Wrapper class for new DispVM, supporting only service call
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -84,7 +84,7 @@ except:
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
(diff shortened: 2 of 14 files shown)


qubesadmin/vm/__init__.py
```diff
@@ -394,6 +394,23 @@ def klass(self):
             self._klass = super().klass
         return self._klass
 
+    def get_notes(self) -> str:
+        ''' Get qube notes '''
+        try:
+            response = self.qubesd_call(self._method_dest, 'admin.vm.notes.Get')
+        except Exception as exc:
```
marmarek: This is definitely too broad. If anything, maybe just `qubesadmin.exc.QubesDaemonNoResponseError`, but even that I'm not sure. qubesd should return already correct exception on error.

qubesadmin/vm/__init__.py
```diff
@@ -394,6 +394,23 @@ def klass(self):
...
+            raise qubesadmin.exc.QubesNotesError(str(exc))
+        return response.decode()
+
+    def set_notes(self, notes: str):
+        ''' Set qube notes '''
+        try:
+            self.qubesd_call(self._method_dest, 'admin.vm.notes.Set', \
+                payload = str(notes).encode('ascii'))
+        except Exception as exc:
```
marmarek: and same here

marmarek (changes requested):

qubesadmin/vm/__init__.py
alimirjamali: This is done

qubesadmin/vm/__init__.py
alimirjamali: This is also done

alimirjamali:
@marmarek please let me know if you agree with `qvm-notes` syntax and options. So I can proceed with its unittests. in the meantime, I will write the unittests for the rest.

marmarek:
Generally qvm-notes interface looks fine (based on its man page). I think append action needs a clarification regarding newlines if any are inserted automatically.

As for the implementation, IMO too many of the actions ask for confirmation. Maybe remove should, but others seems unnecessary.

alimirjamali:
This is ready for review. Only 3 lines are not covered by tests but two lines are abstract lines (maybe I should add `#pragma: no cover`) and one other is for backup restore notes failure which is highly unlikely to happen (this is covered in lots of other places).

> Generally qvm-notes interface looks fine (based on its man page). I think append action needs a clarification regarding newlines if any are inserted automatically.

This is done. I also added some more notes for better clarification

> 
> As for the implementation, IMO too many of the actions ask for confirmation. Maybe remove should, but others seems unnecessary.

Theoretically It is possible to check with `get_notes` and only ask for confirmation on `set, delete & import` actions that the action is going to replace existing note. On the other hand, it will require `admin.vm.notes.Get` permission. So I implemented your recommendation.

alimirjamali:
@marmarek I just rebased this and excluded few irrelevant lines for codecov. Otherwise nothing is changed.

marmarek (approved):

PR #335, opened 2025-02-17
Revert to red label while restoring on missing labels

### LINKED ISSUE #9781: Allow users to restore backup of qubes with custom labels on new systems
### Qubes OS release

Qubes OS 4.2

### Brief summary

If we allow/provide users with tools and/or API to create custom labels and they backup qubes with such labels, the restore operation would fail for such qubes on new systems without those custom labels.

### Steps to reproduce

1. Create custom label with `admin.label.Create` API call.
2. Set a qube label to this new custom label
3. Backup that qube
4. Try to restore it

### Expected behavior

Restore operation could revert to an existing label and allow user to access their backup

### Actual behavior

Restore operation for qubes with custom label fails with vague error

### Additional information

_No response_

alimirjamali:
If we allow users to create custom labels and they backup such qubes, restore operation will fail on new systems without those custom labels. Fix this by reverting to red label and allow users to restore their backup.

fixes: https://github.com/QubesOS/qubes-issues/issues/9781

### DIFF
--- a/qubesadmin/tests/backup/backupcompatibility.py
+++ b/qubesadmin/tests/backup/backupcompatibility.py
@@ -695,7 +695,7 @@ parsed_qubes_xml_v4 = {
         },
         'test-net': {
             'klass': 'AppVM',
-            'label': 'red',
+            'label': 'scarlet',
             'properties': {
                 'maxmem': '300',
                 'memory': '300',
@@ -1357,7 +1357,31 @@ class TC_10_BackupCompatibility(qubesadmin.tests.backup.BackupTestCase):
             if name == 'dom0':
                 continue
 
-            if self.storage_pool:
+            if self.storage_pool and vm['label'] == 'scarlet':
+                self.app.expected_calls[
+                    ('dom0', 'admin.vm.CreateInPool.' + vm['klass'],
+                     templates_map.get(vm['template'], vm['template']),
+                    'name={} label={} pool={}'.format(
+                        name, 'scarlet', self.storage_pool).encode())] = \
+                    b'2\0QubesLabelNotFoundError\0\0No such label\0'
+                self.app.expected_calls[
+                    ('dom0', 'admin.vm.CreateInPool.' + vm['klass'],
+                     templates_map.get(vm['template'], vm['template']),
+                    'name={} label={} pool={}'.format(
+                        name, 'red', self.storage_pool).encode())] = \
+                    b'0\0'
+            elif vm['label'] == 'scarlet':
+                self.app.expected_calls[
+                    ('dom0', 'admin.vm.Create.' + vm['klass'],
+                     templates_map.get(vm['template'], vm['template']),
+                    'name={} label={}'.format(name, 'scarlet').encode())] =\
+                    b'2\0QubesLabelNotFoundError\0\0No such label\0'
+                self.app.expected_calls[
+                    ('dom0', 'admin.vm.Create.' + vm['klass'],
+                     templates_map.get(vm['template'], vm['template']),
+                    'name={} label={}'.format(name, 'red').encode())] =\
+                    b'0\0'
+            elif self.storage_pool:
                 self.app.expected_calls[
                     ('dom0', 'admin.vm.CreateInPool.' + vm['klass'],
                      templates_map.get(vm['template'], vm['template']),
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -2029,6 +2029,14 @@ class BackupRestore(object):
                         label=vm.label,
                         pool=self.options.override_pool,
                         **kwargs)
+                except qubesadmin.exc.QubesLabelNotFoundError:
+                    # do not fail if label is not present. revert to red label
+                    new_vm = self.app.add_new_vm(
+                        vm.klass,
+                        name=vm_name,
+                        label='red',
+                        pool=self.options.override_pool,
+                        **kwargs)
                 except Exception as err:  # pylint: disable=broad-except
                     self.log.error('Error restoring VM %s, skipping: %s',
                         vm.name, err)
--- a/qubesadmin/tests/backup/v4-qubes.xml
+++ b/qubesadmin/tests/backup/v4-qubes.xml
@@ -8,6 +8,7 @@
     <label color="0x3465a4" id="label-6">blue</label>
     <label color="0x75507b" id="label-7">purple</label>
     <label color="0x000000" id="label-8">black</label>
+    <label color="0xff2400" id="label-9">scarlet</label>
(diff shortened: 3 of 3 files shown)


alimirjamali:
I am not certain why codecov has marked those two lines as uncovered. The unittests had to be specifically changed since those two lines were indeed working

marmarek:
> I am not certain why codecov has marked those two lines as uncovered. The unittests had to be specifically changed since those two lines were indeed working

Looks like some backup tests are skipped because `scrypt` tool is not installed. It isn't available in Fedora repo, only in Qubes repo and this one is not included in the docker used for the test...

alimirjamali:
> > I am not certain why codecov has marked those two lines as uncovered. The unittests had to be specifically changed since those two lines were indeed working
> 
> Looks like some backup tests are skipped because `scrypt` tool is not installed. It isn't available in Fedora repo, only in Qubes repo and this one is not included in the docker used for the test...

So what is the appropriate solution here? Is there a way to include `qubes-vm-r4.3-current-testing` in docker to install scrypt binary and re-enable the skipped tests? Or it is necessary to switch to `python3-scrypt` package and do the decryption directly in Python instead of depending on the external binary?

marmarek:
Rebase on top of https://github.com/QubesOS/qubes-core-admin-client/pull/337 (soon main)

alimirjamali:
> Rebase on top of #337 (soon main)

OK. Unit tests are green again and looking at Github report, it appears that backupcompatibility tests are not skipped. But the Codecov report is still not covered.

Another issue is that the backupcompatibility tests are kinda slow. I believe after this is done, it might be better to disable `ENABLE_SLOW_TESTS` in `.gitlab-ci.yml` via another PR to make the life easier for other PRs which are not backup related.

marmarek:
> But the Codecov report is still not covered.

I think you changed label in test in a wrong place. Change it in the qubes.xml, not the parsed structure.

marmarek:
openQA says it breaks restoring backup using a DispVM:
```
Feb 23 15:01:58.885143 dom0 qrexec-policy-daemon[1693]: qrexec: admin.label.List: disp-backup-restore -> dom0: denied: no matching rule found
```
Theoretically the policy could be extended to allow listing labels, but maybe it isn't needed? Maybe instead of check before, you can try setting the label and fallback to "red" in case of `QubesLabelNotFoundError` exception?

alimirjamali:
OpenQA's tests for restore in DispVM fail since it does not have permission for `admin.label.List`. Is it in this repo to fix?

marmarek:
The policy for DispVM restore is in core-admin. But as said above, maybe this extra permission (and also extra call during usual restore) isn't actually needed?

alimirjamali:
> The policy for DispVM restore is in core-admin. But as said above, maybe this extra permission (and also extra call during usual restore) isn't actually needed?

Unfortunatley it is needed. Unless we want to hard-code the choices of default labels, I do not know any other way to test if the label of the VM to restore is one of them.

marmarek:
Ah, because label is mandatory to create a VM, setting it cannot be deferred to later. I was thinking about something like:
```python
try:
    vm.label = label
except QubesLabelNotFoundError:
    vm.label = "red"
```
but that indeed won't work. And worse - it looks like `admin.vm.Create` raises generic `QubesValueError` on unknown label (which BTW might be a bug - `QubesLabelNotFoundError` would be more logical), so you can't even do that try/except with the create call.

alimirjamali:
There is a less invasive and more privacy friendly way by using `admin.label.Get` call to check of the label exists

alimirjamali:
> (which BTW might be a bug - `QubesLabelNotFoundError` would be more logical)

Should I implement it for core?

marmarek:
https://github.com/QubesOS/qubes-core-admin/pull/663

alimirjamali:
> [QubesOS/qubes-core-admin#663](https://github.com/QubesOS/qubes-core-admin/pull/663)

Very nice. I will amend this PR and revert to `try: except:` approach as soon as the above is merged.

marmarek:
Now that https://github.com/QubesOS/qubes-core-admin/pull/663 is merged you can update this PR. Or do you want me to push an updated package to current-testing first (doesn't matter for openqa)?

alimirjamali:
> Now that [QubesOS/qubes-core-admin#663](https://github.com/QubesOS/qubes-core-admin/pull/663) is merged you can update this PR. Or do you want me to push an updated package to current-testing first (doesn't matter for openqa)?

Very nice. Let's see if this works. What about the unittests? Will they fetch the main branch or use current-testing?

alimirjamali:
Ok. It appears that the unittests fail since they depend on `current-testing`.

marmarek:
No, unittests do not use real core-admin. There is a mockup that is [configured to mimic its behavior](https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/tests/backup/backupcompatibility.py#L1348). You need to update it to match the change, specifically to return an exception for the call with scarlet label (the one that is now shown in test failure message as unexpected call).

qubesadmin/tests/backup/backupcompatibility.py
```diff
@@ -1357,6 +1357,20 @@ def setup_expected_calls(self, parsed_qubes_xml, templates_map=None):
...
+                     templates_map.get(vm['template'], vm['template']),
+                    'name={} label={} pool={}'.format(
+                        name, 'red', self.storage_pool).encode())] = \
+                    b'0\0'
+            elif vm['label'] == 'scarlet':
+                self.app.expected_calls[
+                    ('dom0', 'admin.vm.Create.' + vm['klass'],
+                     templates_map.get(vm['template'], vm['template']),
+                    'name={} label={}'.format(name, 'red').encode())] =\
+                    b'0\0'
```
marmarek, line 1383: In addition to the expected call using `red` label, you need to add also expected call with `scarlet` label that fails with QubesLabelNotFoundError. Exception is coded in the response, so instead of `b'0\0'`, it should be `b'2\0QubesLabelNotFoundError\0\0No such label\0'`
Right now I think the test will simply pass because there will be "expected call" with label scarlet that succeeds, which means your added code won't be covered...

alimirjamali: Sorry for mistake. I should stop jumping between different PRs and focus.

alimirjamali: @marmarek This is now ready for review. CI unittests are green. Codecov changes are green.

Sorry again for not running the tests locally before pushing.

After this is done, I guess it would be good to disable slow tests in `,gitlab-ci.yml` once more (`ENABLE_SLOW_TESTS`).

marmarek (approved):

PR #342, opened 2025-03-09
Add boot mode support

ArrayBolt3:
Boot modes provide a way for VMs to advertise that they can have their behavior changed by booting them with a predefined combination of additional kernel parameters set. This allows qubes to offer features like a secure system maintenance mode in a way that works seemlessly with Qubes OS.

This commit documents the features and properties used in boot mode support, and fixes some automated tests.


marmarek (approved):
This looks okay, but I'll wait with merging until tests for boot mode in qubes manager are complete, in case mockups needs some more changes

marmarek:
Actually, pylint proposed one improvement, which is a good idea.

PR #344, opened 2025-03-23
Client API for denied devices

piotrbartman:


marmarek:
pylint complains...

marmarta:
Could we also be able to deny a particular device (does this make sense?)? Is the only way to deny by listing interfaces? Can I deny all block devices, or all USB devices, or all devices?

piotrbartman:
> Could we also be able to deny a particular device (does this make sense?)? 

no since denied list only applies to manually connected devices. The goal was to prevent connecting something that looks like a safe device (e.g., a USB stick) but actually has more interfaces (e.g., a keyboard). So, if we don’t want to manually connect a specific (known) device, we can simply avoid connecting it.

> Is the only way to deny by listing interfaces? 

yes

> Can I deny all block devices, or all USB devices, or all devices?

yes
`*******` - block all devices
`u******` - block USB device
`m******` - block mic
`b******` - block block devices
moreover interface encoding (from file `/usr/share/hwdata/usb.ids`) for USB contains CCSSII where  CC is class ID, SS is subclass ID and II is interface ID, so
`u03****` - block HID devices (keyboards, mouses and others)
`u03**01` - block all USB keyboards. 
etc.

marmarta:
alright, I understand, this looks good then :)

qubesadmin/tests/devices.py
```diff
@@ -323,3 +323,47 @@ def test_073_list(self):
...
+        self.app.domains['test-vm'].devices.allow()
+
+    def test_085_allow_device_multiple(self):
+        self.app.expected_calls[
+            ('test-vm', 'admin.vm.device.denied.Remove', None,
+             b"uabcdefm******")] = b"0\x00"
+        self.app.domains['test-vm'].devices.allow(
+            qubesadmin.device_protocol.DeviceInterface("uabcdef"),
+            qubesadmin.device_protocol.DeviceInterface("m******"),
+        )
```
marmarek: All those tests want `self.assertAllCalled()` at the end.

marmarek (changes requested):

marmarek (approved):

PR #346, opened 2025-04-02
Enable more common compression filters for restoring

alimirjamali:
Since some users want pigz, zstd and some other compressions which might not be installed by default, enable them if they are installed.

Related: https://github.com/QubesOS/qubes-issues/issues/8291
Related: https://github.com/QubesOS/qubes-issues/issues/8211

### DIFF
--- a/qubesadmin/tests/backup/backupcompatibility.py
+++ b/qubesadmin/tests/backup/backupcompatibility.py
@@ -2048,6 +2048,83 @@ class TC_10_BackupCompatibility(qubesadmin.tests.backup.BackupTestCase):
 
     @unittest.skipUnless(shutil.which('scrypt'),
         "scrypt not installed")
+    def test_230_r4_optional_compression(self):
+        self.create_v4_backup("zstdmt")
+
+        self.app.expected_calls[('dom0', 'admin.vm.List', None, None)] = (
+            b'0\0dom0 class=AdminVM state=Running\n'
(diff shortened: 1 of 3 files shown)


qubesadmin/tests/backup/backupcompatibility.py
```diff
@@ -1563,13 +1563,19 @@ def test_210_r2(self):
...
         patches = [
             mock.patch('qubesadmin.storage.Volume',
-                functools.partial(MockVolume, qubesd_calls_queue, 0)),
+                staticmethod(
+                    functools.partial(MockVolume, qubesd_calls_queue, 0))
+                ),
             mock.patch(
                 'qubesadmin.backup.restore.BackupRestore._handle_appmenus_list',
-                functools.partial(self.mock_appmenus, qubesd_calls_queue)),
+                staticmethod(
```
marmarek, line 1571: Adding staticmethod don't look related to anything described in the commit message - should be a separate commit with info what/why.

marmarek (changes requested):

qubesadmin/tests/backup/backupcompatibility.py
alimirjamali: I split the PR to two individual PRs. The 2nd commit is to reduce the clutter in unittest output. As Python asks the `functools.partial` to be wrapped within `staticmethod`.

marmarek (approved):

PR #351, opened 2025-04-17
Assure backup restore/verify tmpdir is properly deleted

### LINKED ISSUE #9387: Backup verification leaves files in dom0 home directory

### Qubes OS release

4.2

### Brief summary

After verifying a backup, Qubes leaves leftover files in the user's dom0 home directory.

### Steps to reproduce

Use the GUI backup restore tool to verify the integrity of a backup file that resides in an app qube.

### Expected behavior

1. Qubes OS shouldn't put files in the user's home directory without user consent and informing the user.
2. Qubes OS should clean up after itself, not leave temp files lying around.

### Actual behavior

There is an unexpected leftover `QubesIncoming/` directory in the user's dom0 home directory containing some files related to the operation. It is not clear why this exists or what it's for unless you're at least somewhat familiar with how the backup verification process works.

### Commentary

The home directory is the *one* place in the system where users are told to store their own personal files. Users should be able to expect to exercise full control over that one directory. The entire directory should be considered "user data." Finding files there that the user did not put there can be alarming and could be an indicator of compromise. At no point during the backup verification process is the user warned or informed that any files will be created in the dom0 home directory. The system should never touch that directory as part of its own internal operations. (It has the whole rest of the file system to work with.) Even if it does have to touch the user's data for some reason, it should clean up its own temp files after it's done, not force the user to clean up after it.

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/9387

### DIFF
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -956,7 +956,7 @@ class BackupRestore(object):
         # due to possible location in ~/QubesIncoming, the prefix should not be
         # a valid VM name
         os.makedirs(tmpdir, exist_ok=True)
-        self.tmpdir = tempfile.mkdtemp(prefix="backup#restore-", dir=tmpdir)
+        self.tmpdir = tempfile.mkdtemp(prefix=".backup#restore-", dir=tmpdir)
 
         #: list of processes (Popen objects) to kill on cancel
         self.processes_to_kill_on_cancel = []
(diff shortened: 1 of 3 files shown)


qubesadmin/backup/restore.py
```diff
@@ -977,6 +977,10 @@ def __init__(self, app, backup_location, backup_vm, passphrase, *,
         #: VMs included in the backup
         self.backup_app = self._process_qubes_xml()
 
+    def __del__(self):
+        if os.path.exists(self.tmpdir):
```
marmarek: The "normal" place where the tmpdir is removed is a `finally` block in the `restore_do` function. See that there is an extra condition based on log level - the idea is that with enough `-v` (debug log level) the temp dir should remain.

The fix should either adjust that `finally` section to work in verify mode too (I'm not sure why it doesn't), or explain in the commit message why it isn't possible and `__del__` must be used. And in the latter case, preserve the functionality to optionally leave the directory in place.

marmarek (changes requested):

qubesadmin/backup/restore.py
alimirjamali: > or explain in the commit message why it isn't possible and `__del__` must be used. And in the latter case, preserve the functionality to optionally leave the directory in place.

I updated the commit message and applied the requested functionality

marmarek (approved):

PR #352, opened 2025-04-23
Pass recognized boolean values to qubes-guid

alimirjamali:
Since libconfig's `config_setting_get_bool` recognises limited values, convert feature value to something it properly recognises

related: https://github.com/QubesOS/qubes-issues/issues/7730
reference: https://hyperrealm.github.io/libconfig/libconfig_manual.html#Settings

### DIFF
--- a/qubesadmin/tests/tools/qvm_start_daemon.py
+++ b/qubesadmin/tests/tools/qvm_start_daemon.py
@@ -784,9 +784,23 @@ HDMI1 connected 2560x1920+0+0 (normal left inverted right x axis y axis) 206mm x
             ('gui-vm', 'admin.vm.feature.Get',
              'gui-default-override-redirect', None)] = \
                  b'0\x00ask'
+        self.app.expected_calls[
+            ('gui-vm', 'admin.vm.feature.Get',
+             'gui-default-allow-fullscreen', None)] = \
+                 b'0\x00True'
+        self.app.expected_calls[
+            ('gui-vm', 'admin.vm.feature.Get',
+             'gui-default-allow-utf8-titles', None)] = \
+                 b'0\x000'
+        self.app.expected_calls[
+            ('gui-vm', 'admin.vm.feature.Get',
+             'gui-default-override-redirect-protection', None)] = \
+                 b'0\x00Maybe'
 
         _args, config = self.run_common_args()
         self.assertEqual(config, '''\
 global: {
+  allow_fullscreen = true;
+  allow_utf8_titles = false;
 }
 ''')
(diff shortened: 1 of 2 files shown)


marmarek:
What is the purpose of this change? The feature value was converted to bool consistently with other features already. With this change, features for GUI settings behave differently (for example string `"false"` is now considered False, while it works as True everywhere else).

alimirjamali:
> What is the purpose of this change? The feature value was converted to bool consistently with other features already. With this change, features for GUI settings behave differently (for example string `"false"` is now considered False, while it works as True everywhere else).

The [manpage](https://github.com/QubesOS/qubes-core-admin-client/blob/main/doc/manpages/qvm-features.rst) for `qvm-features` mentions:

> For example, to enable allow_utf8_titles for a qube, set gui-allow-utf8-titles to True.

But user has no clue that `False` or `false` also means true. Whereas in the previous behaviour (where the `/etc/qubes/guid.conf` were used, `false` meant false. And it's documentation aslo says `false` means false:

https://github.com/QubesOS/qubes-gui-daemon/blob/main/gui-daemon/guid.conf#L48-L52

alimirjamali:
p.s.: We should also considered that the change which introduced `bool()` was not backported to r4.2 (rationale link below). So the majority of users are still familiar with old behaviour (where `false` is false).

https://github.com/QubesOS/qubes-issues/issues/7730#issuecomment-2481260897

marmarek:
> The [manpage](https://github.com/QubesOS/qubes-core-admin-client/blob/main/doc/manpages/qvm-features.rst) for `qvm-features` mentions:

This very same man page says also:
> Some extensions interpret the values as boolean. In this case, the empty string means [:py:obj:`False`](https://github.com/QubesOS/qubes-core-admin-client/blob/main/doc/manpages/qvm-features.rst#id5) and non-empty string (commonly '1') means [:py:obj:`True`](https://github.com/QubesOS/qubes-core-admin-client/blob/main/doc/manpages/qvm-features.rst#id7). An absence of the feature means "default", which is extension-dependent. In most cases the default value for feature is retrieved from a qube template.
---

> p.s.: We should also considered that the change which introduced `bool()` was not backported to r4.2 (rationale link below). So the majority of users are still familiar with old behaviour (where `false` is false).

Nope, `false` is true in R4.2 too: https://github.com/QubesOS/qubes-core-admin-client/blob/release4.2/qubesadmin/tools/qvm_start_daemon.py#L145-L146

alimirjamali:
> Nope, `false` is true in R4.2 too: https://github.com/QubesOS/qubes-core-admin-client/blob/release4.2/qubesadmin/tools/qvm_start_daemon.py#L145-L146

Ok. I close the PR. It is just a little bit confusing.

PR #353, opened 2025-04-23
SSL auth and user GPG key support for user templates repos

Guiiix:
This commit adds the support of user defined GPG keys and SSL authentication for user templates repositories.
GPG keys and SSL cert/key must be configured in DNF repositories files using the well-known options
(gpgkey, sslclientcert, sslclientkey).
If the keys are stored in /etc/qubes/repo-templates/keys/, they will be added to the payload sent to proxy in base64 and will be written to the Proxy VM before DNF command execution

https://github.com/QubesOS/qubes-issues/issues/9850

### DIFF
--- a/qubesadmin/tests/tools/qvm_template.py
+++ b/qubesadmin/tests/tools/qvm_template.py
@@ -5559,6 +5559,205 @@ CPE_NAME="cpe:/o:ITL:qubes:4.2"
         gen_rpm(False, execute)
         self.assertAllCalled()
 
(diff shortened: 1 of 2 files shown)


marmarek:
Generally looks good, but it could use an unit test. See test_110_qrexec_payload_refresh_success in qubesadmin/tests/tools/qvm_template.py for similar case.

marmarek:
Looks like `$releasever` needs to be supported in the path - at least the default repo config uses it. As said on the chat, this one qvm-template knows.

marmarek (approved):

PR #354, opened 2025-04-24
Document preloaded disposables

ben-grande:

---

One thing that I am not happy about is how many times I had to write _preloaded disposables_, _disposable to preload_, mostly because it lacks synonyms and if I start using synonyms, it might cause conflicts or confusion in the future. See the [glossary](https://www.qubes-os.org/doc/glossary/), especially the _previously known as_ lines.


marmarek:
PipelineRetry

marmarek (approved):

PR #358, opened 2025-05-15
Improve `qvm-backup-restore --paranoid-mode`

### LINKED ISSUE #9962: Fresh R4.2 install gives `qvm-backup-restore: command not found` during paranoid restore
### Qubes OS release

Qubes OS 4.2

### Brief summary

On a fairly fresh install of R4.2 (pre-installed at NovaCustom and updated afterwards), `qvm-backup-restore --paranoid-mode` gives:

```
qvm-backup-restore: line 1: qvm-backup-restore: command not found
exit code: 127
```

in the console window that opens up.

### Steps to reproduce

1. Run `qvm-backup-restore --paranoid-mode` without installing any extra packages in TemplateVMs.

### Expected behavior

Works

### Actual behavior

_No response_

### Additional information

This is due to `qubes-core-admin-client` not being installed.

alimirjamali:
Before openning a Shell window in the DispVM, check for availability of `core-admin-client` in the DisposableVM and show error message if not available

resolves: https://github.com/QubesOS/qubes-issues/issues/9962

### DIFF
--- a/doc/manpages/qvm-backup-restore.rst
+++ b/doc/manpages/qvm-backup-restore.rst
@@ -101,6 +101,8 @@ Options
     - dom0 home directory (desktop environment settings)
     - PCI devices assignments
 
+  This operation requires `qubes-core-admin-client` package in the DisposableVM
+
 .. option:: --auto-close
 
   When running with --paranoid-mode (see above), automatically close restore
--- a/qubesadmin/backup/dispvm.py
+++ b/qubesadmin/backup/dispvm.py
@@ -276,10 +276,27 @@ class RestoreInDisposableVM:
         lock = qubesadmin.utils.LockFile(LOCKFILE, True)
         lock.acquire()
         try:
+            self.app.log.info("Starting restore process in a DisposableVM...")
             self.create_dispvm()
             self.clear_old_tags()
             self.register_backup_source()
             self.dispvm.start()
+            try:
+                self.app.log.debug(
+                    "Checking for existence of qubes-core-admin-client"
+                )
+                self.dispvm.run("command -v qvm-backup-restore")
+            except subprocess.CalledProcessError:
+                raise qubesadmin.exc.QubesException(
+                    'qvm-backup-restore tool '
+                    'missing in {} template, install qubes-core-admin-client '
+                    'package there'.format(
+                        getattr(self.dispvm.template,
+                                'template',
+                                self.dispvm.template).name)
+                )
+            self.app.log.info("When operation completes, close its window "
+                              "manually.")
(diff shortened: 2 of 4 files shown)


ben-grande:
Use `command -v` (POSIX) instead of `which`, which might not be present by default on Alpine Linux.

alimirjamali:
> Use `command -v` (POSIX) instead of `which`, which might not be present by default on Alpine Linux.

Justed tested it on Alpine. `which` is indeed a part of Busybox. But OK. I will switch the code to `command -v` as soon as the current iteration of CI/CD tasks is complete.

ben-grande:
> Justed tested it on Alpine. `which` is indeed a part of Busybox.

Oh, ok, they might have their own implementation because when I saw a [separate package](https://pkgs.alpinelinux.org/package/edge/main/x86_64/which), the `Required by` did not give me much clues, but that package is for the Gnu version of which? While busybox which is probably builtin then.

alimirjamali:
> > Justed tested it on Alpine. `which` is indeed a part of Busybox.
> 
> Oh, ok, they might have their own implementation because when I saw a [separate package](https://pkgs.alpinelinux.org/package/edge/main/x86_64/which), the `Required by` did not give me much clues, but that package is for the Gnu version of which? While busybox which is probably builtin then.

`which` of GNU Coreutils, `which` of Busybox and `which` of MIT licensed Rust rewrite of uutils (Ubuntu is switching to it). Finally I have a good reason to to break all of English language grammar and say: `Which which which works` \o/

alimirjamali:
PipelineRetryFailed

Minimalist73:
I'm currently on Qubes 4.2 and the `-q` option of `which` doesn't exist in dom0:
```
$ which -q qvm-backup-restore
/usr/bin/which: invalid option -- 'q'
/usr/bin/qvm-backup-restore
```

Is it different on 4.3? I tried inside a Fedora 41 app qube but it's also missing from there so I'm not sure.

alimirjamali:
> I'm currently on Qubes 4.2 and the `-q` option of `which` doesn't exist in dom0:
> 
> ```
> $ which -q qvm-backup-restore
> /usr/bin/which: invalid option -- 'q'
> /usr/bin/qvm-backup-restore
> ```
> 
> Is it different on 4.3? I tried inside a Fedora 41 app qube but it's also missing from there so I'm not sure.

Yes. You are absolutely right. The `-q` is not a `which` option at all. I must have mixed it up with `grep -q`.

Anyways, I switched the check from `which` to `command -v` as was suggested by @ben-grande . One less call to external binaries. And `command -v` works in both `bash` and `zsh` (which is the default in Whonix templates)

marmarek (approved):

PR #365, opened 2025-06-08
qvm_ls.py : add template tree mode

RandyTheOtter:
Add template tree mode (`-T`), because I like it.

Almost exactly as network tree, but returns tree of templates, i.e.:
```
NAME                             STATE    CLASS         LABEL   TEMPLATE                 NETVM
sys-tmpl                         Halted   TemplateVM    black   -                        -
└─sys-net                        Running  AppVM         red     sys-tmpl                 -
└─sys-dvm                        Halted   AppVM         red     sys-tmpl                 sys-firewall
  └─sys-usb                      Running  DispVM        red     sys-dvm                  -
  └─sys-firewall                 Running  DispVM        green   sys-dvm                  sys-net
```

This is by no means complete, I pretty much just crammed my code the same way `sort_to_tree` is integrated into the program.

### To do:
- [ ] Edit manual once we agree on UX

### Maybe do, if people want it:
- [ ] Choose a good exception for the while loop getting stuck / somehow get rid of the while loop
- [ ] Find a better way to integrate this functionality. Having multiple tree "modes" seems more logical than two separate tree-producing functions.
- [ ] UX: What flag should be used, and how? Again, maybe something similar to `--tree TYPE` is a better choice than `--tree` and `--ttree`?

### What I would personally like to do, but probably won't get there:
- [ ] Make better tree-drawing algorithm, this one isn't very comprehensible. Programs like `tree` do it better.

I don't have a dev setup, so no testing with `main` or `release4.3` can be done by me, but similar change to 4.2 version of `qvm_ls.py` works.
(Is there a way to somehow install these packages over 4.2 stuff without breaking anything?)

Also, I've noticed that `sort_to_tree` says that it returns a list of sets, which isn't true, it returns a list of tuples.

### DIFF
--- a/qubesadmin/tools/qvm_ls.py
+++ b/qubesadmin/tools/qvm_ls.py
@@ -404,14 +404,15 @@ class Table(object):
     :param list colnames: Names of the columns (need not to be uppercase).
     '''
     def __init__(self, domains, colnames, spinner, *, raw_data=False,
-                tree_sorted=False, sort_order='NAME', reverse_sort=False,
-                ignore_case=False):
+                tree_sorted=False, ttree_sorted=False, sort_order='NAME',
+                reverse_sort=False, ignore_case=False):
         self.domains = domains
         self.columns = tuple(Column.columns[col.upper().replace('_', '-')]
                 for col in colnames)
         self.spinner = spinner
         self.raw_data = raw_data
         self.tree_sorted = tree_sorted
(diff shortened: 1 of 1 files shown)


RandyTheOtter:
Closing because I:
1. forgot about this
2. wrote alternative to qvm-ls for my needs
3. think this is not good

PR #370, opened 2025-07-07
Documentation to volume `revisions_to_keep == -1`

alimirjamali:
Also better explanation for `qvm-volume info VMNAME:VOLUMENAME` if `revisions_to_keep == -1`.

Related: https://github.com/QubesOS/qubes-issues/issues/8767

### DIFF
--- a/doc/manpages/qvm-pool.rst
+++ b/doc/manpages/qvm-pool.rst
@@ -68,6 +68,7 @@ drivers
 
 List all known drivers with their options.
 The listed driver options can be used with the ``-o options`` switch.
+See `qvm-volume(1)` manpage for detailed explanation of individual properties.
 
 aliases: d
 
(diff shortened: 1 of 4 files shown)


alimirjamali:
PipelineRetryFailed

alimirjamali:
Review notes: `revisions_to_keep` is not mentioned in `qvm-pool` manpage. It should be decided if the `revisions_to_keep == -1` should be be prettified in the output of `qvm-pool` or not. And should  it be mentioned in its manpage or not.

marmarek:
Yes, definitely should be added to the man page of qvm-pool. As for prettifying there, I'm not sure as some might parse this output (there is no printing single property there).

alimirjamali:
> Yes, definitely should be added to the man page of qvm-pool.

Instead of repeating everything once more in `qvm-pool` manpage, I added reference to `qvm-volume` manpage. Since explanation of other properties (e.g. `ephemeral_volatile`) is also missing from `qvm-pool` manpage. This proper explanations could be easily updated in one place.

marmarek (approved):

PR #375, opened 2025-07-16
Fix outdated cache on domain restart to fix audio-vm connection for restarted VMs

twd64:
The AudioVM connection for restarted Qubes was broken after the domain started. After some debugging, I found that the XID was not properly updated in the cache. As a result, the AudioVM daemon attempted to connect to the restarted Qube using the outdated XID.

I'm not certain if this is the appropriate place to address the issue, but clearing the cache resolves the problem.

Please let me know if there's a better way or place to handle this, and I’ll be happy to update the PR accordingly.

### DIFF
--- a/qubesadmin/events/__init__.py
+++ b/qubesadmin/events/__init__.py
@@ -264,7 +264,7 @@ class EventsDispatcher(object):
         elif event in ('domain-pre-start', 'domain-start', 'domain-shutdown',
                        'domain-paused', 'domain-unpaused',
                        'domain-start-failed'):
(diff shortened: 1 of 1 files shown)


marmarek:
There should be `property-reset:xid` event fired on domain start: https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/vm/qubesvm.py#L1462
It should invalidate relevant cache, just a few lines above your change. If that doesn't work, it need investigating why.

twd64:
Thanks for the pointer, my policies were wrong/out of date. It only allowed the domain-* events to my audio qube. Sorry for wasting your time ;).

PR #387, opened 2025-10-03
Add `qvm-restart`

### LINKED ISSUE #4747: [feature request] qvm-restart
An easy fix for most VM-internal problems is to reboot the VM. Especially with VMs supplying networking, USB-device or similar, this isn't a simple matter of `qvm-shutdown VM --wait && qvm-start VM`. What I'm hoping for is a command that restarts the VM and restores the state as good as possible. That means unplugging everything connected to the VM, unplugging everything the VM is connected to, shutting down (asking weather a kill is in order if that takes too long), rebooting and reconnecting everything.

In the case of networking VMs, it should be a simple matter of storing the connections, rebooting and reconnecting all connected VMs.
It's more tricky when block-devices are involved. E.g. a VM decrypting a block device and that being attached to another VM, Users have to take care of decryption themselves. I'm not sure weather a simple error after the fact ("can't reconnect DecryptVM:dm-0 to ConnectedVM: does not exist!") is okay, or a warning/conformation before restart is necessary ("You are about to restart a VM that supplies devices: dm-0 to ConnectedVM; continue?") 
Furthermore, I'm not sure weather a dispVM should actually restart, or by default shut down, be deleted, and another dispVM takes its place.

Since this is intended to recover from error-states, excluding devices to be attached after reboot might be a good idea, something like `--no-usb`; `--no-block`, `--no-pci`. (e.g. a network-vm with attached wifi-dongle might have issues because of it.)

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/4747

### DIFF
--- a/doc/manpages/qvm-restart.rst
+++ b/doc/manpages/qvm-restart.rst
@@ -1,0 +1,54 @@
+.. program:: qvm-restart
+
+:program:`qvm-restart` -- Restart selected or currently running qubes
+=====================================================================
+
+Synopsis
+--------
+
+:command:`qvm-restart` [-h] [--verbose] [--quiet] [--all] [--exclude *EXCLUDE*] [--timeout *TIMEOUT*] [*VMNAME*]
+
+Options
+-------
+
+.. option:: --help, -h
+
+   show the help message and exit
+
+.. option:: --verbose, -v
+
+   increase verbosity
+
+.. option:: --quiet, -q
+
+   decrease verbosity
+
+.. option:: --all
+
+   perform the action on all running qubes except dom0 & unnamed DispVMs
+
+.. option:: --exclude=EXCLUDE
+
+   exclude the qube from :option:`--all`
+
+.. option:: --timeout
+
+   timeout after which domains are killed. The default is decided by global
+   `default_shutdown_timeout` property and qube `shutdown_timeout` property
+
+.. option:: --version
+
+   Show program's version number and exit
+
+
+Authors
+-------
+
+| Joanna Rutkowska <joanna at invisiblethingslab dot com>
+| Rafal Wojtczuk <rafal at invisiblethingslab dot com>
+| Marek Marczykowski <marmarek at invisiblethingslab dot com>
+| Wojtek Porczyk <woju at invisiblethingslab dot com>
+
+| For complete author list see: https://github.com/QubesOS/qubes-core-admin-client.git
+
+.. vim: ts=3 sw=3 et tw=80
--- a/qubesadmin/tools/qvm_restart.py
+++ b/qubesadmin/tools/qvm_restart.py
@@ -1,0 +1,88 @@
+# encoding=utf-8
+#
+# The Qubes OS Project, http://www.qubes-os.org
+#
+# Copyright (C) 2010-2016  Joanna Rutkowska <joanna@invisiblethingslab.com>
+# Copyright (C) 2011-2025  Marek Marczykowski-Górecki
+#                                              <marmarek@invisiblethingslab.com>
+# Copyright (C) 2016       Wojtek Porczyk <woju@invisiblethingslab.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation; either version 2.1 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
+"""Restart a qube / qubes"""
+
+import sys
+
+try:
+    import qubesadmin.events.utils
+
+    have_events = True
+except ImportError:
+    have_events = False
+import qubesadmin.tools
+from qubesadmin.tools import qvm_start, qvm_shutdown
+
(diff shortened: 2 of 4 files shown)


qubesadmin/tools/qvm_shutdown.py
```diff
@@ -74,6 +83,17 @@ def main(args=None, app=None):  # pylint: disable=missing-docstring
 
     force = args.force or bool(args.all_domains)
 
+    running_vms = []
+    if args.restart:
+        args.wait = True
+        running_vms = [
+            vm.name
+            for vm in args.domains
+            if vm.get_power_state() == "Running"
```
ben-grande: Why only `Running`? What if I want to have a command that ensures that certain qubes are either running or with the latest state:

- `debian`: open the qube and do some changes
- `work`: state: Halted, template: debian
- `personal`: state: Running, template: debian

Hum, I will need to user work and personal now, let's make sure they are in their latest state:
```
qvm-shutdown --restart work personal
```

With the only running check, it would be a bit more longer:
```
qvm-shutdown --restart personal
qvm-start work
```

marmarek: Honestly, it's rather weird to use qvm-shutdown tool to _start_ a qube in your example...

marmarek: Maybe it should be `qvm-start --restart` instead?

alimirjamali: > Why only `Running`?

To make `qvm-shutdown --all --restart` a possibility. To make it possible to restart all currently running qubes.

> Maybe it should be `qvm-start --restart` instead?

After some feedback here and on Matrix channel, I believe a new dedicated `qvm-restart` tool is desirable.

@ArrayBolt3 and some other users had some good explanation on Matrix on why a dedicated tool is a good idea. I will amend this PR and come back.

ben-grande: > To make `qvm-shutdown --all --restart` a possibility. To make it possible to restart all currently running qubes.

-Restart all running qubes is `--all`;
- If `--all` is absent, shutdown and start or only start the specified qubes, being shutdown not enforced if the qube is Halted.

About other power states such as paused, I am not sure, possibly it shouldn't mess with those.

alimirjamali: @marmarek @ben-grande I rewrote a new dedicated `qvm-restart` tool among with manpage and unittests. I appreciate if you could review.

doc/manpages/qvm-restart.rst
```diff
@@ -0,0 +1,54 @@
...
+
+   increase verbosity
+
+.. option:: --quiet, -q
+
+   decrease verbosity
+
+.. option:: --all
+
+   perform the action on all running qubes except dom0 & true DispVMs
```
ben-grande: Is there a better phrase than "true" DispVMs? Maybe "unnamed DispVMs" or "DispVMs with `auto_cleanup` enabled"?

alimirjamali: > Is there a better phrase than "true" DispVMs? Maybe "unnamed DispVMs" or "DispVMs with `auto_cleanup` enabled"?

You are right. I agree with "unnamed DispVMs". I will amend the PR accordingly (I should also link the related Github issue).

ben-grande:
`qvm-unpause` also has `--all` if you want to do it.

doc/manpages/qvm-restart.rst
```diff
@@ -0,0 +1,54 @@
+.. program:: qvm-restart
```
marmarek, line 1: Man page is not installed, you need to add it to the list in `doc/conf.py`

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,88 @@
...
+        target_domains = [
+            vm
+            for vm in args.domains
+            if vm.get_power_state() == "Running"
+            and vm.klass != "AdminVM"
+            and not (vm.klass == "DispVM" and vm.auto_cleanup)
+        ]
+
+    # Forcing shutdown to allow graceful restart of ServiceVMs
+    shutdown_cmd = [vm.name for vm in target_domains] + ["--wait", "--force"]
```
marmarek, line 77: Automatic `--force` is a bad idea. While user might want to use it when restarting `sys-firewall` for example, the flag will override also other checks - for example qube being used as audiovm or guivm, where restart might not be as seamless as with netvm.

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,88 @@
...
+    shutdown_cmd += ["--timeout", str(args.timeout)] if args.timeout else []
+    shutdown_cmd += ["--quiet"] if args.quiet else []
+    qvm_shutdown.main(shutdown_cmd, app=args.app)
```
marmarek, line 80: I understand the convenience, but error handling here will be bad - if at least one qube fails to shutdown (and something fails when forcefully killing it on shutdown), no qube will be started again. Quite bad if one of restarted qubes was sys-usb, and you have USB keyboard...

Better to handle shutdown here directly, and then start the qubes that were powered off properly, and skip those that failed.

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,88 @@
...
+
+    start_cmd = [vm.name for vm in target_domains]
+    start_cmd += ["--quiet"] if args.quiet else []
+    qvm_start.main(start_cmd, app=args.app)
```
marmarek, line 84: Similarly to the above, error handling will be bad when doing it this way. `qvm-start` will stop at first startup failure. This might be okay when just starting qubes, but not really on restart.

marmarek (changes requested):

marmarek:
ping?

marmarek:
This continues in #470

PR #389, opened 2025-10-11
Exclude preloaded disposables from `qvm-run --all`

### LINKED ISSUE #10314: `qvm-run --all` will target preloaded disposables
### Qubes OS release

Qubes OS 4.3

### Brief summary

The `qvm-run` program has an `--all` option that runs commands on all running qubes. Preloaded disposables will become used when unpaused by `qvm-run`

### Steps to reproduce

1. Preload disposables
2. Run `qvm-run --all ...`
3. Preloaded disposables will be marked as used

### Expected behavior

Do not mess with preloaded disposables.

### Actual behavior

Runs on preloaded disposables.

### Additional information

I think that qubes that have this property or state should be excluded:

- When state `is_paused()` is True;
- When the `is_preload` property enabled, which may not be paused yet.

alimirjamali:
Also exclude all other internal qubes from the same command

fixes: https://github.com/QubesOS/qubes-issues/issues/10314

### DIFF
--- a/qubesadmin/tools/qvm_run.py
+++ b/qubesadmin/tools/qvm_run.py
@@ -370,7 +370,13 @@ def main(args=None, app=None):
         domains = [dispvm]
     elif args.all_domains:
         # --all consider only running VMs
-        domains = [vm for vm in domains if vm.is_running()]
+        domains = [
+            vm
+            for vm in domains
+            if vm.is_running()
+            and not vm.features.get("internal")
+            and not vm.is_paused()
+        ]
         if args.gui is None:
             for qube in domains:
                 gui_per_domain[qube] = has_gui(qube)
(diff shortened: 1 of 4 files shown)


qubesadmin/tools/qvm_run.py
```diff
@@ -370,7 +370,13 @@ def main(args=None, app=None):
         domains = [dispvm]
     elif args.all_domains:
         # --all consider only running VMs
-        domains = [vm for vm in domains if vm.is_running()]
+        domains = [
+            vm
+            for vm in domains
+            if vm.is_running()
+            and not vm.features.get("internal")
+            and vm.get_power_state() != "Paused"
```
ben-grande: All fine. I agree that querying internal is better than querying preloads. Nitpick but I prefer `vm.is_paused()` rather than string matching because linters can now if a method exists or not while string matching depends on having typed the correct string. Yes, unittest detects if it is wrong in this case.

Thanks for the PR

alimirjamali: > I prefer `vm.is_paused()` rather than string matching

This is done.

Thank you for review

ben-grande:
Can you please add in the commit message:

```
```

alimirjamali:
> Can you please add in the commit message:
> 
> ```
> For: https://github.com/QubesOS/qubes-issues/issues/1512
> ```

Done

ben-grande:
`qvm-unpause` also has `--all` if you want to do it.

alimirjamali:
> `qvm-unpause` also has `--all` if you want to do it.

This is also done (including the relevant unittest).

marmarek (approved):

PR #390, opened 2025-10-13
Ignore preload as a dependency

### LINKED ISSUE #10227: Can't change property (referencing a qube) or state of qube when there are preloaded disposables in chain
### Qubes OS release

Qubes OS 4.2

### Brief summary

Changing template is blocked if there are descendants running. In this case, preloaded disposables are always up and many current actions need a workaround that should be done transparently.

Code changes in

- qubes-core-admin
- qubes-manager

### Steps to reproduce

1. `qubes-prefs default-template fedora-42-xfce`
2. `qvm-features dom0 preload-dispvm-max 1`
3. `qubes-prefs default-template debian-13-xfce`

### Expected behavior

Able to switch templates if the only running qubes are preloaded disposables.

### Actual behavior

Not able to switch templates.

https://openqa.qubes-os.org/tests/152980#step/switch_template/13

### Additional information

- [x] Changing disposable template property `template_for_dispvms`
- [x] Changing template of disposable template
- [x] Renaming template via qube manager
- [x] Renaming disposable template via qube manager
- [x] Removing disposable template via qube manager
- [x] Shutting down AudioVM
- [x] Shutting down NetVM

ben-grande:
There doesn't seem to be a sane setup that would set a preloaded disposable as a property of another qube or a system property. The core already ignores preloaded disposables in certain cases, do the same on the client to not have derived applications such as Qube Manager having to deal with it.

Fixes: https://github.com/QubesOS/qubes-issues/issues/10227

### DIFF
--- a/qubesadmin/tests/utils.py
+++ b/qubesadmin/tests/utils.py
@@ -35,7 +35,10 @@ class TestVMUsage(qubesadmin.tests.QubesTestCase):
             b'template2 class=TemplateVM state=Running\n' \
             b'vm2 class=AppVM state=Running\n' \
             b'sys-net class=AppVM state=Running\n' \
-            b'sys-firewall class=AppVM state=Running\n'
+            b'sys-firewall class=AppVM state=Running\n' \
+            b'test-dvm class=AppVM state=Running\n' \
+            b'disp1 class=DispVM state=Running\n' \
+            b'disp2 class=DispVM state=Running\n'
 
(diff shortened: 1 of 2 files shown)


ben-grande:
PipelineRetryFailed

ben-grande:
Didn't identify any related failure on openqa.

marmarek (approved):

PR #391, opened 2025-10-15
Document `servicevm` feature

alimirjamali:
Since setting the `servicevm` feature is not done automatically anymore, let the user know about its existence. User might decide to set it for their other desired qubes (e.g. sys-nas, sys-backup, ...)

related: https://github.com/QubesOS/qubes-issues/issues/7298

### DIFF
--- a/doc/manpages/qvm-service.rst
+++ b/doc/manpages/qvm-service.rst
@@ -219,6 +219,14 @@ custom-persist:
     This will also disable ``/home`` and ``/usr/local`` persistence as well as
     ``/rw/config/*`` automatically executed configs.
 
+servicevm
+    Default: disabled
+
+    Changes the qube icon to ServiceVM theme and moves the qube to ``SERVICE``
+    tab of Qubes Application Menu.
+    If the qube ``provides_network`` property set/reset to True or False,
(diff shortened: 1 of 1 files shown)


alimirjamali:
PipelineRetry

doc/manpages/qvm-service.rst
```diff
@@ -219,6 +219,14 @@ custom-persist:
     This will also disable ``/home`` and ``/usr/local`` persistence as well as
     ``/rw/config/*`` automatically executed configs.
 
+servicevm
+    Default: enabled for sys-usb, sys-net, sys-firewall, sys-audio
```
marmarek: The "default" section describes what happens if the service is not set at all (either enabled or disabled). So, this list isn't really necessary here, because service is explicitly enabled for them. But a comment about relation to provides_network might be relevant here.

marmarek (changes requested):

doc/manpages/qvm-service.rst
alimirjamali: Ok. I changed the "default" to "disabled"

Also keeping the note on the relation with the icon and App Menu.

PR #392, opened 2025-10-27
Remove `--just-db` option from auto-generated man of qvm-remove

jamke:
Remove `--just-db` option from auto-generated man of qvm-remove.
I checked the code, qvm-remove does not support this option.

### DIFF
--- a/doc/manpages/qvm-remove.rst
+++ b/doc/manpages/qvm-remove.rst
@@ -14,7 +14,7 @@
 
 Synopsis
 --------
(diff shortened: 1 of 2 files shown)


jamke:
Unintentionally made a second commit (Fix mistake in man qvm-template: use info instead of list).
Reverted it.

marmarek:
Is that second change unwanted? Looks like a good thing to fix too.

jamke:
> Is that second change unwanted? Looks like a good thing to fix too.

I wanted to PR it later one by one. The second change is valid and correct, though. You may use it too if it's OK to combine changes like that.

marmarek:
I merged the two fixes without the revert. Thanks!

PR #393, opened 2025-10-30
Cache GUIVM keyboard layout in `qvm-start-daemon`

### LINKED ISSUE #6517: Keyboard layout switching unreliable
### Qubes OS version:

R4.1

### Affected component(s) or functionality:

GUI

---

keyboard layout switcher is ru++grp:shifts_toggle

### Expected or desired behavior:

shift+shift consistently switches layouts in any VM

### Actual behavior:

layout switcher is reliable in dom0 but not in domU, sometimes it is not changed at all and gets out of sync with xfce layout indicator if there is one in place.

### General notes:

Apparently turning HT on makes things worse but I am not sure

---

### I am aware of the following related, [non-duplicate](https://www.qubes-os.org/doc/reporting-bugs/#new-issues-should-not-be-duplicates-of-existing-issues) issues:

https://github.com/QubesOS/qubes-issues/issues/1396

### LINKED ISSUE #8441: `XKLAVIER_ALLOW_SECONDARY` property changed twice with a time gap, that breaks layout switching sometimes

### Qubes OS release
R4.1 (probably R4.2, too)

### Brief summary
Keyboard layout switch shortcut causes property change of `XKLAVIER_ALLOW_SECONDARY` to be fired twice, and not in one moment, but sometimes even with time gap up to a second.
It breaks keyboard layout, at least in case the user switches between active windows in this time gap.

### Steps to reproduce

Currently in R4.1 (and R4.2) keyboard layout switch is applied and propagated by reacting on `XKLAVIER_ALLOW_SECONDARY` property being changed in `dom0`, these changes can be monitored with command in `dom0`:
`$ xev -root -event property`

Keyboard layout switch shortcut causes property change of `XKLAVIER_ALLOW_SECONDARY` appear in the output of this command. That is expected. But it happens twice! And not in one moment, but sometimes even with time gap up to a second. It breaks keyboard layout switch sometimes, if user switches between active windows at that moment. Because the mechanism of propagation also happens twice and enforces possibly wrong (outdated) layout to a possibly wrong qube.

### Expected behavior
`XKLAVIER_ALLOW_SECONDARY` property change event fired once, code in `dom0`instantly finds the active window, active qube and passes the correct layout there.

### Actual behavior
`XKLAVIER_ALLOW_SECONDARY` property change event fired twice, with a time gap and sometimes sets wrong layout in the qube.

### More: 
Topic on forum if useful
https://forum.qubes-os.org/t/why-xev-root-event-property-shows-duplicate-events-it-breaks-keyboard-layout/19656

### Log
The example of output from `xev` I provided in the next message, as github has no collapsible spoilers.

alimirjamali:
On XFCE systems with multiple keyboard layouts and with XFCE Keyboard Layout widget, X raises two `XKLAVIER_ALLOW_SECONDARY` events shortly after each other (see 1st issue).

The current intended behaviour is to ignore the 2nd event by checking if the GUIVMs current layout is similar to the new requested layout. Unfortunately, after setting the GUIVM's keyboard layout, Qubes Core goes to this loop to propagate the new layout to all child qubes: https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/ext/gui.py#L162-L170

And even if we move the condition after that loop just before it, it is still too slow to report back the new layout via qubesdb for `qvm-stat-daemon` use.

In order to avoid this race condition, it would be better to cache the layout within `qvm-start-daemon` itself.

fixes: https://github.com/QubesOS/qubes-issues/issues/8441
fixes: https://github.com/QubesOS/qubes-issues/issues/6517

### DIFF
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -363,6 +363,7 @@ class KeyboardLayout:
         self.options = ",".join(
             opt for opt in self.options.split(",") if not opt.startswith("grp:")
         )
+        self.current_layout = None
 
     def get_property(self, layout_num):
         """Return the selected keyboard layout as formatted for keyboard_layout
@@ -449,10 +450,9 @@ class XWatcher:
         """Update current vm's keyboard_layout property"""
         new_property = self.keyboard_layout.get_property(self.selected_layout)
 
-        current_property = self.current_vm.keyboard_layout
-
-        if new_property != current_property:
+        if new_property != self.keyboard_layout.current_layout:
             self.current_vm.keyboard_layout = new_property
+            self.keyboard_layout.current_layout = new_property
 
     def event_reader(self, callback):
         """Poll for X events related to keyboard layout"""


alimirjamali:
PipelineRetryFailed

marmarek (approved):

PR #400, opened 2025-12-08
Return UnknownDevice for assignments where device is unknown

marmarta:
references QubesOS/qubes-issues#10409

### DIFF
--- a/qubesadmin/device_protocol.py
+++ b/qubesadmin/device_protocol.py
@@ -1364,8 +1364,9 @@ class DeviceAssignment:
     @property
     def device(self) -> DeviceInfo:
         """
-        Get single DeviceInfo object or raise an error.
-
+        Get single DeviceInfo object or UnknownDevice, if the device has
+         not been found.
+        If there are more devices than one matching, raise ProtocolError.
         If port id is set we have exactly one device
         since we can attach ony one device to one port.
         If assignment is more general we can get 0 or many devices.
@@ -1375,7 +1376,8 @@ class DeviceAssignment:
             return devices[0]
         if len(devices) > 1:
             raise ProtocolError("Too many devices matches to assignment")
-        raise ProtocolError("No devices matches to assignment")
+        return UnknownDevice(port=self.port,
+                             device_id=self.device_id)
 
     @property
     def port(self) -> Port:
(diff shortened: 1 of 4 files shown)


qubesadmin/device_protocol.py
```diff
@@ -1367,7 +1367,9 @@ def device(self) -> DeviceInfo:
             return devices[0]
         if len(devices) > 1:
             raise ProtocolError("Too many devices matches to assignment")
-        raise ProtocolError("No devices matches to assignment")
+        return UnknownDevice(port=self.port,
+                             device_id=self.device_id,
+                             attachment=self)
```
marmarek: First, check carefully places where this attribute is used (in global config, VM settings, widgets, ansible etc). The behavior change may be unexpected in some, especially since the old behavior was explicitly documented in the docstring.

Then, update docstring.

And finally, I don't think `attachment` attribute should be set. Its docstring says "VM to which device is attached (frontend domain)", which cannot be true for not present device.

marmarek (changes requested):

qubesadmin/device_protocol.py
marmarta: I think you're implying this is a bad idea? I think it makes sense not to raise a ProtocolError if we have an UnknownDevice for those purposes, but I can try to think about tacking this problem in another way.

(and about the attachment attribute, it is incorrect to set here self, yes, but this is a property of a DeviceAssignment that knows its frontend domain, shouldn't we set it?)

marmarek: I think the change makes sense in general. But since it's behavior change, it needs at least checking if other places don't rely on the old behavior. For example `attach_device` in qubesadmin/tools/qvm_device.py seems to rely on this exception. I'm sure there are other places too (maybe also in other repositories...)

> (and about the attachment attribute, it is incorrect to set here self, yes, but this is a property of a DeviceAssignment that knows its frontend domain, shouldn't we set it?)

"attachment" is about where the device is _attached_ (like, currently connected), not _assigned_ (potentially attached, automatically or not, may be even multiple assignments for the same device)

marmarek: Oh, and note this file is also copied to core-admin, so check if that side doesn't rely on this exception too.

marmarta:
I tried to find uses of this in other modules, didn't find reliance on the exception outside of core-admin (https://github.com/QubesOS/qubes-core-admin/pull/799)  and core-admin-client.

marmarta:
I'm not sure here; the tests failed when I did a change to react to this change in qvm_detach, but I'm not sure if it's not because the tests do not take into account the concurrent change in core-admin?

marmarek (approved):

PR #401, opened 2025-12-09
Add method to solely create disposable

ben-grande:
In line with the qubes module, disposable creation happens on from_appvm(). This change is intended to measure how much time it takes to get a fresh disposable object, improving the performance tests distinction of the from_appvm() phase from the execution/run*() phase.

For: https://github.com/qubesos/qubes-issues/issues/10230
For: https://github.com/qubesos/qubes-issues/issues/1512

### DIFF
--- a/qubesadmin/vm/__init__.py
+++ b/qubesadmin/vm/__init__.py
@@ -454,34 +454,31 @@ class QubesVM(qubesadmin.base.PropertyHolder):
 
 
 class DispVMWrapper(QubesVM):
-    """Wrapper class for new DispVM, supporting only service call
+    """
+    Wrapper class for new disposable, supporting only service call.
 
-    Note that when running in dom0, one need to manually kill the DispVM after
-    service call ends.
+    Note that when running in dom0, one need to manually kill the disposable
+    after service call ends.
     """
 
     def run_service(self, service, **kwargs):
-        if self.app.qubesd_connection_type == "socket":
-            # create dispvm at service call
-            if self._method_dest.startswith("$dispvm"):
-                if self._method_dest.startswith("$dispvm:"):
-                    method_dest = self._method_dest[len("$dispvm:") :]
-                else:
-                    method_dest = "dom0"
-                dispvm = self.app.qubesd_call(
-                    method_dest, "admin.vm.CreateDisposable"
-                )
-                dispvm = dispvm.decode("ascii")
-                self._method_dest = dispvm
-                # Service call may wait for session start, give it more time
-                # than default 5s
-                kwargs["connect_timeout"] = self.qrexec_timeout
+        """Create disposable if absent and run service."""
+        if (
+            self.app.qubesd_connection_type == "socket"
+            and self._method_dest.startswith("$dispvm")
+        ):
+            self.create_disposable()
+            # Service call may wait for session start, give it more time
+            # than default 5s
+            kwargs["connect_timeout"] = self.qrexec_timeout
         return super().run_service(service, **kwargs)
 
     def cleanup(self):
-        """Cleanup after DispVM usage"""
-        # in 'remote' case nothing is needed, as DispVM is cleaned up
-        # automatically
+        """
+        Cleanup after disposable usage.
+
+        Disposable is cleaned up automatically in 'remote' case.
+        """
         if (
             self.app.qubesd_connection_type == "socket"
             and not self._method_dest.startswith("$dispvm")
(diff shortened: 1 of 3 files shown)


qubesadmin/vm/__init__.py
```diff
@@ -514,11 +485,11 @@ class DispVM(QubesVM):
...
-
         if appvm:
-            method_dest = "$dispvm:" + str(appvm)
+            method_dest = str(appvm)
         else:
-            method_dest = "$dispvm"
-
-        wrapper = DispVMWrapper(app, method_dest)
+            method_dest = "dom0"
+        dispvm = app.qubesd_call(method_dest, "admin.vm.CreateDisposable")
```
marmarek: This is a bad idea. Normal qubes intentionally have no permission to call `admin.vm.CreateDisposable` (nor any other `admin.*` calls), but they may be granted calls to the `@dispvm` target (possibly choosing different disposable template depending on the requested service).

Requiring extended admin API access just to measure time a bit better is no-go...

ben-grande: Background:

```
commit 9bb59cdd20fa551563a04346559be2f09500708a
Author: Marek Marczykowski-Górecki <marmarek@invisiblethingslab.com>
Date:   Sun Aug 6 12:22:47 2017 +0200

    vm: add DispVMWrapper for calling a single service in new DispVM

    This is a wrapper to use `$dispvm` target of qrexec call, just like any
    other service call in qubesadmin module - using vm.run_service().
    When running in dom0, qrexec-client-vm is not available, so DispVM needs
    to be created "manually", using appropriate Admin API call
    (admin.vm.CreateDisposable).

    QubesOS/qubes-issues#2974
```

I did not comprehend why using `$dispvm` was necessary, thought it served the necessity at that time only.

> This is a bad idea. Normal qubes intentionally have no permission to call `admin.vm.CreateDisposable` (nor any other `admin.*` calls), but they may be granted calls to the `@dispvm` target (possibly choosing different disposable template depending on the requested service).

I'm not sure I completely understand this. What you are saying is that this is problematic when core-admin-client is running on a domU?

> Requiring extended admin API access just to measure time a bit better is no-go...

What about a using the old code, maintaining the use of `$dispvm*`, and adding method just to create the disposable `if self._method_dest.startswith("$dispvm")`?

```python3
class DispVMWrapper(QubesVM):
    """Wrapper class for new DispVM, supporting only service call

    Note that when running in dom0, one need to manually kill the DispVM after
    service call ends.
    """

    def create_disposable(self):
        """Create dispvm at service call or run service directly."""
        if self._method_dest.startswith("$dispvm"):
            if self._method_dest.startswith("$dispvm:"):
                method_dest = self._method_dest[len("$dispvm:") :]
            else:
                method_dest = "dom0"
            dispvm = self.app.qubesd_call(
                method_dest, "admin.vm.CreateDisposable"
            )
            dispvm = dispvm.decode("ascii")
            self._method_dest = dispvm
        return self

    def run_service(self, service, **kwargs):
        """Create disposable if absent and run service."""
        if (
            self.app.qubesd_connection_type == "socket"
            and self._method_dest.startswith("$dispvm")
        ):
            self.create_disposable()
            # Service call may wait for session start, give it more time
            # than default 5s
            kwargs["connect_timeout"] = self.qrexec_timeout
        return super().run_service(service, **kwargs)

    def cleanup(self):
        """Cleanup after disposable usage."""
        # in 'remote' case nothing is needed, as DispVM is cleaned up
        # automatically
        if (
            self.app.qubesd_connection_type == "socket"
            and not self._method_dest.startswith("$dispvm")
        ):
            try:
                self.kill()
            except qubesadmin.exc.QubesVMNotRunningError:
                pass

    def start(self):
        """Create disposable if absent and start it."""
        if self._method_dest.startswith("$dispvm"):
            self.create_disposable()
        super().start()


class DispVM(QubesVM):
    """Disposable VM"""

    @classmethod
    def from_appvm(cls, app, appvm):
        """Returns a wrapper for calling service in a new DispVM based on given
        AppVM. If *appvm* is none, use default DispVM template"""

        if appvm:
            method_dest = "$dispvm:" + str(appvm)
        else:
            method_dest = "$dispvm"

        wrapper = DispVMWrapper(app, method_dest)
        return wrapper
```

ben-grande: The solution above allows:

```
dispvm = qubesadmin.vm.DispVM.from_appvm(app, appvm).create_disposable()
```

marmarek: > What you are saying is that this is problematic when core-admin-client is running on a domU?

Yes.

> What about a using the old code, maintaining the use of `$dispvm*`, and adding method just to create the disposable `if self._method_dest.startswith("$dispvm")`?

This approach is okay, it will preserve using special dispvm target, but will allow explicitly created disposable if somebody really wants.

BTW, while at it, it would be good to change deprecated `$dispvm` to `@dispvm`.

ben-grande:
PipelineRetryFailed

ben-grande:
Openqa me please.

marmarek (approved):

PR #405, opened 2025-12-18
Introduce `audio-volume` feature

### LINKED ISSUE #2724: Disable speaker output for Whonix WS-based AppVMs by default
To prevent them from communicating to other mic-equipped devices and deanonymizing the user. (Note we do not expose mic to any AppVM by default since forever.)

Seems like we can easily do that during installation via Salt config. But maybe consider also some runtime check to ensure this also (in Qubes 4.x)?

/cc @adrelanos

alimirjamali:
Allowing user to set the default audio volume per qube via `audio-volume` feature. By default it is 100% (except some select qubes which will be always muted).

resolves: https://github.com/QubesOS/qubes-issues/issues/2724

### DIFF
--- a/doc/manpages/qvm-features.rst
+++ b/doc/manpages/qvm-features.rst
@@ -404,6 +404,14 @@ device instead of audio agent installed in the qube itself. The value is audio
 model to be emulated, supported values are `ich6`, `sb16`, `ac97`, `es1370`.
 Recommended is `ich6`. This is useful to get audio in a Windows qube.
 
+
+audio-initial-volume
+^^^^^^^^^^^^^^^^^^^^
+
+Optionally sets the audio volume of the qube at every launch. Valid values are
+from 0 to 150; or `mute`. Default for `untrusted` and Whonix workstation based
+qubes is mute.
+
 uefi
 ^^^^
 
--- a/qubesadmin/tools/qvm_start_daemon.py
+++ b/qubesadmin/tools/qvm_start_daemon.py
@@ -793,6 +793,26 @@ class DAEMONLauncher:
             str(self.pacat_domid(vm)),
             vm.name,
         ]
+
+        try:
+            volume = vm.features.check_with_template(
+                "audio-initial-volume",
+                self.app.domains[
+                    self.app.local_name
+                ].features.check_with_template(
+                    "audio-initial-volume",
+                    None,
+                ),
+            )
+            if volume:
+                if volume.endswith("%"):
+                    volume = volume[:-1]
+                pacat_cmd.insert(1, "-V")
+                pacat_cmd.insert(2, volume)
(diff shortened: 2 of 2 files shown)


marmarek:
pulseaudio / pipewire (or rather wireplumber) has similar feature already - it already remembers volume of each stream. You can adjust it in pavucontrol and it should be preserved across qube restart. Overriding it via qubes-specific feature isn't great, will lead to problems and unexpected situations (like needing to re-adjust volume at each qube start again, GUI tools for volume control no longer effective etc).

There is probably a way to set it also for future streams. Quick search suggests it's stored in `~/.local/state/wireplumber/stream-properties`, but editing it directly sounds like a very bad idea. Maybe `wpctl` or `pactl` can do that, or some other related tool.

alimirjamali:
> pulseaudio / pipewire (or rather wireplumber) has similar feature already - it already remembers volume of each stream. You can adjust it in pavucontrol and it should be preserved across qube restart. Overriding it via qubes-specific feature isn't great, will lead to problems and unexpected situations (like needing to re-adjust volume at each qube start again, GUI tools for volume control no longer effective etc).
> 
> There is probably a way to set it also for future streams. Quick search suggests it's stored in `~/.local/state/wireplumber/stream-properties`, but editing it directly sounds like a very bad idea. Maybe `wpctl` or `pactl` can do that, or some other related tool.

How should it be dealt with DispVMs based on `whonix-workstation-xx-dvm`? Shouldn't they be muted by default?

For those we do not want to set and we want pulse to remember the previous setting, I could easily adjust `pacat-simple-vchan` to pass NULL (like before) to `pa_stream_connect_playback` if `audio-volume` feature is not set. So the previous volume is preserved.

marmarek:
Should it be maybe called `audio-initial-volume`? It's just about the state at the qube startup after all.

And also, can you add some integration tests for this? See `qubes/tests/integ/audio.py` in core-admin. It's probably enough to check if audio stream is in desired state initially (especially for numerical value). But if want to check if really nothing got played, there are similar tests for microphone already (search for `audio_rec_muted`).

alimirjamali:
> Should it be maybe called `audio-initial-volume`? It's just about the state at the qube startup after all.

This is done. Salt formula PR is also amended accordingly

> And also, can you add some integration tests for this? See `qubes/tests/integ/audio.py` in core-admin. It's probably enough to check if audio stream is in desired state initially (especially for numerical value). But if want to check if really nothing got played, there are similar tests for microphone already (search for `audio_rec_muted`).

I submitted a patch to add the integration unittests. It currently only checks the audio volume and muting status:

https://github.com/QubesOS/qubes-core-admin/pull/768/files

alimirjamali:
PipelineRetryFailed

PR #407, opened 2026-01-07
Do not hide `qubes-backup-restore` working files

### LINKED ISSUE #10551: `qubes-backup-restore` maybe shouldn't hide the working files it places in `~/QubesIncoming`

### Qubes OS release

4.3

### Brief summary

After a fresh install of 4.3 I initiated a restore of a few templates and qubes to get my system going. It was taking a very long time. Hours! for just a few vms and <10 GB. The process tree wasn't changing so I declared it a failure and killed the program and investigated what went wrong.

What had happened was while waiting I noticed there was a seemingly empty `~/QubesIncoming` in dom0, nothing showing up in there from `ls` or `tree`, so I moved it to the trash as I was in a system cleanup / customization mindset. But, whoops! It actually wasn't empty. There was a hidden directory in there with a name like `.backup#restore-nkx7zprt` holding important restore-related disk images. Clearly the workspace for `qubes-backup-restore`. When I moved it to the trash I hung the restore.

If I had run `ls -A` or `tree -a` or `find`, that would have saved me from my mistake. If had run `rmdir` instead of `mv` (presumably also bad: `rm -r`), that would have saved me too.

### Steps to reproduce

- Initiate a restore
- Notice that there's a seemingly empty `~/QubesIncoming` in dom0
- Try to be tidy
- Wait for hours while the restore is hung (but appears productive)

### Expected behavior

- Simplest: probably the `.backup#restore-foobar` directory should not be hidden (no dot), so it's clear to the user that `~/QubesIncoming` (generally regarded as an ephemeral dir) is being used for something important
- Hard: `qubes-backup-restore` could detect the user has done something pretty dumb, and report.

### Actual behavior

- The progress bar continued to animate while I sat and waited, becoming increasingly concerned

alimirjamali:
Also put them all together within a dedicated subdirectory nested in `~/QubesIncoming` directory.

The mentioned tmp directory should be automatically cleaned up if the restore/verification operation is successful due to the patch for: https://github.com/QubesOS/qubes-issues/issues/8307

resolves: https://github.com/QubesOS/qubes-issues/issues/10551

### DIFF
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -954,13 +954,13 @@ class BackupRestore(object):
         if tmpdir is None:
             # put it here, to enable qfile-unpacker even on SELinux-enabled
             # system
-            tmpdir = os.path.expanduser("~/QubesIncoming")
+            tmpdir = os.path.expanduser("~/QubesIncoming/backup#restore")
         #: temporary directory used to extract the data before moving to the
         # final location
         # due to possible location in ~/QubesIncoming, the prefix should not be
         # a valid VM name
         os.makedirs(tmpdir, exist_ok=True)
-        self.tmpdir = tempfile.mkdtemp(prefix=".backup#restore-", dir=tmpdir)
+        self.tmpdir = tempfile.mkdtemp(dir=tmpdir)
 
         #: list of processes (Popen objects) to kill on cancel
         self.processes_to_kill_on_cancel = []


marmarek:
PipelineRetry

marmarek:
The directory was renamed to hidden in #351, but I cannot find explanation why, looks like unintentional change.

marmarek (approved):

marmarek:
> The mentioned tmp directory should be automatically cleaned up if the restore/verification operation is successful due to the patch for: [QubesOS/qubes-issues#8307](https://github.com/QubesOS/qubes-issues/issues/8307)

That issue was about App qubes, not dom0. The fix wasn't implemented in dom0 - but also, dom0 rarely has QubesIncoming directory - mostly (only?) after restoring a backup.

PR #409, opened 2026-01-21
Disentangle reusable template logic from `qvm-template` CLI

eaon:
This extracts the core template management functionality from `qvm_template.py` into a new `qubesadmin.templates` module, separating reusable library code from CLI-specific concerns like argument parsing and terminal output.

These changes primarily move code rather than modify it, meaning the logic should not have changed unless that was required for functionality extraction.

The original code had `parser.error()` calls scattered throughout, resulting in `sys.exit()`. Functions now signal errors and progress in Python rather than just in stdout/stderr. In short, `qvm-template-gui` could avoid parsing/showing the cli tool's output, as with this change it could use the module provide native Qt progress bars etc.

I did cursory testing (both on the cli and with `qvm-template-gui`) and didn't run into issues. Happy to do more if there's interest in merging this but you want more thorough manual testing. I couldn't find an issue that suggested this should happen.

### DIFF
--- a/qubesadmin/exc.py
+++ b/qubesadmin/exc.py
@@ -217,8 +217,13 @@ class BackupRestoreError(QubesException):
         super().__init__(msg)
         self.backup_log = backup_log
 
+
 class SignatureVerificationError(Exception):
     """Package signature is invalid or missing"""
+
+
+class AlreadyRunning(Exception):
+    """Another qvm-template is already running"""
 
 
 # pylint: disable=too-many-ancestors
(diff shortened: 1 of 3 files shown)


eaon:
I had avoided uploading my key to keyservers but now did it.

Edit: I'll address the lint/test issues in upcoming commits. Converted to draft for now.

marmarek:
Just to confirm: yes, I'm happy with the overall idea of this PR. But all tests need to pass.

ben-grande:
ping

eaon:
Hiyah, I'm still planning on getting back to this - technical issues and life got in the way. Should have some time later this week/early next week.

PR #410, opened 2026-01-22
restore: add hyphen before number suffix for conflicting name

rustybird:
See commit message

### DIFF
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -1594,7 +1594,7 @@ class BackupRestore(object):
         while (new_name in restore_info.keys() or
                new_name in [x.name for x in restore_info.values()] or
                new_name in self.app.domains):
-            new_name = str('{}{}'.format(orig_name, number))
+            new_name = str('{}-{}'.format(orig_name, number))
             number += 1
             if number == 100:
                 # give up


marmarek:
I'm pretty sure backup tests in https://github.com/QubesOS/qubes-core-admin/tree/main/qubes/tests/integ will need adjustment too. I see it at least at https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/tests/integ/backup.py#L641, but probably in few other tests too.

rustybird:
@marmarek 

Thanks, looks like it's just that one integration test. I'll adjust it.

Come to think of it, how about starting the numerical suffix from 2 instead of 1? IMO that would make more sense. (Or something like <code>vmname<em>-restore</em>-1</code> - but that seems maybe a little bit too long.)

marmarek:
> Come to think of it, how about starting the numerical suffix from 2 instead of 1?

Usually file managers in similar situation (file name conflicts) start the number suffix at 1.

rustybird:
Nautilus adds `(2)`, and Thunar adds `(copy 1)`. Both look fine to me because "the 2nd foo" is the "the 1st *copy of* foo". OTOH Firefox and Chromium really do just add `(1)` without any copy marker string for automatic downloads. I guess it's not that weird.

marmarek (approved):

PR #412, opened 2026-02-13
Add type hints for `qubesadmin/backup`

hippalectryon-0:
This PR should in principle add no functional changes.

The vast majority of changes are pure typing.
The rest are small refactors to help typing checkers, typically:
- explicitely indicating that `Popen` has `stdin/out`: `p.stderr` -> `typing.cast(IO, p.stderr)`
- using getattr / setattr on unexpected field accessions rather than `object.field`
- splitting variables when types are mixed and type checks complain
```python
# before
a: byte = some_value()
a = a.decode()  # a: str
f(a)
# after
a = some_value
a_str = a.decode()
f(a_str)
```
- etc...

Note: as noted in `pyproject.toml` this bumps the minimum python version to `3.10`.
Note: running a type checker on this code reveals many errors and warnings. After manual review, most of those errors are either 1) complex logic where the type checker is not able to reduce type unions (=> can be fixed by e.g. `assert var is not None`, but I want to keep that in a separate PR since that may add some functional changes or 2) what i believe to be actual errors in the code, e.g. value that actually can be None where they shouldn't be. Same, I'll keep that for a separate PR.

----

Some additional info concerning the type checking environment:

I use a combination of [ty](https://docs.astral.sh/ty/) and [ruff](https://docs.astral.sh/ruff/).

As written in `pyproject.toml`:
- `ruff` enforces that everything in `backup/` is type-checked (`select = ["ANN", "RUF045"]`), except for `**kwargs` (`select = ["ANN", "RUF045"]`) and `__init__` returns (`mypy-init-return = true`)
- `ty` is the actual type checker, with all the default rules

Note that `ANN` also enforces that by default `typing.Any` is forbidden, unless explicitly ignored `# noqa: ANN401`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,38 @@
+
+# Note: we're not using a "standard" pyproject.toml with a [project] section to avoid
+# clashing with the old-school setup.py build (for now... hopefully will be upgraded later ?)
+[dependency-groups]
+current-deps = [  # project dependencies
+    "lxml>=6.0.2",
+    "python-daemon>=3.1.2",
+    "pyxdg>=0.28",
+    "pyyaml>=6.0.3",
+    "tqdm>=4.67.3",
+    "xcffib>=1.12.0",
+]
+current-dev = [  # dev-depencencies
+    "types-lxml>=2026.1.1",
+    "pylint>=4.0.4",
+    "mock>=5.2.0",
+    "docutils>=0.22.4",
+    "sphinx>=9.1.0",
+    "setuptools>=82.0.0",
+    "black>=26.1.0",
+    "mypy>=1.19.1",
+]
+future-dev = [  # those tools are very efficient but not packaged for major distributions yet
+    "ruff>=0.15.1",
+    "ty>=0.0.16",
+]
+
+[tool.ty.rules]
+
+[tool.ruff]
+preview = true
+
+[tool.ruff.lint]
+select = ["ANN", "RUF045"]  # Require type annotations for every function
+ignore = ["ANN003"]  # type annotation for **kwargs
+
+[tool.ruff.lint.flake8-annotations]
+mypy-init-return = true  # do not require __init__ return annotation `->None` if there's at least one typed argument
--- a/qubesadmin/backup/__init__.py
+++ b/qubesadmin/backup/__init__.py
@@ -20,26 +20,29 @@
 
 '''Qubes backup'''
 import collections
+import io
+
+from qubesadmin.vm import QubesVM
 
 
 class BackupApp(object):
     '''Interface for backup collection'''
     # pylint: disable=too-few-public-methods
-    def __init__(self, qubes_xml):
+    def __init__(self, qubes_xml: str | None):
         '''Initialize BackupApp object and load qubes.xml into it'''
         self.store = qubes_xml
         self.domains = {}
         self.globals = {}
         self.load()
 
-    def load(self):
+    def load(self) -> bool | None:
         '''Load qubes.xml'''
         raise NotImplementedError
 
 class BackupVM(object):
     '''Interface for a single VM in the backup'''
     # pylint: disable=too-few-public-methods
-    def __init__(self):
+    def __init__(self) -> None:
         '''Initialize empty BackupVM object'''
         #: VM class
         self.klass = 'AppVM'
@@ -64,14 +67,14 @@ class BackupVM(object):
         self.size = 0
 
     @property
-    def included_in_backup(self):
+    def included_in_backup(self) -> bool:
         '''Report whether a VM is included in the backup'''
         return False
 
-    def handle_firewall_xml(self, vm, stream):
+    def handle_firewall_xml(self, vm: QubesVM, stream: io.BytesIO) -> None:
         '''Import appropriate format of firewall.xml'''
         raise NotImplementedError
 
-    def handle_notes_txt(self, vm, stream):
+    def handle_notes_txt(self, vm: QubesVM, stream: io.BytesIO) -> None:
         '''Import qubes notes.txt'''
         raise NotImplementedError  # pragma: no cover
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -274,7 +274,9 @@ class BackupHeader(object):
                     continue
                 f_header.write("{!s}={!s}\n".format(key, getattr(self, attr)))
 
-def launch_proc_with_pty(args: list[str], stdin: int | None=None, stdout: int | None=None, stderr: int | None=None, echo: bool=True) -> tuple[Popen, FileIO]:
+def launch_proc_with_pty(args: list[str], stdin: int | None=None,
+                         stdout: int | None=None, stderr: int | None=None,
+                         echo: bool=True) -> tuple[Popen, FileIO]:
     """Similar to pty.fork, but handle stdin/stdout according to parameters
     instead of connecting to the pty
 
@@ -299,7 +301,8 @@ def launch_proc_with_pty(args: list[str], stdin: int | None=None, stdout: int | 
     os.close(pty_slave)
     return p, open(pty_master, 'wb+', buffering=0)
 
-def launch_scrypt(action: str, input_name: str, output_name: str, passphrase: str) -> Popen:
+def launch_scrypt(action: str, input_name: str, output_name: str,
+                  passphrase: str) -> Popen:
     '''
     Launch 'scrypt' process, pass passphrase to it and return
     subprocess.Popen object.
@@ -356,10 +359,13 @@ def _fix_threading_after_fork() -> None:
 class ExtractWorker3(Process):
     '''Process for handling inner tar layer of backup archive'''
     # pylint: disable=too-many-instance-attributes
-    def __init__(self, queue: Queue, base_dir: str, passphrase: str, encrypted: bool, *,
+    def __init__(self, queue: Queue, base_dir: str, passphrase: str,
+                 encrypted: bool, *,
                  progress_callback: Callable, vmproc: Popen | None =None,
-                 compressed: bool=False, crypto_algorithm: str=DEFAULT_CRYPTO_ALGORITHM,
-                 compression_filter: str | None=None, verify_only: bool=False, handlers: dict[str, Callable] | None=None):
+                 compressed: bool=False,
+                 crypto_algorithm: str=DEFAULT_CRYPTO_ALGORITHM,
+                 compression_filter: str | None=None, verify_only: bool=False,
+                 handlers: dict[str, Callable] | None=None):
         '''Start inner tar extraction worker
 
         The purpose of this class is to process files extracted from outer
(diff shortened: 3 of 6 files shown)


hippalectryon-0:
Side note as a new contributor: manually fixing `pylint` errors is a big pain. I understand QubesOS development started a long time ago, but it would be very nice to have automated formatting tools (e.g. [ruff](https://docs.astral.sh/ruff/), or previously `black`), as do all modern projects.

andrewdavidwong:
Suggestion: Since this is for QubesOS/qubes-issues#10673, you can [automatically link this PR to that issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) by including the text `QubesOS/qubes-issues#10673` somewhere in the body of your PR message.

In the future, when you're eventually ready to close that issue, you can do so by including `Closes QubesOS/qubes-issues#10673` in the message of the PR that should close the issue. `Closes` is one of several [special keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) that GitHub will recognize. If and when that future PR is merged, GitHub will automatically close the associated issue.

qubesadmin/backup/restore.py
```diff
@@ -311,45 +324,48 @@ def launch_scrypt(action, input_name, output_name, passphrase):
...
+        actual_prompt = typing.cast(IO, p.stderr).read(len(prompt))
         if actual_prompt != prompt:
             raise QubesException(
                 'Unexpected prompt from scrypt: {}'.format(actual_prompt))
         pty.write(passphrase.encode('utf-8') + b'\n')
         pty.flush()
     # save it here, so garbage collector would not close it (which would kill
     #  the child)
-    p.pty = pty
+    setattr(p, 'pty', pty)
```
marmarek, line 335: Why?

pyproject.toml
```diff
@@ -0,0 +1,35 @@
+
+# Note: we're not using a "standard" pyproject.toml with a [project] section to avoid
+# clashing with the old-school setup.py build (for now... hopefully will be upgraded later ?)
+[tool.uv]
```
marmarek: Generally, we prefer to rely on distribution packages, also for development tools, not fetching from pypi or similar directly.
This is mostly for software supply chain reasons. While a lock file (like `uv.lock` you added here) might theoretically combat some of the attacks, reviewing this 800+ lines file and carefully checking if each hash corresponds to a legitimate release is not realistic. Distribution packagers do this work already.

marmarek:
For the tools choice, I'd prefer to keep it consistent across repositories. This means:
- pylint
- mypy
- black (with `-l80`)
- shellcheck (where relevant)

We work on many repositories at once, and switching between tools (slightly different styles enforced, different "special" comments etc) will be very annoying.

See for example core-admin or app-linux-split-gpg2 (the latter doesn't have black yet).

marmarek:
> manually fixing `pylint` errors is a big pain.

I agree fixing style manually is pain, and that's why we have black in some of the repositories already (this one has it applied to some files already, but it isn't enforced yet). But pylint isn't just about the code style, it also do quite a bit of static analysis (like checking variables existence, simple cases of resource leaks etc) - and those require manual fixes.

qubesadmin/backup/restore.py
hippalectryon-0: Because `pty` is not a normal field of `p` here. As indicated by the comment above, we are overloading it as a garbage collection hack, but type checkers will complain (and rightfully so).
Using setattr is more explicit with respect to what we are doing.

pyproject.toml
hippalectryon-0: I totally understand that point, but I still think we should include at least `pyproject.toml`.
- including the current file does *not* change anything to the package build process
- it enables us to declaratively define what versions of packages are supported. This in turn is used by automatic tools like IDEs to generate better warnings
- it allows identifying version conflicts, which the distribution packages will not warn about
- *if* anyone wishes to use package-resolution tools like `uv` (which are very handy), they have that option
- it's a modern, PEP-compliant, unified way to keep the settings of the tools (linter etc.) used, instead of having one `.toml` file per tool

marmarek: I'm okay with pyproject.toml in general. But I'm not okay with committing lock file with some arbitrary versions (hopefully not malicious ones, but honestly, even that is not easy to check). Since this isn't method we recommend, and use ourself, this file would also get stale very quickly.

What about using standard `[dependencies]` section? That should be compatible with any tool then, not just `uv`, right? And still put some discouraging comment (see for example `ci/requirements.txt` file).

hippalectryon-0:
> For the tools choice, I'd prefer to keep it consistent across repositories. This means:
> 
>     * pylint
> 
>     * mypy
> 
>     * black (with `-l80`)
> 
>     * shellcheck (where relevant)
> 
> 
> We work on many repositories at once, and switching between tools (slightly different styles enforced, different "special" comments etc) will be very annoying.
> 
> See for example core-admin or app-linux-split-gpg2 (the latter doesn't have black yet).

As a maintainer myself of other projects, I understand your point. Still, I would like to entertain the idea to consider more modern tools.
It is evident from the codebase of QubesOS that there are many remnants of very old (by my standards) coding practices. I understand not wanting to change some of those, but consider the following:
- `ty` is about 20x faster than `mypy` are functionally nearly equivalent for what we are doing
- `ruff` is ~300x faster than `pylint` and more complete. It does both the linting and error detection, and is also a replacement for `black`. It's a stable and very widespread tool that has become an industry standard.

From my experience, those speedups make a real difference. (e.g. concretely: running `pylint` on `qubes-core-admin-client` is terribly slow - running `ruff` is instantaneous.)

I think this is a strong choice for the project's long-term maintainability. Would it be possible to consider using the qubes-core-admin-client for an initial test before we proceed with broader adoption? I would be more than happy to assist and contribute my time.

marmarek:
ruff I see is packaged in Fedora, but in Debian it's only in unstable (not even testing).
With `ty` it's even worse - it seems to be in Fedora 44 (not released yet), but missing in the current stable Fedora, and not included in Debian at all.
At least both tools seems to be packaged in Arch, but it's just a community template, not the official one...

So, for now I'm afraid it isn't an option, at least as the primary tool. But, `ruff` claims to be compatible with black, so if one wishes so, I guess it can be used as an alternative (including relevant config in pyproject.toml if needed). But CI should enforce based on the "primary" tools.

pyproject.toml
hippalectryon-0: addcad3 I removed  `uv.lock` and `.python-version`.

> What about using standard [dependencies] section? That should be compatible with any tool then, not just uv, right? And still put some discouraging comment (see for example ci/requirements.txt file).

Ideally yes - unfortunately earlier tests showed that adding the dependencies the "standard" way

```
[project]
name = "your-project-name"
version = "0.1.0"
dependencies = [
    "lxml>=6.0.2",
    "python-daemon>=3.1.2",
    "pyxdg>=0.28",
    "pyyaml>=6.0.3",
    "tqdm>=4.67.3",
    "xcffib>=1.12.0",
]
```

makes `setup.py` change its build pattern to use `pyproject.toml`, which breaks the package.

This can likely be mitgated inside `setup.py`, but I think that's outside of this PR's scope for now

hippalectryon-0:
Understood. I'll raise those questions again when the packages become available.

Let me modify the current `pyproject.toml` accordingly

Note: despite mentions of `black` in `CHANGELOG`, I don't see where it's actually used anywhere in the repo ?

marmarek:
> we have black in some of the repositories already (this one has it applied to some files already, but it isn't enforced yet)

^

marmarek:
Basically, it's never a good time to run black on the whole repo, because it will require rebasing all open PRs (and in this repo always there are some). So, generally the approach is when modifying some file that isn't touched by another PR, run black on it first (in a separate commit!). Actually, backup-related files might fit this criteria right now.

hippalectryon-0:
499178f It's purely cosmetic since no CI uses `pyproject.toml` anyways, but I hope it's more aligned with the discussion so far.

>  Actually, backup-related files might fit this criteria right now.

I'll leave that to another PR ;)

Let me know if there's any remaining discussion points I've missed ?

qubesadmin/backup/__init__.py
```diff
@@ -20,26 +20,29 @@
...
-    def __init__(self, qubes_xml):
+    def __init__(self, qubes_xml: str | None):
         '''Initialize BackupApp object and load qubes.xml into it'''
         self.store = qubes_xml
         self.domains = {}
         self.globals = {}
         self.load()
 
-    def load(self):
+    def load(self) -> bool | None:
```
marmarek, line 38: The return value is never used, IMO better to drop the single `return False` (or even change it to an exception / or just re-raise the exception it's handling?), and make it `-> None`.

qubesadmin/backup/__init__.py
```diff
@@ -20,26 +20,29 @@
...
+import io
+
+from qubesadmin.vm import QubesVM
 
 
 class BackupApp(object):
     '''Interface for backup collection'''
     # pylint: disable=too-few-public-methods
-    def __init__(self, qubes_xml):
+    def __init__(self, qubes_xml: str | None):
```
marmarek, line 31: At this point maybe this can be just `str`? All the subclasses handle `None` case (with an exception) so here it will specific string. Alternatively, move that `None` handling here, to deduplicate code.

My main point here is to make `self.store` just `str`, not `str | None`, to better reflect reality (and avoid false positives from mypy down the road).

qubesadmin/backup/restore.py
marmarek, line 1141: After this line, `algorithm` is no longer `str | None`, but just `str`. Maybe add an isinstance assert or such?

marmarek (changes requested):
I've checked added type hints and generally they look good, generally they match the current code. There are bunch of cases, where a slight code change would make them more useful (for example eliminating `... | None` where it really can't happen). That would help type checker to produce less false positives. I've given few examples below.

Let me know what you think about it - would you like do changes like this in this PR, or merge this as-is and do a separate follow-up one? I'm okay with either approach.

The only real request I have in this PR, is to clean the commits history a bit - specifically, amend initial commit to remove uv.lock, instead of adding the file and removing it again later.

hippalectryon-0:
> Let me know what you think about it - would you like do changes like this in this PR, or merge this as-is and do a separate follow-up one? I'm okay with either approach.

For now my logic was to make non-breaking changes here, and add the breaking changes in #413 #415 . I agree with all your comments, so I'll modify those PRs accordingly.

hippalectryon-0:
> The only real request I have in this PR, is to clean the commits history a bit - specifically, amend initial commit to remove uv.lock, instead of adding the file and removing it again later.

I've force-pushed a clean(er) version

qubesadmin/backup/__init__.py
hippalectryon-0: 5a176af6249c549ba9114fa0db587278186d285b

Same for the two comments below

marmarek (approved):

PR #413, opened 2026-02-13
Enforce `is not None` for `core2.py`, fix type checker errors

hippalectryon-0:
Follow-up of #412 

This is a very small PR, since it is not clear to me how the maintainers want to proceed (=resolve the choices below) on this kind of work. I'm hoping this will give me a clearer view.

# Changes

## L188

No XML validation is performed whatsoever. As it stands, the code will crash if `default_template` is not defined. Even though *currently* this may not cause a crash, it's likely to produce one down the line if we change what we store in the xml.

Two options to fix it:
* enforce that it's not None `assert default_template is not None`. This is functionally equivalent to the current code, and "fixes" the type checker error, but will still throw an Exception.
* add a fallback `element.get("default_template", "none")`

I chose option 1 here for no good reason - this is not something I can decide myself.

## L270

Likewise, code will crash if `element.get('template_qid')` is not defined.

Same question regarding the resolution. Chose option 2 just to showcase it.

## L349

old code:
```python
        pcidevs = element.get('pcidevs')
        if pcidevs:
            pcidevs = ast.literal_eval(pcidevs)
        for pcidev in pcidevs:
```

this will trivially error if `pcidevs` is not in the XML. Proposed solution is to fallback to `[]`.

## L361

Will trivially fail if `self.store` is not defined

# Remaining errors

`uv run ruff check qubesadmin/backup/core2.py && uv run ty check qubesadmin/backup/core2.py` raises two remaining errors. Those errors are du to missing type hints in other files, and will be fixed down the line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,4 +6,4 @@ https://www.qubes-os.org/doc/admin-api/ for protocol specification.
 Compatibility
 =============
 
-This package requires Python >= 3.5
+This package requires Python >= 3.10
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ current-dev = [  # dev-depencencies
     "pylint>=4.0.4",
     "mock>=5.2.0",
     "docutils>=0.22.4",
-    "sphinx>=9.1.0",
+    "sphinx>=3.5.3",
     "setuptools>=82.0.0",
     "black>=26.1.0",
     "mypy>=1.19.1",
@@ -29,9 +29,11 @@ future-dev = [  # those tools are very efficient but not packaged for major dist
 
 [tool.ruff]
 preview = true
+line-length = 80
 
 [tool.ruff.lint]
-select = ["ANN", "RUF045"]  # Require type annotations for every function
+select = ["ANN", "RUF045", # Require type annotations for every function
+            "E501"]  # line-too-long
 ignore = ["ANN003"]  # type annotation for **kwargs
 
 [tool.ruff.lint.flake8-annotations]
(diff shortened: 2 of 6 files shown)


hippalectryon-0:
Note: additional changes since initial commits:

Since as reported in QubesOS/qubes-issues#10684 we already use python 3.12+ code in some places, I've bumped the minimum python version accordingly, which enables replacing `typing.TypeAlias` by `type ...=`

hippalectryon-0:
Note: the CI fails because of the issue raised in QubesOS/qubes-issues#10684 : even though we have 3.12+ bits of code, the environment itself only has python 3.10 installed, therefore Sphinx fails to parse new syntax.

I can easily revert my 3.12-specific changes to pass the CI, but that wouldn't resolve the underlying issue. I'd rather we settled the matter "officially".

marmarek:
Generally `core2.py` is about loading old backup from Qubes 3.x. That format isn't going to change ever, since that version isn't supported anymore for a long time. In fact, I doubt anybody will make new backups on that old version. But we should retain ability to load those old backups.
So, it's okay to add a bunch of asserts to make type checker happy, no need to make it future-proof.

hippalectryon-0:
Python version incompatibilities on this branch are fixed. Let me know if there's any blocking point remaining.

marmarek:
Do you think the remaining TODO comments are still applicable, given my response above?

hippalectryon-0:
Oops forgot about those - fixed

marmarek:
Same request as in the other PR - squash removing TODOs to not add them

hippalectryon-0:
done

marmarek (approved):

PR #415, opened 2026-02-13
Fix/enforce typing errors / inconsistencies in `backup/`

hippalectryon-0:
Follow-up of #413 

Most changes are in the same spirit. Below I list the changes where I need guidance.

# Logical mistake in `cleanup_tar2`

```python
        if terminate:
            if self.import_process is not None:
                self.tar2_process.terminate()
            self.import_process.terminate()
```

The last line will error if `self.import_process` is `None`.

I "fixed" it by moving it into the `if`, but need approval.

# `hmac_algorithm` could be `None`

In `BackupRestore > _verify_hmac() > load_hmac()` we have

```python
        if algorithm is None:
            algorithm = self.header_data.hmac_algorithm
        #...
        with open(os.path.join(self.tmpdir, filename), 'rb') as f_input:
            with subprocess.Popen(
                    ["openssl", "dgst", "-" + algorithm, "-hmac", passphrase],
```

this will error if `self.header_data.hmac_algorithm` is None.
I of course added `assert self.header_data.hmac_algorithm is not None` for the type checker, but am worried that this may be an actual bug, since it is perfectly legal afaik for `header_data.hmac_algorithm` to be `None`:

```python

    def __init__(self,
            header_data: bytes | None=None,
            *,
            version: int | None=None,
            encrypted: bool | None=None,
            compressed: bool | None=None,
            compression_filter: str | None=None,
            hmac_algorithm: str | None=None,
            crypto_algorithm: str | None=None,
            backup_id: str | int | None=None):
        # repeat the list to help code completion...
        self.version = version
        self.encrypted = encrypted
        self.compressed = compressed
        # Options introduced in backup format 3+, which always have a header,
        # so no need for fallback in function parameter
        self.compression_filter = compression_filter
        self.hmac_algorithm = hmac_algorithm
        self.crypto_algorithm = crypto_algorithm
        self.backup_id = backup_id

        init_supported_hmac_and_crypto()

        if header_data is not None:
            self.load(header_data)
```

# Possible `None` fields in `ExtractWorker3` args

In `BackupRestore > _start_inner_extraction_worker()` we have

```python
extractor_params = {
            'queue': queue,
            'base_dir': self.tmpdir,
            'passphrase': self.passphrase,
            'encrypted': self.header_data.encrypted,
            'compressed': self.header_data.compressed,
            'crypto_algorithm': self.header_data.crypto_algorithm,
            'verify_only': self.options.verify_only,
            'progress_callback': self.progress_callback,
            'handlers': handlers,
        }
        self.log.debug(
            'Starting extraction worker in %s, file handlers map: %s',
            self.tmpdir, repr(handlers))
        format_version = self.header_data.version
        if format_version in [3, 4]:
            extractor_params['compression_filter'] = \
                self.header_data.compression_filter
            if format_version == 4:
                # encryption already handled
                extractor_params['encrypted'] = False
            extract_proc = ExtractWorker3(**extractor_params)
```

All of the following: `header_data.crypto_algorithm, header_data.compressed, header_data.encrypted` can be `None` for the same reason as above.
However, `ExtractWorker3` will error if passed any of those as `None`.

# A comment on `qubesadmin.base.PropertyHolder`

Objects inheriting from `qubesadmin.base.PropertyHolder`, such as `QubesBase`, have a very strict rule that is not documented anywhere in their own docstring.

`PropertyHolder` does a very hack-y thing, which is that it considers that local-only parameters are the ones defined in the class attributes. In other terms,

```python
class QubesBase(qubesadmin.base.PropertyHolder):
   my_local_attribute = None
```

was the previous way to declare those. Except this isn't mentioned anywhere in the docstring of `QubesBase` itself - and due to the hack-y (and very unusual) implementation, it's very easy to break it inadvertently. I therefore added a small comment.

However, there remained a main issue: setting `my_local_attribute = None` is very bad for type checking (and overall), since if `my_local_attribute` is initialized in `__init__()`, it will never be`None`.
The correct way to do this is to type-check it instead:

```python
class QubesBase(qubesadmin.base.PropertyHolder):
   my_local_attribute: MyType
```

To make this work, I had to modify PropertyHolder accordingly:

```python
    def _local_properties(cls):
        '''
        Get set of property names that are properties on the Python object,
        and must not be set on the remote object
        '''
        if "_local_properties_set" not in cls.__dict__:
            props = set()
            for class_ in cls.__mro__:
                for key in class_.__dict__:
                    props.add(key)
                if hasattr(class_, "__annotations__"):  # NEW
                    for key in class_.__annotations__:  # NEW
                        props.add(key)  # NEW
            cls._local_properties_set = props

        return cls._local_properties_set
```

### DIFF
--- a/qubesadmin/backup/dispvm.py
+++ b/qubesadmin/backup/dispvm.py
@@ -175,6 +175,8 @@ class RestoreInDisposableVM:
 
     def transfer_pass_file(self, path: str) -> str:
         """Copy passhprase file to the DisposableVM"""
+        assert self.dispvm is not None
+
         subprocess.check_call(
             ['qvm-copy-to-vm', self.dispvm_name, path],
             stdout=subprocess.DEVNULL,
@@ -217,12 +219,16 @@ class RestoreInDisposableVM:
 
     def invalidate_backup_access(self) -> None:
         """Revoke access to backup archive"""
+        assert self.storage_access_proc is not None
+
         self.backup_storage_vm.tags.discard(self.storage_tag)
         typing.cast(typing.IO, self.storage_access_proc.stdin).close()
         self.storage_access_proc.wait()
 
     def prepare_inner_args(self) -> list:
         """Prepare arguments for inner (in-DispVM) qvm-backup-restore command"""
+        assert self.storage_access_id is not None
+
         new_options = []
         new_positional_args = []
 
@@ -269,6 +275,8 @@ class RestoreInDisposableVM:
 
     def extract_log(self) -> bytes:
         """Extract restore log from the DisposableVM"""
+        assert self.dispvm is not None
+
         untrusted_backup_log, _ = self.dispvm.run_with_args(
             'cat', self.backup_log_path,
             stdout=subprocess.PIPE,
@@ -278,11 +286,13 @@ class RestoreInDisposableVM:
 
     def run(self) -> bytes | None:
         """Run the backup restore operation"""
+
         lock = qubesadmin.utils.LockFile(LOCKFILE, True)
         lock.acquire()
         try:
             self.app.log.info("Starting restore process in a DisposableVM...")
             self.create_dispvm()
+            assert self.dispvm is not None
             self.clear_old_tags()
             self.register_backup_source()
             self.dispvm.start()
@@ -340,6 +350,7 @@ class RestoreInDisposableVM:
             return backup_log
         except subprocess.CalledProcessError as e:
             if e.returncode == 127:
+                assert self.dispvm is not None
                 raise qubesadmin.exc.QubesException(
                     '{} missing in {} template, install it there '
                     'package there'.format(self.terminal_app[0],
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -1139,6 +1139,8 @@ class BackupRestore(object):
             return hmac_text
         if algorithm is None:
             algorithm = self.header_data.hmac_algorithm
+        assert algorithm is not None
+
         passphrase = self.passphrase.encode('utf-8')
         self.log.debug("Verifying file %s", filename)
 
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -521,8 +521,8 @@ class ExtractWorker3(Process):
             return
         if terminate:
             if self.import_process is not None:
-                self.tar2_process.terminate()
                 self.import_process.terminate()
+            self.tar2_process.terminate()
         if wait:
             self.tar2_process.wait()
             if self.import_process is not None:
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -180,6 +180,7 @@ class BackupHeader(object):
 
         if header_data is not None:
             self.load(header_data)
+        self.validate()
 
     def load(self, untrusted_header_text: bytes) -> None:
         """Parse backup header file.
@@ -236,8 +237,6 @@ class BackupHeader(object):
                             f=value))
                 raise QubesException("Invalid value for header: {}".format(key))
             setattr(self, header.field, value)
-
-        self.validate()
 
     def validate(self) -> None:
         '''Validate header data, according to header version'''
--- a/qubesadmin/backup/restore.py
+++ b/qubesadmin/backup/restore.py
@@ -361,9 +361,9 @@ class ExtractWorker3(Process):
     # pylint: disable=too-many-instance-attributes
     def __init__(self, queue: Queue, base_dir: str, passphrase: str,
                  encrypted: bool, *,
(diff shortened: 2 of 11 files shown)


qubesadmin/backup/dispvm.py
```diff
@@ -27,52 +27,57 @@
...
 import qubesadmin.exc
 import qubesadmin.utils
 import qubesadmin.vm
+from qubesadmin.app import QubesBase
+from qubesadmin.vm import QubesVM
 
 LOCKFILE = '/var/run/qubes/backup-paranoid-restore.lock'
 
 Option = collections.namedtuple('Option', ('opts', 'handler'))
+type ValidArgValue = str | bool | int | list[str] | None
```
marmarek: CI says it fails on Debian bookworm:
```
02:01:59 [qb.build_deb.core-admin-client.vm-bookworm.build]   File "/build/qubes-core-admin-client-4.3.27/qubesadmin/backup/dispvm.py", line 43
02:01:59 [qb.build_deb.core-admin-client.vm-bookworm.build]     type ValidArgValue = str | bool | int | list[str] | None
02:01:59 [qb.build_deb.core-admin-client.vm-bookworm.build]          ^^^^^^^^^^^^^
02:01:59 [qb.build_deb.core-admin-client.vm-bookworm.build] SyntaxError: invalid syntax
```

The way as it's on main now should work.

marmarek:
> To make this work, I had to modify PropertyHolder accordingly:

Makes sense.

marmarek:
> Logical mistake in `cleanup_tar2`

Looks like they should be swapped - import_process under the "if", instead of tar2_process. The latter surely isn't None as that's handled at the start of the function.

marmarek:
> this will error if `self.header_data.hmac_algorithm` is None.

`self.header_data.validate()` (called from `load()`) ensures that `hmac_algorithm is not None` (see `expected_attrs`).

marmarek:
> All of the following: `header_data.crypto_algorithm, header_data.compressed, header_data.encrypted` can be `None` for the same reason as above.

That again is validated by `BackupHeader.validate()`

qubesadmin/backup/dispvm.py
hippalectryon-0: True - but this should have been caught by e.g. pylint in the previous steps. Would it be possible to run pylint either on the oldest supported version, or better, on a matrix of all the supported versions ?

hippalectryon-0:
> self.header_data.validate() (called from load()) ensures that hmac_algorithm is not None (see expected_attrs).

I don't follow - `load()` is only called in `__init__`:
```python3
        if header_data is not None:
            self.load(header_data)
```

but `algorithm` could still be `None` if `header_data` is None

marmarek:
> but `algorithm` could still be `None` if `header_data` is None

`BackupHeader()` is constructed without header_data (or `hmac_algorithm` set explicitly) only for header version 1 (which didn't really have a header at all), in which case ExtractWorker3 is not used at all. In all the other versions, header data is provided, so validate is called.

Maybe to make it easier to follow, `validate()` call could be moved to `__init__()`, to be called always, not only if `load()` was called?

hippalectryon-0:
> Maybe to make it easier to follow, validate() call could be moved to __init__(), to be called always, not only if load() was called?

Added

hippalectryon-0:
Failing QA tests very likely due to my mistake fixed in ed0f5d5.

I'm not quite sure how to read the "fixed failures" section of the report though

hippalectryon-0:
I see that QA has been reran and that the previous commit fixed the related errors. However it seems that some errors remain.
As in my previous comment, I'm afraid I'm not sure how to read the report, and for the reported errors I'm not sure I see any link with my PRs.

Could you confirm on the validity of the reported errors ?

marmarek:
Remaining error looks unrelated to your changes, it looks kinda similar to https://github.com/QubesOS/qubes-issues/issues/9964 (but I'm not sure if it's really the same).

marmarek:
Can you cleanup the commit history a bit? As in - squash commits that change the same thing back and forth (for example movement of `self.import_process.terminate()`, or TODO comments added and then remove).

qubesadmin/backup/restore.py
```diff
@@ -180,6 +180,7 @@ def __init__(self,
 
         if header_data is not None:
             self.load(header_data)
+        self.validate()
```
marmarek, line 183: Now the validate() call in load() is not needed anymore (load() isn't called anywhere else)

hippalectryon-0:
done

qubesadmin/backup/restore.py
marmarek: you removed wrong validate :) this one should stay, the one in `load()` is to be removed

hippalectryon-0: oof - my bad. fixed.

marmarek (approved):

PR #418, opened 2026-02-14
Raise NotImplementedError instead of returning NotImplemented

hippalectryon-0:
In a few instance in the code we `return NotImplemented`, but the returned value is not handled anywhere.

I believe this is an error, and should have been `raise NotImplementedError` instead.


marmarek:
Nope, the code is correct, see https://docs.python.org/3/library/constants.html#NotImplemented

PR #419, opened 2026-02-14
Upgrade deprecated datetime calls in firewall.py

hippalectryon-0:
In `firewall.py` we use two deprecated methods `utcfromtimestamp` and `utcnow`.

As per the official docs they have been replaced by `fromtimestamp` and `now`.

# Note: No UTC ?
This raises a question regarding why we are using timezone-unaware datetimes in the first place. I believe we should be using

`.now(timezone.utc)` and `.fromtimestamp(..., timezone.utc)` (and update accordingly the tests)

Otherwise, if the user's timezone changes, the `Expire` rules will be shifted around for no good reason.

### DIFF
--- a/qubesadmin/firewall.py
+++ b/qubesadmin/firewall.py
@@ -213,7 +213,8 @@ class Expire(RuleOption):
     '''Rule expire time'''
     def __init__(self, value):
         super().__init__(value)
-        self.datetime = datetime.datetime.utcfromtimestamp(int(value))
+        self.datetime = datetime.datetime.fromtimestamp(int(value),
+                                                        datetime.timezone.utc)
 
     @property
     def rule(self):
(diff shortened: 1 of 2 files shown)


marmarek:
> I believe we should be using
> 
> `.now(timezone.utc)` and `.fromtimestamp(..., timezone.utc)` (and update accordingly the tests)

Yes, exactly - it should be UTC. It was before in practice (by using now deprecated utc function variants). It should keep using UTC, using the new way of doing that in python.

hippalectryon-0:
fixed

marmarek:
Can you squash the two commits together?

Generally, I when fixing a PR, I prefer to fix existing commits (so, the state between commits in a PR is not broken). Technically, I recommend `git commit --fixup`, and later `git rebase --autosquash -i`.

hippalectryon-0:
done

marmarek (approved):

PR #423, opened 2026-02-17
enforce 3.10 interface for `qvm_notes.py`

### LINKED ISSUE #10684: [dev] Inconsistent minimum python version in `qubes-core-admin-client`
In `qubes-core-admin-client`:
- the `README` mentions a minimum version `python3.5`
- #10673 discussions suggest:
> Generally, the lowest that needs to be supported is whatever is in Debian oldstable (3.11 at this moment). And many places can require even newer (for example dom0-only parts can require 3.13 in R4.3).
- However recents commits [ee39d272](https://github.com/QubesOS/qubes-core-admin-client/commit/ee39d272ac4912653ddbe7d5f364ce8dd5dee08b) use python 3.12+ features (e.g. `delete_on_close` argument in `NamedTemporaryFile`), even though afaik `qubes-core-admin-client` does not need to run in `dom0`

Can we clarify the current minimum supported version and the rationale behind it ?

Thank you

hippalectryon-0:
Closes https://github.com/QubesOS/qubes-issues/issues/10684

The current code is using 3.12 code `delete_on_close`. Reverting to the 3.10 option `delete`.

From what I gather from the docs, this split (in python 3.12) was introduced to fix a subtle issue for windows users, hopefully it shouldn't break anything on our end

### DIFF
--- a/qubesadmin/tests/tools/qvm_notes.py
+++ b/qubesadmin/tests/tools/qvm_notes.py
@@ -85,7 +85,7 @@ class TC_00_qvm_notes(qubesadmin.tests.QubesTestCase):
         ] = b"0\x00"
         with tempfile.NamedTemporaryFile(
             mode="w+",
-            delete_on_close=False,
+            delete=False,
         ) as temp:
             temp.write("For Your Eyes Only")
             temp.close()
@@ -235,7 +235,7 @@ class TC_00_qvm_notes(qubesadmin.tests.QubesTestCase):
         )
         with tempfile.NamedTemporaryFile(
             mode="w+",
-            delete_on_close=False,
+            delete=False,
         ) as temp:
             temp.write("New note")
             temp.close()
--- a/qubesadmin/tools/qvm_notes.py
+++ b/qubesadmin/tools/qvm_notes.py
@@ -144,7 +144,7 @@ def main(args=None, app=None):
                     mode="w+",
                     prefix=qube.name + "_qube_",
                     suffix="_notes.txt",
-                    delete_on_close=False,
+                    delete=False,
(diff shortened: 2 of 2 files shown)


marmarek (changes requested):
But this will leave temp files forever... Maybe add manual unlink call later? I see it's even there, but commented out. And just to be sure, in try/finally block.

hippalectryon-0:
I'm not sure I follow - the docs of 3.10 indicate

```
'delete' -- whether the file is deleted on close (default True).
...
The file will be automatically
    deleted when it is closed unless the 'delete' argument is set to False.
```

What risk are we introducing that was not already present with the 3.12 code ?

marmarek:
`delete_on_close=False` prevents deleting on explicit `close()`, but does _not_ prevent delete on context manager exit (or rather it being destroyed).

hippalectryon-0:
I understand the difference now.

That being said, those calls are only used for the tests - does it really matter if we leave out some temp files while waiting for the transition to p3.12 ?

marmarek:
[‎qubesadmin/tools/qvm_notes.py](https://github.com/QubesOS/qubes-core-admin-client/pull/423/changes#diff-82540dd7e7a8e3a4e0d5fd1feb8830181bf4757f8b55bd3b4f8e7f27cf05b962) is not just tests

hippalectryon-0:
Fair. I added a check in the portion that's outside tests.

marmarek:
pylint (rightfully) complains

hippalectryon-0:
:( mybad

fixed

marmarek (approved):

PR #426, opened 2026-02-18
Add Dockerfile to run CI tests locally

### LINKED ISSUE #10700: [dev] Clarify local test status for `qubes-core-admin-client`
## The problem you're addressing

`qubes-core-admin-client` has a number of automated tests in `qubesqdmin/tests/`. Those tests are run by CI for automated QA.

Those tests are very useful for developers, however as it stands there is no (documented) way to run them locally. Pushing commits to PRs to force CI runs is not a great dev experience ;) (and it's a waste of CI resources).

## In more detail

I wrote above `there is no (documented) way to run them locally.` with an emphasis on `documented`.
Indeed, in theory, nothing stops the devs from running `python -m unittest discover -s qubesadmin/tests -p '*.py' -v "$@"` locally, However, the test results thus obtained vary from the CI test results.

Example on https://github.com/QubesOS/qubes-core-admin-client/commit/8ebd0961a61bdfa96feb4012c0d1378e5dcd4e9e:

#### CI result

([CI link](https://gitlab.com/QubesOS/qubes-core-admin-client/-/pipelines/2329978108))
```
Ran 789 tests in 773.381s
OK (skipped=8, expected failures=6)
```

#### Local results

```
Ran 600 tests in 2.192s
FAILED (failures=11, errors=10, skipped=4, expected failures=6)
```

using this image
```Dockerfile
FROM registry.gitlab.com/qubesos/docker-images/qubesos-ci:latest
RUN sudo dnf install -y  openssl python3-rpm sequoia-sqv python3-xlib \
    libnotify zstd xorg-x11-server-Xvfb && sudo dnf clean all
USER root
COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/uv
RUN useradd -m -u 1000 -G wheel devcontainer
RUN echo '%devcontainer ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/devcontainer-nopasswd && \
    chmod 0440 /etc/sudoers.d/devcontainer-nopasswd
WORKDIR /workspace
USER devcontainer
ENV PATH=$HOME/.local/bin:$PATH
ENV ENABLE_SLOW_TESTS=1
```

### The solution you'd like

Ideally, a Dockerfile / image / ... on which we can run the tests "out-of-the-box".

If that is not possible, documentation regarding the expected differences between the local and CI tests.

### The value to a user and who that user might be

Major QOL for devs, less strain on CIs.

### Comment

I tried to use `gitlab-ci-local` directly, unfortunately is not compatible with QubesOS's "complex" CIs.

hippalectryon-0:
Note: requires #424 (could do without, but might as well - this PR is not urgent)

Maybe closes https://github.com/QubesOS/qubes-issues/issues/10700 ?

### DIFF
--- a/Dockerfile.run_tests
+++ b/Dockerfile.run_tests
@@ -1,0 +1,34 @@
+# Builds an image to run local tests
+# Build using e.g. `podman build -t qos-test . -f Dockerfile.run_tests`
+# Execute tests with `podman run --rm -it qos-test:latest`
+
+FROM registry.gitlab.com/qubesos/docker-images/qubesos-ci:latest
+USER root
+RUN dnf install -y \
+    openssl \
+    python3-rpm \
+    sequoia-sqv \
+    python3-xlib \
+    libnotify \
+    zstd \
+    xorg-x11-server-Xvfb \
+    && dnf clean all
+RUN dnf install -y --enablerepo=qubes-host-r4.3-current-testing scrypt
+
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/uv
+
+WORKDIR /workspace
+COPY . /workspace
+RUN chown -R gitlab-runner:gitlab-runner /workspace
+
+# PAM modified to allow tests to run (e.g. via podman) on systems (=OS outside the container) which require a root pwd
+RUN echo "gitlab-runner ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers \
+    && echo "#%PAM-1.0"                       > /etc/pam.d/sudo \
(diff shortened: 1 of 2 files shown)


marmarek:
is the PAM change necessary? commit message could explain why...

hippalectryon-0:
The PAM change is necessary (from my experience at least) when running podman on a system which itself requires a password for root.
It is not necessary when running podman from QubesOS itself.

I added a comment to that effect in the Dockerfile.

marmarek (approved):

PR #427, opened 2026-02-19
Upgrade syntax to 3.10

hippalectryon-0:
This is a "long" PR diff-wise, but the content is very simple (and reproducible, if anyone wants to double-check I didn't add any weird code):

#### Step 1
- Run `find qubesadmin -name '*.py' -exec uvx -p 3.10 pyupgrade --py310-plus {} +`
 
#### Step 2
- Manually review every change.
- Make some very small adjustments (there were three at most, extremely minor - you can view them by diffing 2eb6fef04480d328fa1f9e44cc70b1c1a56e7ec7 against the output of the command above)

#### Step 3
- Run `pylint` and correct errors (a132ad6ee3ed7b27e2c98d2af32457c6b365de0e)


marmarek:
Before doing more tree-wide changes, I'd prefer remaining PRs that will need backporting to be finished first (especially https://github.com/QubesOS/qubes-core-admin-client/pull/400).

hippalectryon-0:
Makes sense !
Ping me when you think it's the right time to integrate the present PR, I'll rebase on main and re-run `pyupgrade`to make sure new changes are also 3.10-upgraded.

PR #429, opened 2026-02-20
Refine logic in SinglePropertyAction

hippalectryon-0:
### Main point
- replace the `default={}` `properties = properties.copy()` mutable hack by non-mutable `default=None`
- allow `SinglePropertyAction` to handle optional values with `const!=None`

The previous implementation either accepts `const=None` (`--my-arg value`) or `const=val` (`--my_arg`). It is not possible to add a default value if the flag has no argument, or use the provided value if there is any.

The new implementation allows that.
- Cleanup logic in `SinglePropertyAction.__call__` if we set `const`

At the top of `__call__` we had
```python3
        if values is self.default and self.default == {}:
            return
```

1) This breaks the expected argument type of `values` which is supposed to be `str | Sequence | None`
2) If `self.const` was set during `__init__`, then we still will not set the property to this value, which is very counter-intuitive.

### Minor points
- Re-add non-forwarded `required` argument to `__init__`
- Reword some docstrings in a way that imo is clearer - but that is subjective
- Add typing hints, since this is a cherry-pick from https://github.com/hippalectryon-0/qubes-core-admin-client which is fully typed
- I took the liberty to remove `VM_ALL`. It's not used anywhere, it comes from a `qubes-core-admin` file that was last modified in 2016:
```bash
> git log -S "VM_ALL"
commit d4c74d210f7f139748d311cec4b94914f3da6544
Author: Bahtiar `kalkin-` Gadimov <bahtiar@gadimov.de>
Date:   Wed May 18 17:18:54 2016 +0200

    Implement vmname parsing

commit adbca5c0f663f2bd010bead8ccc909d65d6b6543
Author: Wojtek Porczyk <woju@invisiblethingslab.com>
Date:   Mon Apr 18 22:54:55 2016 +0200

    qubes/tools/qvm_run: fix --all/--passio exclusion

commit f1a0b1af3947a71269763e37eaba7809f59376b7
Author: Wojtek Porczyk <woju@invisiblethingslab.com>
Date:   Mon Dec 28 18:14:37 2015 +0100

    qubes/tools: add qvm-run, qvm-{,un}pause
    
    Also change convention of calling main(): now command returns its
    numeric value instead of bool.
    
    Also fixed QSB#13
    
    fixes QubesOS/qubes-issues#1226
```

### DIFF
--- a/qubesadmin/tools/__init__.py
+++ b/qubesadmin/tools/__init__.py
@@ -38,9 +38,6 @@ import qubesadmin.log
 import qubesadmin.exc
 import qubesadmin.vm
 
-#: constant returned when some action should be performed on all qubes
-VM_ALL = object()
-
 class QubesAction(argparse.Action):
     """
     Custom Action for Qubes
--- a/qubesadmin/tools/__init__.py
+++ b/qubesadmin/tools/__init__.py
@@ -31,7 +31,8 @@ import logging
 import os
 import subprocess
 import sys
-import textwrap
+from argparse import Namespace
+from collections.abc import Sequence
 
 import qubesadmin.log
 import qubesadmin.exc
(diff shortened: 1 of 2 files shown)


qubesadmin/tools/__init__.py
```diff
@@ -31,94 +31,107 @@
...
+                 option_string: str | None=None)\
+            -> None:
+        if not values:
+            # If we omit the argument and it's optional, __call__ will
+            #  not be called.
+            # If we omit it and it's positional, `values` will be either
+            #  `None` or `[]` depending on `nargs`.
+            # We don't want to modify `propertie` in either case, unless
+            #  `const` is set.
+            if not self.const:
```
marmarek:
```suggestion
            if self.const is None:
```
?
to allow `const=0` for example

hippalectryon-0: good catch - I added an additional check for [], which we can get when nargs>=0.

marmarek (approved):

PR #431, opened 2026-02-20
remove unused test stubs

hippalectryon-0:
Requires #424

Removed unused `./test-packages`

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -2,10 +2,6 @@
 
 : "${PYTHON:=python3}"
 : "${ROOTDIR:=.}"
-: "${TESTPYTHONPATH:=$ROOTDIR/test-packages}"
-
-PYTHONPATH="${TESTPYTHONPATH}:${PYTHONPATH}"
-export PYTHONPATH
 
 [ -r version ] || ln -s ${ROOTDIR}/version ./
 "${PYTHON}" -m pip install -e . --no-deps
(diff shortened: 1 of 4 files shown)


marmarek:
Hm, it used to be used by pylint to check some of the external modules not directly available in the CI env. I see CI passed - does pylint now ignore external modules? I don't see `qubesdb` on the `ignored-modules` list and also I'm pretty sure it isn't installed in the CI env. What changed?
`dbus` stub indeed seems not used anymore.

hippalectryon-0:
> What changed?

I don't remember making any intentional changes to that effect - I suppose it's been that way for some time now ?

marmarek:
PipelineRetry

marmarek:
So, it looks like removing qubesdb stub causes:
```
AssertionError: Unexpected call ('runner-t2kmvhtw-project-22468674-concurrent-0', 'admin.vm.property.Get', 'default_dispvm', None)
```
The stub did:
```
class QubesDB:  
    def read(self, key):
        return b'testvm'
```
I guess an alternative would be to override `QubesTest.local_name`.

As for the pylint, I think pylint 4 notices `try/except ImportError` around `import qubesdb` and that's why it doesn't complain later on. But also, this disables looking into `qubesdb` module:
- bug with stubs: https://gitlab.com/QubesOS/qubes-core-admin-client/-/jobs/13818370918
- bug without stubs: https://gitlab.com/QubesOS/qubes-core-admin-client/-/jobs/13818393212

PR #432, opened 2026-02-20
Correct assert volume length in `tools/__init__.py`

### LINKED ISSUE #10682: [dev] comparison will raise TypeError in `qubes-core-admin-client/qubesadmin/tools/__init__.py`
In `qubes-core-admin-client/qubesadmin/tools/__init__.py`

https://github.com/QubesOS/qubes-core-admin-client/blob/3974e0b01c6e08f4f30cc0fdd4c4f9b00b49d563/qubesadmin/tools/__init__.py#L246-L268

we have

```python
volume = [v for v in pool.volumes if v.vid == vid]
assert volume > 1, 'Duplicate vids in pool %s' % pool_name
```

This will trivially throw a `TypeError: '>' not supported between instances of 'list' and 'int'`.

### Comments

- I suppose the original aim was to assert `len(volume)>1` ?
- This code goes back to https://github.com/QubesOS/qubes-core-admin-client/blob/32e2ce8454b81fd026a61b68fcb803dbf2e884c84 from 2017 :`tools: copy unmodified file from core-admin)`. This means that this code has never been executed properly since then. Could we consider removing it in the first place ?

hippalectryon-0:
Closes https://github.com/QubesOS/qubes-issues/issues/10682

### DIFF
--- a/qubesadmin/tools/__init__.py
+++ b/qubesadmin/tools/__init__.py
@@ -238,47 +238,6 @@ class RunningVmNameAction(VmNameAction):
                 parser.error_runtime("domain {!r} is not running".format(
                     vm.name))
 
-
-class VolumeAction(QubesAction):
-    ''' Action for argument parser that gets the
-        :py:class:``qubes.storage.Volume`` from a POOL_NAME:VOLUME_ID string.
-    '''
-
-    def __init__(self, help='A pool & volume id combination',
-                 required=True, **kwargs):
-        # pylint: disable=redefined-builtin
-        super().__init__(help=help, required=required,
-                                           **kwargs)
-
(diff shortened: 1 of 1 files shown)


hippalectryon-0:
CI failed with
```bash
$ docker pull registry.gitlab.com/qubesos/docker-images/${DIST-${CI_JOB_NAME#r4.3:install:}}:latest
Error response from daemon: Head "https://registry.gitlab.com/v2/qubesos/docker-images/vm-trixie/manifests/latest": received unexpected HTTP status: 502 Bad Gateway
```

Likely not my fault

marmarek:
PipelineRetryFailed

marmarek (approved):

PR #433, opened 2026-02-21
tools: fix qvm-run -u root on template imported from R4.2

### LINKED ISSUE #10590: Root access fails on debian-12-minimal templates created in r4.2 and restored in r4.3, causing qubes-dist-upgrade error
**Note:** this bug report comes from [ephile on the forum](https://forum.qubes-os.org/t/root-access-fails-on-templates-causing-qubes-dist-upgrade-error/38695/5) but I can reproduce.

## Qubes OS version:

r4.3

## Affected component(s) or functionality:
For debian-12-minimal templates restored from r4.2.x or 4.1.x:
qvm-run -u root debian-12-minimal xterm opens the terminal as user, not root,
and qubes-dist-upgrade fails with the following error message:

```
----> Upgrading debian-12-minimal...
ERROR: This script must be run with root permissions
+  1000 -ne 0 
+ echo 'ERROR: This script must be run with root permissions'
+ exit 1
ERROR: A general error occurred while upgrading debian-12-minimal (exit code 1).
```

## Steps to reproduce the behavior:
Migrate debian-12-minimal template from a r4.2.4 machine to r4.3. Once restored, run

```
sudo qubes-dist-upgrade --releasever=4.3 --template-standalone-upgrade
```

## Expected or desired behavior:
All supported templates migrated from any r4.x should maintain root terminal access and be upgradeable with qubes-dist-upgrade.

## Actual behavior:
The upgrade process fails and attempts to access the root terminal open the user terminal instead.
qvm-console-dispvm --autostart debian-12-minimal works as expected, so root access is available via this method.

## General notes:
debian-12-minimal templates installed to r4.3 with qvm-templates behave as expected.

marmarek:
Template imported from R4.2 already has qubes.VMRootShell service, but
it isn't really running as root by default. It's only R4.3+ when both
qubes.VMRootShell and qubes.VMRootExec got also force-user=root setting.
Since the check for VMRootExec is already needed anyway, use the same to
check if using qubes.VMRootShell makes sense, as both changes were
introduced at the same time in core-agent-linux:

    1c538af2 Use rpc-config for running qubes.VMRoot* services as root
    563a9cfa Add qubes.VMRootExec service

This is more flexible method than hardcoding version check.

Fixes QubesOS/qubes-issues#10590

### DIFF
--- a/qubesadmin/tools/qvm_run.py
+++ b/qubesadmin/tools/qvm_run.py
@@ -284,15 +284,19 @@ def run_command_single(args, vm):
             service = "qubes.VMShell"
             if args.gui and args.dispvm:
                 service += "+WaitForSession"
-            elif args.user == "root":
-                service = "qubes.VMRootShell"
-                args.user = None
+            # if there is no vmexec, surely qubes.VMRootShell is also too old
+            # - see comment below
             shell_cmd = " ".join(shlex.quote(arg) for arg in all_args)
     else:
         service = "qubes.VMShell"
         if args.gui and args.dispvm:
             service += "+WaitForSession"
-        elif args.user == "root":
+        elif args.user == "root" and vm.features.check_with_template(
+                "supported-rpc.qubes.VMRootExec", False
+        ):
+            # The above intentionally checks for VMRootExec, not VMRootShell.
+            # When VMRootExec was introduced, both VMRootExec and VMRootShell
+            # got also force-user=root setting necessary for the below to work.
             service = "qubes.VMRootShell"
             args.user = None
         shell_cmd = args.cmd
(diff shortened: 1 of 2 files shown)


qubesadmin/tools/qvm_run.py
parulin: Using the condition from line 287 here works.

qubesadmin/tools/qvm_run.py
```diff
@@ -284,7 +284,12 @@ def run_command_single(args, vm):
             service = "qubes.VMShell"
             if args.gui and args.dispvm:
                 service += "+WaitForSession"
-            elif args.user == "root":
+            elif args.user == "root" and vm.features.check_with_template(
```
parulin: Does not change anything.

parulin (changes requested):
Tested on 4.3 with a remaining template from 4.2.

marmarek:
Thanks! I moved the condition.

PR #435, opened 2026-02-27
Document kernel option-related features

RandyTheOtter:


doc/manpages/qvm-features.rst
```diff
@@ -181,6 +181,21 @@ avoid calling `qubes.SetMonitorLayout` in this qube.
 
 Default: send monitor layout
 
+no-default-kernelopts
+^^^^^^^^^^^^^^^^^^^^^
+
+When set, qube is started without default kernel options, which include
```
marmarek: Maybe add something like:

```suggestion
When set, qube is started without default kernel options (in addition to kernelopts property), which include 
```
to make it clear it's about options not already visible in kernelopts property.

doc/manpages/qvm-features.rst
```diff
@@ -181,6 +181,21 @@ avoid calling `qubes.SetMonitorLayout` in this qube.
...
+``root=...``. Set your own options via ``qvm-prefs kernelopts``
+
+Default: start with default kernel options
+
+no-nomodeset
+^^^^^^^^^^^^
+
+Omits ``nomodeset`` from default kernel options.
+
+Default: start with ``nomodeset`` kernel option.
```
marmarek: It's a bit more complicated:

```suggestion
Default: start with ``nomodeset`` kernel option unless GPU is attached to the qube.
```

marmarek (changes requested):

doc/manpages/qvm-features.rst
RandyTheOtter: This way it may seem like this feature disables both... (although the next sentence should clarify it to the reader)

Perhaps something like this:

```suggestion
When set, qube is started without default kernel options, which include 
``root=...``. This feature does not affect options provided by the 
``kernelopts`` property.
```

It also makes it obvious that `kernelopts` works regardless of `no-default-kernelopts`

ben-grande:
When ready for review, please squash the commits.

RandyTheOtter:
Marek isn't answering, I follow my assumption of his intent

marmarek (approved):

PR #436, opened 2026-02-28
Add type hints for base.py

hippalectryon-0:
Follows #416 

## Comments / Open questions

* In `qubesd_call` we had
```python3
 if not self.app:
            raise NotImplementedError
``` 

I was unable to find any case in which self.app could be None in the first place, so I removed it.
* right after we have
```python3
dest: QubesVM | None = getattr(self.app, "default_dispvm", None)
return self.app.qubesd_call(dest, method, arg, payload, payload_stream)
```

but `qubesd_call` expects a non-None `dest`. I had to add an `assert dest is not None`, but what prevents it from being None in the first place here ?
* Error type:
```python3
        if prop_type == 'bool':
            if value == '':
                return AttributeError
```
Shouldn't we be returning a ValueError ?

### DIFF
--- a/qubesadmin/base.py
+++ b/qubesadmin/base.py
@@ -19,10 +19,23 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 '''Base classes for managed objects'''
+from __future__ import annotations
+
+import typing
+from typing import BinaryIO, Any, TypeAlias, TypeVar, Generic
+from collections.abc import Generator
 
 import qubesadmin.exc
 
+if typing.TYPE_CHECKING:
+    from qubesadmin.vm import QubesVM
+    from qubesadmin.app import QubesBase
+
 DEFAULT = object()
+
+# We use Any because the dynamic metatada handling of the current code
+# is too complex for type checkers otherwise
+VMProperty: TypeAlias = Any  # noqa: ANN401
 
 
 class PropertyHolder:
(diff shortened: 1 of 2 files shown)


marmarek:
> Shouldn't we be returning a ValueError ?

Makes sense. Generally, this means a bug somewhere in qubesd.

marmarek:
> but `qubesd_call` expects a non-None `dest`.

Most calls are going through `PropertyHolder.qubesd_call`, and it has handling for `dest = None` (well, almost, see discussion in the other PR).

As for `QubesLocal`/`QubesRemote`, it's an error to have `None` there, adding assert is okay. @ben-grande already found one case that, if such assert would be present, would be much easier to debug: https://github.com/QubesOS/qubes-core-admin-client/pull/438#discussion_r2921917359

hippalectryon-0:
> Makes sense. Generally, this means a bug somewhere in qubesd.

After further investigation this is intentional - since this is a wrapper to access named field <prop_type>, it should indeed raise an AttributeError if the prop_type doesn't exist

marmarek:
You mean a case where property is declared but not set? Like for example `stubdom_mem` for PVH? Yeah, indeed in that case qubesd may return an empty value... The type situation here is not great, you get type=int, but empty string as serialized value. But then, if you have type=str, there is no way to express that situation...

hippalectryon-0:
Regarding the discussion about `dest` being None, after re-reviewing this PR, I don't think there's anything blocking left - the required work (if any) will be done in other PRs.

Can you confirm if there's any other open comments/... for this PR ?

marmarek:
In that case please remove the TODO comment about AttributeError

hippalectryon-0:
done

marmarek:
should be okay now, but I'd like to wait for CI to become green first.

marmarek (approved):

PR #437, opened 2026-02-28
Add type hints for device_protocol.py

hippalectryon-0:
Follows #436 

Comment:
* the function `qbool` has nothing to do with `device_protocol` intrinsically, and is never used within - it's used in `backup/core3.py` as `device_protocol.qbool`. I believe it should be moved elsewhere (either inside `core3.py` or in some `util` / `lib` / `misc` file)

### DIFF
--- a/qubesadmin/backup/core3.py
+++ b/qubesadmin/backup/core3.py
@@ -28,10 +28,8 @@ from lxml.etree import _Element
 
 import qubesadmin.backup
 import qubesadmin.firewall
-from qubesadmin import device_protocol
+from qubesadmin import utils
 from qubesadmin.vm import QubesVM
-
-
 
 
 class Core3VM(qubesadmin.backup.BackupVM):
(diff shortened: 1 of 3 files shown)


marmarek:
> * the function `qbool` has nothing to do with `device_protocol` intrinsically

It used to be used here, but not anymore (since 410215730e947f7fc0bd89a249bbdab6d7fa1b9c). I agree, should be moved.

hippalectryon-0:
Do you want me to move it in this PR ? If so what would be the appropriate destination ?

ben-grande:
> Do you want me to move it in this PR ? If so what would be the appropriate destination ?

`qubesadmin.utils`.

marmarek:
PipelineRetry

marmarek:
I'd like to merge #442 first (when it will be ready), and then I'd ask you to rebase (which would be useful anyway, since some commits from other PRs here are outdated.

marmarek (approved):

PR #438, opened 2026-02-28
Add type hints to app.py

hippalectryon-0:
Follows #437 

#### _vm_list

* `_vm_list` is no a list - it's a dict. I renamed it accordingly.
* I added a new `_vm_dict_initialized` variable that explicitly tracks the initialisation rather than relying on `is None`. (which avoids putting `assert is not None` everywhere)
* (not done in this PR) I believe we should add a comment / docstring to state clearly what's supposed to be in both `_vm_list` and `_vm_objects`. Would be happy to add it to the PR if I get some suggestions.

#### deviceclass

```python3
def list_deviceclass(self) -> list[str]: 
```

I would like to make `deviceclass` a `Literal` list of allowed strings, but I'm not sure what the exhaustive list of allowed classes are ?

#### Safe asserts `is not None`

* What guarantees below that `volume.name` is not `None` ?

```python3
default_pool = getattr(
                    self.app, "default_pool_" + volume.name, volume.pool
                )
```

* Same for `dst_volume.name` a bit after:
```python3
src_volume = src_vm.volumes[dst_volume.name]
```

* in `qubesd_call` we have
```python3
    def qubesd_call(
        self, dest, method, arg=None, payload=None, payload_stream=None
    ):
        if payload_stream:
            method_path = os.path.join(
                qubesadmin.config.QREXEC_SERVICES_DIR, method
            )
            if not os.path.exists(method_path):
                raise qubesadmin.exc.QubesDaemonCommunicationError(
                    "{} not found".format(method_path)
                )
            command = [
                "env",
                "QREXEC_REMOTE_DOMAIN=dom0",
                "QREXEC_REQUESTED_TARGET=" + dest,
                method_path,
                arg,
            ]
          self._call_with_stream(
                command, payload, payload_stream
            )
```

  * This will fail if `dest=None` but this is not mentioned in the docstring
  * This should also fail if `arg` is `None` since _call_with_stream excepts a `list[str]`, not `list[str|None]`. Is that OK ?

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -878,15 +878,15 @@ class QubesLocal(QubesBase):
                 raise qubesadmin.exc.QubesDaemonCommunicationError(
                     "{} not found".format(method_path)
                 )
-            assert arg is not None
             assert dest is not None
             command = [
                 "env",
                 "QREXEC_REMOTE_DOMAIN=dom0",
                 "QREXEC_REQUESTED_TARGET=" + dest,
                 method_path,
-                arg,
             ]
+            if arg is not None:
+                command.append(arg)
             if os.getuid() != 0:
                 command.insert(0, "sudo")
             (_, stdout, _) = self._call_with_stream(
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -32,8 +32,11 @@ import subprocess
 import sys
 
 import logging
+import typing
 from logging import Logger
-from typing import Literal
+from subprocess import Popen
+from typing import IO, TypeVar
+from collections.abc import Generator, Iterable
 
 import qubesadmin.base
 import qubesadmin.exc
@@ -41,62 +44,72 @@ import qubesadmin.label
 import qubesadmin.storage
 import qubesadmin.utils
 import qubesadmin.vm
+from qubesadmin.label import Label
+from qubesadmin.vm import Klass, PowerState
 import qubesadmin.config
 import qubesadmin.device_protocol
 from qubesadmin.vm import QubesVM
 
 try:
-    import qubesdb
+    import qubesdb  # type: ignore
 
     has_qubesdb = True
 except ImportError:
     has_qubesdb = False
 
+# ["mic", "block", "pci", "usb", "webcam"]
+# but can be extended
+DeviceClass = str
+
 
 class VMCollection:
     """Collection of VMs objects"""
 
-    def __init__(self, app):
+    def __init__(self, app: "QubesBase"):
         self.app = app
-        self._vm_list = None
-        self._vm_objects = {}
+        # TODO we should properly document what's in / the
+        #  purpose of _vm_dict and _vm_objects
+        self._vm_dict: dict[str, dict[str, str]] = {}
+        self._vm_objects: dict[str, QubesVM] = {}
+        self._vm_dict_initialized: bool = False
 
-    def clear_cache(self, invalidate_name=None):
+    def clear_cache(self, invalidate_name: str | None=None) -> None:
         """Clear cached list of VMs
         If *invalidate_name* is given, remove that object from cache
         explicitly too.
         """
-        self._vm_list = None
+        self._vm_dict_initialized = False
+        self._vm_dict = {}
         if invalidate_name:
             self._vm_objects.pop(invalidate_name, None)
 
-    def refresh_cache(self, force=False):
+    def refresh_cache(self, force: bool=False) -> None:
         """Refresh cached list of VMs"""
-        if not force and self._vm_list is not None:
+        if not force and self._vm_dict_initialized:
             return
         vm_list_data = self.app.qubesd_call("dom0", "admin.vm.List")
-        new_vm_list = {}
+        new_vm_dict = {}
         # FIXME: this will probably change
         for vm_data in vm_list_data.splitlines():
             vm_name, props = vm_data.decode("ascii").split(" ", 1)
             vm_name = str(vm_name)
             props = props.split(" ")
-            new_vm_list[vm_name] = dict(
+            new_vm_dict[vm_name] = dict(
                 [vm_prop.split("=", 1) for vm_prop in props]
             )
             # if cache not enabled, drop power state
             if not self.app.cache_enabled:
                 try:
-                    del new_vm_list[vm_name]["state"]
+                    del new_vm_dict[vm_name]["state"]
                 except KeyError:
                     pass
 
-        self._vm_list = new_vm_list
+        self._vm_dict = new_vm_dict
         for name, vm in list(self._vm_objects.items()):
-            if vm.name not in self._vm_list:
+            if vm.name not in self._vm_dict:
                 # VM no longer exists
                 del self._vm_objects[name]
-            elif vm.klass != self._vm_list[vm.name]["class"]:
+            elif vm.klass != self._vm_dict[vm.name]["class"]:
                 # VM class have changed
                 del self._vm_objects[name]
             # TODO: some generation ID, to detect VM re-creation
@@ -104,6 +117,7 @@ class VMCollection:
                 # renamed
                 self._vm_objects[vm.name] = vm
                 del self._vm_objects[name]
+        self._vm_dict_initialized = True
 
     def __getitem__(self, item: str | QubesVM) -> QubesVM:
         if isinstance(item, QubesVM):
@@ -118,21 +132,24 @@ class VMCollection:
         and checking if exists
         """
         if item not in self._vm_objects:
-            cls = qubesadmin.vm.QubesVM
             # provide class name to constructor, if already cached (which can be
             # done by 'item not in self' check above, unless blind_mode is
             # enabled
-            klass = None
-            power_state = None
-            if self._vm_list and item in self._vm_list:
-                klass = self._vm_list[item]["class"]
-                power_state = self._vm_list[item].get("state")
-            self._vm_objects[item] = cls(
+            klass: Klass | None = None
+            power_state: PowerState | None = None
+            if item in self._vm_dict:
+                klass = typing.cast(Klass | None, self._vm_dict[item]["class"])
+                power_state = typing.cast(PowerState | None,
+                                          self._vm_dict[item].get("state"))
+            self._vm_objects[item] = QubesVM(
                 self.app, item, klass=klass, power_state=power_state
             )
         return self._vm_objects[item]
 
-    def get(self, item, default=None) -> QubesVM:
+    T = TypeVar("T")
+
+    def get(self, item: str | QubesVM, default: QubesVM | T=None)\
+            -> QubesVM | T:
         """
         Get a VM object, or return *default* if it can't be found.
         """
@@ -141,30 +158,30 @@ class VMCollection:
         except KeyError:
             return default
 
-    def __contains__(self, item):
+    def __contains__(self, item: QubesVM | str) -> bool:
         if isinstance(item, qubesadmin.vm.QubesVM):
             item = item.name
         self.refresh_cache()
-        return item in self._vm_list
+        return item in self._vm_dict
 
-    def __delitem__(self, key):
+    def __delitem__(self, key: str) -> None:
         self.app.qubesd_call(key, "admin.vm.Remove")
         self.clear_cache()
 
-    def __iter__(self):
+    def __iter__(self) -> Generator[QubesVM, None, None]:
         self.refresh_cache()
-        for vm in sorted(self._vm_list):
+        for vm in sorted(self._vm_dict):
             yield self[vm]
 
-    def keys(self):
+    def keys(self) -> Iterable[str]:
         """Get list of VM names."""
         self.refresh_cache()
-        return self._vm_list.keys()
+        return self._vm_dict.keys()
 
-    def values(self):
+    def values(self) -> list[QubesVM]:
         """Get list of VM objects."""
         self.refresh_cache()
-        return [self[name] for name in self._vm_list]
+        return [self[name] for name in self._vm_dict]
 
 
 class QubesBase(qubesadmin.base.PropertyHolder):
@@ -179,11 +196,11 @@ class QubesBase(qubesadmin.base.PropertyHolder):
     #: domains (VMs) collection
     domains: VMCollection
     #: labels collection
-    labels: qubesadmin.base.WrapperObjectsCollection
+    labels: qubesadmin.base.WrapperObjectsCollection[Label]
     #: storage pools
-    pools: qubesadmin.base.WrapperObjectsCollection
+    pools: qubesadmin.base.WrapperObjectsCollection[qubesadmin.storage.Pool]
     #: type of qubesd connection: either 'socket' or 'qrexec'
-    qubesd_connection_type: Literal["socket", "qrexec"]
+    qubesd_connection_type: typing.Literal["socket", "qrexec"]
     #: logger
     log: Logger
     #: do not check for object (VM, label etc) existence before really needed
@@ -191,7 +208,7 @@ class QubesBase(qubesadmin.base.PropertyHolder):
     #: cache retrieved properties values
     cache_enabled: bool = False
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(self, "admin.property.", "dom0")
         self.domains = VMCollection(self)
         self.labels = qubesadmin.base.WrapperObjectsCollection(
@@ -201,27 +218,28 @@ class QubesBase(qubesadmin.base.PropertyHolder):
             self, "admin.pool.List", qubesadmin.storage.Pool
         )
         #: cache for available storage pool drivers and options to create them
-        self._pool_drivers = None
+        self._pool_drivers: dict[str, list[str]] | None = None
         self.log = logging.getLogger("app")
         self._local_name = None
 
-    def list_vmclass(self):
+    def list_vmclass(self) -> list[Klass]:
         """Call Qubesd in order to obtain the vm classes list"""
         vmclass = (
             self.qubesd_call("dom0", "admin.vmclass.List").decode().splitlines()
         )
         return sorted(vmclass)
 
-    def list_deviceclass(self):
+    def list_deviceclass(self) -> list[DeviceClass]:
         """Call Qubesd in order to obtain the device classes list"""
         deviceclasses = (
             self.qubesd_call("dom0", "admin.deviceclass.List")
             .decode()
             .splitlines()
         )
+
         return sorted(deviceclasses)
 
-    def _refresh_pool_drivers(self):
+    def _refresh_pool_drivers(self) -> None:
         """
         Refresh cached storage pool drivers and their parameters.
 
@@ -241,17 +259,19 @@ class QubesBase(qubesadmin.base.PropertyHolder):
             self._pool_drivers = pool_drivers
 
     @property
-    def pool_drivers(self):
+    def pool_drivers(self) -> Iterable[str]:
         """Available storage pool drivers"""
         self._refresh_pool_drivers()
+        assert self._pool_drivers is not None
         return self._pool_drivers.keys()
 
-    def pool_driver_parameters(self, driver):
+    def pool_driver_parameters(self, driver: str) -> list[str]:
         """Parameters to initialize storage pool using given driver"""
         self._refresh_pool_drivers()
+        assert self._pool_drivers is not None
         return self._pool_drivers[driver]
 
-    def add_pool(self, name, driver, **kwargs):
+    def add_pool(self, name: str, driver: str, **kwargs) -> None:
         """Add a storage pool to config
 
         :param name: name of storage pool to create
@@ -269,12 +289,12 @@ class QubesBase(qubesadmin.base.PropertyHolder):
             "dom0", "admin.pool.Add", driver, payload.encode("utf-8")
         )
 
-    def remove_pool(self, name):
+    def remove_pool(self, name: str) -> None:
         """Remove a storage pool"""
         self.qubesd_call("dom0", "admin.pool.Remove", name, None)
 
     @property
-    def local_name(self):
+    def local_name(self) -> str:
         """Get localhost name"""
         if not self._local_name:
             local_name = None
@@ -292,7 +312,7 @@ class QubesBase(qubesadmin.base.PropertyHolder):
 
         return self._local_name
 
-    def get_label(self, label):
+    def get_label(self, label: str | int) -> Label:
         """Get label as identified by index or name
 
         :throws QubesLabelNotFoundError: when label is not found
@@ -309,10 +329,10 @@ class QubesBase(qubesadmin.base.PropertyHolder):
             for i in self.labels.values():
                 if i.index == int(label):
                     return i
-        raise qubesadmin.exc.QubesLabelNotFoundError(label)
+        raise qubesadmin.exc.QubesLabelNotFoundError(str(label))
 
     @staticmethod
-    def get_vm_class(clsname):
+    def get_vm_class(clsname: str) -> str:
         """Find the class for a domain.
 
         Compatibility function, client tools use str to identify domain classes.
@@ -324,8 +344,10 @@ class QubesBase(qubesadmin.base.PropertyHolder):
         return clsname
 
     def add_new_vm(
-        self, cls, name, label, template=None, pool=None, pools=None
-    ):
+        self, cls: str | type[QubesVM], name: str, label: str,
+            template: str | QubesVM | None=None, pool: str | None=None,
+            pools: dict | None=None
+    ) -> QubesVM:
         """Create new Virtual Machine
 
         Example usage with custom storage pools:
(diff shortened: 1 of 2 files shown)


qubesadmin/tests/__init__.py
```diff
@@ -150,6 +150,11 @@ class QubesTest(qubesadmin.app.QubesBase):
     expected_calls = None
     actual_calls = None
     service_calls = None
+    # used to be None. casting to "none" for type correctness
+    # but outside of tests, this value is not valid
+    # TODO should likely update tests to avoid
+    #  this hacky choice
+    qubesd_connection_type: str = "none"
```
ben-grande: Would be better to update tests.

qubesadmin/app.py
```diff
@@ -89,12 +100,12 @@ def refresh_cache(self, force=False):
                 except KeyError:
                     pass
 
-        self._vm_list = new_vm_list
+        self._vm_dict = new_vm_list
```
ben-grande: `new_vm_list` should be changed to `new_vm_dict` here and on declaration also.

qubesadmin/app.py
```diff
@@ -200,18 +211,20 @@ def __init__(self):
...
         vmclass = (
             self.qubesd_call("dom0", "admin.vmclass.List").decode().splitlines()
         )
-        return sorted(vmclass)
+        for e in vmclass:
+            assert e in typing.get_args(Klass)
+        return typing.cast(list[Klass], sorted(vmclass))
 
-    def list_deviceclass(self):
+    def list_deviceclass(self) -> list[str]:  # TODO make Literal ?
```
ben-grande:
```
% qubesd-query -e dom0 admin.deviceclass.List dom0
0mic
block
pci
usb
webcam
```

The `0` is the exit code, ignore it.

qubesadmin/base.py
```diff
@@ -69,26 +83,27 @@ def qubesd_call(self, dest, method, arg=None, payload=None,
...
             and dest.startswith("@dispvm")
         ):
             if dest.startswith("@dispvm:"):
                 dest = dest[len("@dispvm:") :]
             else:
-                dest = getattr(self.app, "default_dispvm", None)
+                # TODO what if `dest` remains None here ?
+                #  qubesd_call expects a non-None `dest` arg
+                dest: QubesVM | None = getattr(self.app, "default_dispvm", None)
+                assert dest is not None
```
ben-grande: It can be `None`, please remove this assertion. Let `qubesd` reply with the exception so the caller knows what happened.

qubesadmin/base.py
```diff
@@ -69,26 +83,27 @@ def qubesd_call(self, dest, method, arg=None, payload=None,
...
                 if dest:
-                    dest = dest.name
+                    dest: str = dest.name
+        assert isinstance(dest, str)
```
ben-grande: https://github.com/QubesOS/qubes-core-admin-client/pull/438/changes#r2883262864

qubesadmin/base.py
```diff
@@ -342,15 +362,15 @@ def unescape(line):
             return
         for line in properties_str.splitlines():
             # decode newlines
-            line = bytes(unescape(line))
-            name, property_str = line.split(b' ', 1)
+            line_bytes = bytes(list(unescape(line)))
```
ben-grande, line 368: Why the transformation to `list`?

qubesadmin/device_protocol.py
```diff
@@ -604,22 +611,24 @@ def from_str(
...
         if backend is None:
+            assert get_domain is not None
             backend_name, identity = representation.split(sep, 1)
             if backend_name == "_":
                 backend_name = "*"
             if backend_name != "*":
                 backend = get_domain(backend_name)
         else:
             identity = representation
+        assert backend is not None
```
ben-grande: `backend` doesn't seem to be used till the end of this function, why assert it?

ben-grande (changes requested):

ben-grande:
> #### deviceclass
> 
> ```python
> def list_deviceclass(self) -> list[str]: 
> ```
> 
> I would like to make `deviceclass` a `Literal` list of allowed strings, but I'm not sure what the exhaustive list of allowed classes are ?

Replied in the review above.

> #### Safe asserts `is not None`
> 
>     * What guarantees below that `volume.name` is not `None` ?
> 
> 
> ```python
> default_pool = getattr(
>                     self.app, "default_pool_" + volume.name, volume.pool
>                 )
> ```

`qubesadmin.storage.Volume.name` --> `self._vm_name` -> `__init__` --> `if vm is not None and vm_name is None`. So, as it is looping `src_vm.volume.values()`, it is guaranteed to not be `None`.

>     * Same for `dst_volume.name` a bit after:
> 
> 
> ```python
> src_volume = src_vm.volumes[dst_volume.name]
> ```

Same response as above.

>     * in `qubesd_call` we have
> 
> 
> ```python
>     def qubesd_call(
>         self, dest, method, arg=None, payload=None, payload_stream=None
>     ):
>         if payload_stream:
>             method_path = os.path.join(
>                 qubesadmin.config.QREXEC_SERVICES_DIR, method
>             )
>             if not os.path.exists(method_path):
>                 raise qubesadmin.exc.QubesDaemonCommunicationError(
>                     "{} not found".format(method_path)
>                 )
>             command = [
>                 "env",
>                 "QREXEC_REMOTE_DOMAIN=dom0",
>                 "QREXEC_REQUESTED_TARGET=" + dest,
>                 method_path,
>                 arg,
>             ]
>           self._call_with_stream(
>                 command, payload, payload_stream
>             )
> ```
> 
>     * This will fail if `dest=None` but this is not mentioned in the docstring

Yes, I don't think `dest` can ever be `None`.

>     * This should also fail if `arg` is `None` since _call_with_stream excepts a `list[str]`, not `list[str|None]`. Is that OK ?

I don't think it should fail, but instead omit `arg` if it is `None`.

qubesadmin/tests/__init__.py
marmarek: See discussion in https://github.com/QubesOS/qubes-core-admin-client/pull/416

hippalectryon-0: marking as resolved (unless the above-mentioned discussion is not enough, feel free to reopen)

qubesadmin/base.py
hippalectryon-0: I think this comment belongs to #436 but we can have the conversation here.

Just to be clear: I also had to add an `assert is not None` in `qubesd_call` when `payload_stream` is not `None`, otherwise we raise  a `TypeError: unsupported operand type(s) for +: 'NoneType' ...'`

Can you confirm that what you're suggesting is to remove the above-mentioned first `assert dest is no None` (L97), but not the second one (L884) ?


Note: even if `payload_stream` is `None`, `dest` is then used in `call_header = "{}+{} dom0 name {}\0".format(method, arg or "", dest)`, and I would be quite surprised if a cast to `"None"` rather than to `""` is the expected behavior here, can you confirm ?

qubesadmin/base.py
hippalectryon-0: Because `unescape` returns a `Generator`, and `bytes` expects an `Iterable`. It works without the casting, but is not "proper".

qubesadmin/device_protocol.py
hippalectryon-0: Actually this is wrong - `backend` can still be `None` here. removing.

hippalectryon-0:
> Yes, I don't think dest can ever be None.

Not sure I understand if you mean that it *is not allowed* to ever be None (in which case we should raise an explicit error, or simply remove the `| None` from the signature, but your above comments seem to indicate you'd rather raise an exception) or if in practice you believe it's not ever called with a None value (which I'd argue is not true as seen in https://github.com/QubesOS/qubes-core-admin-client/pull/438#discussion_r2883262864)

> I don't think it should fail, but instead omit arg if it is None.

133e6783

qubesadmin/base.py
ben-grande: Added some logging:

In `qubesadmin` module:
```diff
diff --git a/qubesadmin/base.py b/qubesadmin/base.py
index b07af14..91253a1 100644
--- a/qubesadmin/base.py
+++ b/qubesadmin/base.py
@@ -84,6 +84,7 @@ class PropertyHolder:
                 if dest:
                     dest = dest.name
         # have the actual implementation at Qubes() instance
+        self.app.log.warning("TTT: dest=%s method=%s arg=%s", dest, method, arg)
         return self.app.qubesd_call(dest, method, arg, payload,
             payload_stream)

```

In `qubes` module:
```diff
diff --git a/qubes/api/admin.py b/qubes/api/admin.py
index 2b86b4ce..599a636a 100644
--- a/qubes/api/admin.py
+++ b/qubes/api/admin.py
@@ -1337,6 +1337,7 @@ class QubesAdminAPI(qubes.api.AbstractQubesAPI):
             appvm = self.src.default_dispvm
         else:
             appvm = self.dest
+        self.app.log.warning("TTT: src=%s dest=%s appvm=%s", self.src, self.dest, appvm)

         self.fire_event_for_permission(dispvm_template=appvm)
         if preload:
diff --git a/qubes/vm/dispvm.py b/qubes/vm/dispvm.py
index 2f8edcd4..06ec54f8 100644
--- a/qubes/vm/dispvm.py
+++ b/qubes/vm/dispvm.py
@@ -786,7 +786,7 @@ class DispVM(qubes.vm.qubesvm.QubesVM):
         if not getattr(appvm, "template_for_dispvms", False):
             raise qubes.exc.QubesException(
                 "Refusing to create disposable out of app qube which has "
-                "template_for_dispvms=False"
+                "template_for_dispvms=False out of %s" % appvm
             )
         if preload and not appvm.can_preload():
             # Using an exception clutters the log when 'used' event is
```

Unset the dom0 disposable template:
```
qvm-prefs dom0 default_dispvm ''
```

Unset the global disposable template (just to avoid confusion):
```
qubes-prefs default_dispvm ''
```

And then:
```
% qvm-run -p --dispvm -- echo hey
app: TTT: dest=None method=admin.vm.feature.CheckWithTemplate arg=vmexec
@dispvm: Refusing to create disposable out of app qube which has template_for_dispvms=False out of None
```
```
qubesd[261121]: WARNING: TTT: src=dom0 dest=dom0 appvm=None
```

So, the client passing `None`, the server translates to `dom0` somewhere. So I am unsure if you should block it or not. I guess OpenQA will tell.

However, in this particular case, I think I made a mistake, it shouldn't try to get `self.app.default_dispvm`, but instead, should be `"dom0"`, as a string, not a qube object.

We can't pass `dest=""`, it may fail depending if we are using `VMExec` or `VMShell`, the interaction is a bit weird because of the different calls it needs to make in `qubesadmin.tools.qvm_run`:

```
% qvm-run -p --dispvm -- echo hey
app: TTT: dest= method=admin.vm.feature.CheckWithTemplate arg=vmexec
@dispvm: VM name contains illegal characters
zsh: exit 255   qvm-run -p --dispvm -- echo hey

% qvm-run -p --dispvm -- 'echo hey'
app: TTT: dest=disp5555 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: dest=disp5555 method=admin.vm.feature.CheckWithTemplate arg=os
hey
app: TTT: dest=disp5555 method=admin.vm.Kill arg=None
```

We can pass `dest=None`:
```
% qvm-run -p --dispvm -- echo hey
app: TTT: dest=None method=admin.vm.feature.CheckWithTemplate arg=vmexec
app: TTT: dest=disp6178 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: dest=disp6178 method=admin.vm.feature.CheckWithTemplate arg=os
hey
app: TTT: dest=disp6178 method=admin.vm.Kill arg=None

% qvm-run -p --dispvm -- 'echo hey'
app: TTT: dest=disp4672 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: dest=disp4672 method=admin.vm.feature.CheckWithTemplate arg=os
hey
app: TTT: dest=disp4672 method=admin.vm.Kill arg=None
```

And we can pass `dest=dom0`:
```
% qvm-run -p --dispvm -- echo hey
app: TTT: dest=dom0 method=admin.vm.feature.CheckWithTemplate arg=vmexec
app: TTT: dest=disp8458 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: dest=disp8458 method=admin.vm.feature.CheckWithTemplate arg=os
hey
app: TTT: dest=disp8458 method=admin.vm.Kill arg=None

% qvm-run -p --dispvm -- 'echo hey'
app: TTT: dest=disp4995 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: dest=disp4995 method=admin.vm.feature.CheckWithTemplate arg=os
hey
app: TTT: dest=disp4995 method=admin.vm.Kill arg=None
```

---

> However, in this particular case, I think I made a mistake, it shouldn't try to get `self.app.default_dispvm`, but instead, should be `"dom0"`, as a string, not a qube object.

I will explain this a bit more in depth. The client doesn't need to know the system/global default disposable template, they just need to destine their call to dom0. It is better to use `dest="dom0"` then, because of how it is handled on the server:

https://github.com/QubesOS/qubes-core-admin/blob/53ca30fc3257ac8174ae6a6d37dce70f72e788d5/qubes/api/admin.py#L1336

Using the caller (src) default disposable template.

---


Not to the point of `call_header`, it seems that it is currently passed as `None` and works, if this is correct and if the server does some transformation, I don't know, but it works:
```
% qvm-run -p --dispvm -- echo hey
app: TTT: dest=None method=admin.vm.feature.CheckWithTemplate arg=vmexec
app: TTT: QubesLocal.qubesd_call header: admin.vm.feature.CheckWithTemplate+vmexec dom0 name None
app: TTT: QubesLocal.qubesd_call header: admin.vm.CreateDisposable+ dom0 name dom0
app: TTT: dest=disp220 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: QubesLocal.qubesd_call header: admin.vm.property.Get+qrexec_timeout dom0 name disp220
app: TTT: QubesLocal.qubesd_call header: admin.vm.Start+ dom0 name disp220
app: TTT: dest=disp220 method=admin.vm.feature.CheckWithTemplate arg=os
app: TTT: QubesLocal.qubesd_call header: admin.vm.feature.CheckWithTemplate+os dom0 name disp220
hey
app: TTT: dest=disp220 method=admin.vm.Kill arg=None
app: TTT: QubesLocal.qubesd_call header: admin.vm.Kill+ dom0 name disp220

% qvm-run -p --dispvm -- 'echo hey'
app: TTT: QubesLocal.qubesd_call header: admin.vm.CreateDisposable+ dom0 name dom0
app: TTT: dest=disp7281 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: QubesLocal.qubesd_call header: admin.vm.property.Get+qrexec_timeout dom0 name disp7281
app: TTT: QubesLocal.qubesd_call header: admin.vm.Start+ dom0 name disp7281
app: TTT: dest=disp7281 method=admin.vm.feature.CheckWithTemplate arg=os
app: TTT: QubesLocal.qubesd_call header: admin.vm.feature.CheckWithTemplate+os dom0 name disp7281
hey
app: TTT: dest=disp7281 method=admin.vm.Kill arg=None
app: TTT: QubesLocal.qubesd_call header: admin.vm.Kill+ dom0 name disp7281
```

If I use on the `call_header`, `dest or ""`, it doesn't work always, e.g. with VMExec it fails, with VMShell, it never reached that point, so it was not impacted, but if it reached, it would most likely break also:

```
% qvm-run -p --dispvm -- echo hey
app: TTT: dest=None method=admin.vm.feature.CheckWithTemplate arg=vmexec
app: TTT: QubesLocal.qubesd_call header: admin.vm.feature.CheckWithTemplate+vmexec dom0 name
@dispvm: VM name contains illegal characters
zsh: exit 255   qvm-run -p --dispvm -- echo hey

% qvm-run -p --dispvm -- 'echo hey'
app: TTT: QubesLocal.qubesd_call header: admin.vm.CreateDisposable+ dom0 name dom0
app: TTT: dest=disp4346 method=admin.vm.property.Get arg=qrexec_timeout
app: TTT: QubesLocal.qubesd_call header: admin.vm.property.Get+qrexec_timeout dom0 name disp4346
app: TTT: QubesLocal.qubesd_call header: admin.vm.Start+ dom0 name disp4346
app: TTT: dest=disp4346 method=admin.vm.feature.CheckWithTemplate arg=os
app: TTT: QubesLocal.qubesd_call header: admin.vm.feature.CheckWithTemplate+os dom0 name disp4346
hey
app: TTT: dest=disp4346 method=admin.vm.Kill arg=None
app: TTT: QubesLocal.qubesd_call header: admin.vm.Kill+ dom0 name disp4346
```

ben-grande:
> > Yes, I don't think dest can ever be None.
> 
> Not sure I understand if you mean that it _is not allowed_ to ever be None (in which case we should raise an explicit error, or simply remove the `| None` from the signature, but your above comments seem to indicate you'd rather raise an exception) or if in practice you believe it's not ever called with a None value (which I'd argue is not true as seen in [#438 (comment)](https://github.com/QubesOS/qubes-core-admin-client/pull/438#discussion_r2883262864))

It seems that `dest` can be `None`: https://github.com/QubesOS/qubes-core-admin-client/pull/438#discussion_r2889110523

qubesadmin/base.py
marmarek: tl;dr: `app.qubesd_call()` should never get `dest=None` (it's an error, should raise an exception). It's okay to have `dest=None` at `vm.qubesd_call()`, which will translate it to `vm._method_dest`.

The thing that happens above is bizarre, and "thanks" to lack of enforcement of `dest!=None`, it ends up producing weird results. Specifically:
1. Note this is about `admin.vm.feature.CheckWithTemplate` method, so it got called from `QubesVM.check_with_template()`. It will be important later.
2. `QubesLocal.qubesd_call()` will actually send call with `str(None)` as destination.
3. Qubesd will get it, and will (unsuccessfully) try to find such VM.
4. Qubesd will respond with `QubesVMNotFoundError("None")`.
5. Now, back in `QubesVM.check_with_template()` you get `QubesVMNotFoundError("None")`, which happen to inherit from `KeyError`, so gets handled as "no such feature" and default value is used.

marmarek: In this particular case, the issue is in `PropertyHolder.qubesd_call()` - it should verify if `default_dispvm` it got is not None, and refuse to continue otherwise (not sure what exception would be most appropriate, but the user should get a message about default dispvm not set).

marmarek:
> I would like to make `deviceclass` a `Literal` list of allowed strings, but I'm not sure what the exhaustive list of allowed classes are ?

Technically, the list can be dynamic. Add-on to core-admin can introduce new device classes - for example once you install `qubes-video-companion-dom0`, you'll get `webcam` class, but without that package it isn't known. There can be other classes introduced by other packages (for example https://github.com/QubesOS-contrib/qubes-core-admin-addon-bridge-device adds another one).


> * (not done in this PR) I believe we should add a comment / docstring to state clearly what's supposed to be in both `_vm_list` and `_vm_objects`. Would be happy to add it to the PR if I get some suggestions.

`_vm_list` (now `_vm_dict`) is a cache of known VMs, usually obtained by earlier iteration over the collection (either implicit or explicit). This avoids calling the `admin.vm.List` method over and over.
The `_vm_objects` dict is there to ensure you get the same object when do `app.domains["some-name"]` twice. If you get different objects for the same VM, weird things happen (including outdated cache in some of them...). Note the `_vm_objects` dict is intentionally not cleared if you clear cache, so even after force-refreshing VMs list, you still get the same objects for VMs that still exist.

hippalectryon-0:
> Technically, the list can be dynamic. 

OK

> _vm_list (now _vm_dict) is a cache of known VMs ...

Imo should be "fixed" by #445 so not blocking anymore for this PR

qubesadmin/vm/__init__.py
```diff
@@ -19,35 +19,38 @@
...
 import qubesadmin.features
 import qubesadmin.devices
 import qubesadmin.device_protocol
 import qubesadmin.firewall
-import qubesadmin.tags
+
+if typing.TYPE_CHECKING:
+    import qubesadmin.base
+
+Klass = Literal["AppVM", "AdminVM", "TemplateVM", "DispVM", "StandaloneVM"]
```
marmarek: Similar to device classes, this list can change by installing extra packages in dom0. But also, R4.3 has also `RemoteVM` already (with no need to change anything at core-admin-client side).

qubesadmin/app.py
```diff
@@ -31,71 +31,83 @@
...
-    def refresh_cache(self, force=False):
+    def refresh_cache(self, force: bool=False) -> None:
         """Refresh cached list of VMs"""
-        if not force and self._vm_list is not None:
+        if not force and self._vm_dict_initialized:
             return
         vm_list_data = self.app.qubesd_call("dom0", "admin.vm.List")
-        new_vm_list = {}
+        new_vm_dict = {}
+        self._vm_dict_initialized = True
```
marmarek: Safer to set this after actually assigning to `self._vm_dict`, no?

marmarek (changes requested):

hippalectryon-0:
Regarding the discussion around `dest=None` - thank you for your clarifications. Can you confirm whether you still request `dest=None`-related changes on this PR ?

qubesadmin/app.py
```diff
@@ -846,13 +878,16 @@ def qubesd_call(
...
+            assert arg is not None
+            assert dest is not None
             command = [
                 "env",
                 "QREXEC_REMOTE_DOMAIN=dom0",
                 "QREXEC_REQUESTED_TARGET=" + dest,
                 method_path,
-                arg,
             ]
+            if arg is not None:
```
marmarek, line 888: you assert that it's not None just above... but arg can be None I think (apparently it didn't exploded before only because nobody used arg=None together with payload_stream).

marmarek:
It can be in separate PR (adjusting type hint, adding assert, and handling the case of  default_dispvm=None).

marmarek (changes requested):

qubesadmin/vm/__init__.py
hippalectryon-0: 3f1c9802

hippalectryon-0:
I believe I resolved the comments above

qubesadmin/base.py
ben-grande: Looks ok?

```python3
class QubesEmptyPropertyValueError(QubesException):
    """Using the value of the property is expected, but it is empty"""
```
```python3

                dest = getattr(self.app.domains["dom0"], "default_dispvm", None)
                if not dest: ## W: Line too long (104/80)
                    raise qubesadmin.exc.QubesEmptyPropertyValueError(
                        "dom0 has empty 'default_dispvm'"
                    )
```

marmarek: This is quite unhelpful message... First of all, why is it hardcoding dom0 specifically here? Is it because you call qvm-run from dom0 and want to use default_dispvm of dom0? Then it shouldn't be hardcoded dom0, but `self.app.domains[self.local_name]` (or even `self.app.domains.get_blind(self.local_name)`) instead.
But then, the message should be clearer why this happened, maybe "Requested default disposable, but 'default_dispvm' property is empty"?
As for the exception, maybe just `QubesVMNotFoundError`? Or if going with a new one, sometime like `QubesDefaultNotSetError`?

ben-grande: It is hardcoding dom0 because of this: https://github.com/QubesOS/qubes-core-admin/blob/53ca30fc3257ac8174ae6a6d37dce70f72e788d5/qubes/api/admin.py#L1336

Calling `@dispvm` with `dest=dom0` gets the `dom0.default_dispvm`, which defaults to `app.default_dispvm`.

~I will use `local_name`~, I will use `dom0`, as local is the current qube that loaded the qubesadmin, which might not be dom0.

---

> But then, the message should be clearer why this happened, maybe "Requested default disposable, but 'default_dispvm' property is empty"?
> As for the exception, maybe just `QubesVMNotFoundError`? Or if going with a new one, sometime like `QubesDefaultNotSetError`?

I don't like `QubesDefaultNotSetError` the second because the `dom0.default_dispvm` can use a literal qube reference instead of a default value to use `app.deault_dispvm`.

I think I will settle for `QubesVMNotFoundError`, I just wanted to do a better exception.

ben-grande: > Calling `@dispvm` with `dest=dom0` gets the `dom0.default_dispvm`, which defaults to `app.default_dispvm`.

Correction, it gets `src.default_dispvm`, which depends on the calling qube property..., so yes, indeed `local_name` fits better.

ben-grande: This PR to qubes-builderv2 for reference:

https://github.com/QubesOS/qubes-builderv2/pull/135/changes

ben-grande: Some weird recursion error when using `self.app.domains[self.local_name]` inside `qubesd_call`...

marmarek: that should be `self.app.local_name`...

qubesadmin/app.py
```diff
@@ -118,21 +132,22 @@ def get_blind(self, item: str) -> QubesVM:
...
+                klass = typing.cast(Klass | None, self._vm_dict[item]["class"])
+                power_state = typing.cast(PowerState | None,
+                                          self._vm_dict[item].get("state"))
+            self._vm_objects[item] = QubesVM(
                 self.app, item, klass=klass, power_state=power_state
             )
         return self._vm_objects[item]
 
-    def get(self, item, default=None) -> QubesVM:
+    def get(self, item: str | QubesVM, default: QubesVM | None=None)\
```
marmarek: Technically, `default` can be anything, and then return type would be `QubesVM | that-type`. But in practice, `QubesVM | None` should be good enough.

qubesadmin/app.py
```diff
@@ -309,10 +327,11 @@ def get_label(self, label):
             for i in self.labels.values():
                 if i.index == int(label):
                     return i
+        assert isinstance(label, str)
```
marmarek: Is that really true? It looks like accessing non-existing label index would hit this too. Maybe better add `str(label)` below instead? Or modify exception to handle int too.

marmarek (changes requested):
Please squash "Cleanup dynamic types" commit into the typing one, as it simply undoes some changes.

marmarek (approved):

PR #443, opened 2026-03-06
`qvm-ls`: Remove 3.x compatibility and update manpage

hippalectryon-0:
### Major

* Remove `3.x compatibility` interfaces
* Improve / Update manpage

### Minor
* Remove `--spinner` argument (which is the default)
* Minor cleanup (unused code comments, typos)

### DIFF
--- a/qubesadmin/tools/qvm_ls.py
+++ b/qubesadmin/tools/qvm_ls.py
@@ -446,6 +446,7 @@ class Table:
     def sort_to_tree(self, domains):
         '''Sort the domains as a network tree. It returns a list of sets. Each
         tuple stores the insertion of the cell name and the vm object.
+        Note: mutates `domains` to an empty list
 
         :param list() domains: The domains which will be sorted
         :return list(tuple()) tree: returns a list of tuple(insertion, vm)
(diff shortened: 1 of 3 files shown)


qubesadmin/tools/qvm_ls.py
```diff
@@ -653,22 +653,6 @@ def get_parser():
...
-        help='Same as --format=disk')
-
-    parser_format_exclusive.add_argument('--network', '-n',
-        action='store_const', dest='format', const='network',
-        help='Same as --format=network')
-
-    parser_format_exclusive.add_argument('--kernel', '-k',
-        action='store_const', dest='format', const='kernel',
-        help='Same as --format=kernel')
-
```
marmarek, line 538: Those were created as a compatibility, but IMO they are very useful.

marmarek (changes requested):

marmarek (changes requested):
Removing `--spinner` might be fine, but I don't think removing others is a good idea. `qvm-ls --raw-list` is quite useful in scripts. And this still removes help of options that remained...

PR #444, opened 2026-03-08
qvm-template-postprocess: add installation complete message

### LINKED ISSUE #10753: Installing templates should end with some kind of "success!"-message

installing templates takes long and requires user-interaction when done. The last step in the logs says "Importing data ..."

### The solution you'd like

some form of unmistakable signal the installation is done! A quick fix would be simply to print "[template name] installed successfully!" in bold, but color-changes or other would be equally fine.

### The value to a user and who that user might be

knows installation has finished.

I *think* there is some change on the button-hue, but as I only periodically check on the progress, I don't consider this an unmistakable signal of completion.

therohityadav:
Currently, the final user-facing log during template installation is `"Importing data ..."`, which leaves users uncertain if the background post-installation tasks have finished.

Add an explicit info log at the end of the `post_install` function to provide a clear, unmistakable signal that the template installation is fully complete.

Fixes QubesOS/qubes-issues#10753

### DIFF
--- a/qubesadmin/tools/qvm_template_postprocess.py
+++ b/qubesadmin/tools/qvm_template_postprocess.py
@@ -346,6 +346,8 @@ async def post_install(args):
             subprocess.call(['sync', '-f', os.path.dirname(args.dir)])
             subprocess.call(['fstrim', os.path.dirname(args.dir)])
 
+    vm.log.info("Template '%s' installation complete", vm.name)
+
     return 0
 
 


andrewdavidwong:
> Fixes [QubesOS/qubes-issues#10753](https://github.com/QubesOS/qubes-issues/issues/10753)

I don't think this is enough to close the issue, because it's also asking for GUI changes.

marmarek:
Shouldn't this message be printed in qvm-template tool instead? It does few more things after calling the qvm-template-postprocess tool.

PR #445, opened 2026-03-10
Refactor VMCollection

hippalectryon-0:
This PR is a rewrite of `VMCollection`.
Requires #438 

Goals:
* Clearer code and docstrings
* No functional change

# Main changes:

## Improved docstring

`VMCollection` is actually a fairly counter-intuitive object. It's not a true collection (as a dict or set or Mapping) since it holds two different collections - one for the `vm.List` cache and one containing blind objects - but depending on the method called one or the other may be used.

The new docstring aims to fairly convey that.

## Better variable names

Previous variables names did not properly convey what the variables did.

## Rewrite of the `vm_objects` / `vm_dict` split

The new code uses `vms` / `known_names` instead, which 1) is clearer 2) allows for simpler logic in e.g. `get_blind()`

# Minor changes:

## Split and renamed `refresh_cache`

The old `refresh_cache` by default did *not* refresh the cache (`force=False`). It has been split into `_ensure_cache_loaded` (=`force=False`) and `refresh_cache` (=`force=True`).

## More direct vm access in `__iter__`, `values`

Since we already call `_ensure_cache_loaded` at the start of the method, we can use `get_blind` to avoid `self._vms[name] > __getitem__ > __contains __ > get_blind`

## Remove `app.cache_enabled` logic

The `VMCollection` should not have to interact with `app.cache_enabled`, this should be internal to `app`. In other words, even if we pass a `power_state=` in `__init__`, if `cache_enabled=False`, then the `app` object should not used the passed value.
Looking at `QubeBase`'s code, this is already the case: `_power_state_cache` is only not-`None` if `cache_enabled=True` - so in practice this should not change any behavior, only cleanup the code.

Note (see test `test_101_list_selected`): this "fixes" the fact that calling `qvm-ls` on a selected subset of VMs calls `admin.vm.CurrentState` even though this data is already returned by `admin.vm.List`. If I understand correctly this behavior was due to the fact that when `ArgParser` fetches a subset of VMs, it creates a new `app` object which by default has cache disabled, before the calling method can itself enable the cache.

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -63,61 +63,80 @@ DeviceClass = str
 
 
 class VMCollection:
-    """Collection of VMs objects"""
+    """A lazily-loaded, cached view of the VMs known to qubesd.
+
+    Membership (``__contains__``, ``__iter__``, ``keys()``, ``values()``)
+    reflects exclusively the VMs returned by the most recent
+    ``admin.vm.List`` call.  That list is fetched on the first membership
+    query and cached; call :py:meth:`refresh_cache` to reload.
+
+    :py:meth:`get_blind` is **not** a membership operation.
+    It returns (or creates) a :py:class:`~qubesadmin.vm.QubesVM` handle for
+    the given name without querying the API and without registering the name
+    in the membership view.  Its purpose is to provide a unique
+     :py:class:`~qubesadmin.vm.QubesVM` object for the same name on repeated
+     calls.
+    Blind objects desynced with ``vm.List`` are pruned from the internal cache
+    on the next :py:meth:`refresh_cache` call.
+
+    ``del domains[name]`` removes the VM by calling ``admin.vm.Remove``.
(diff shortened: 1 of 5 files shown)


marmarek:
This needs a rebase

hippalectryon-0:
Rebased

marmarek:
See failures at https://openqa.qubes-os.org/tests/174895/file/system_tests-tests-qubes.tests.integ.dispvm_perf.log, looks like some tests in https://github.com/QubesOS/qubes-core-admin/tree/main/tests use an API you changed. But TBH, I'm concerned with such changes, this isn't really an internal function, there may be more users of it, so we shouldn't really break it.

marmarek:
@hippalectryon-0 are you still working on this?

hippalectryon-0:
Sorry, I don't have a lot of time lately. If it's better for the project I can revert it into a draft PR until then ?

PR #448, opened 2026-03-12
Cleanup qvm_shutdown

hippalectryon-0:
*Mostly cleanup, with subtle inner API call changes that shouldn't affect users*

####  7d730eb6 harmonize behavior when has/hasnot_event

When `have_events` is `False`, after reaching timeout, the previous code calls `Kill` on the filtered `current_vms = failed_domains(current_vms)current_vms = failed_domains(current_vms)`, whereas if it is `True` it calls `Kill` on all `this_round_domains`.

The commit harmonises this behavior to always filter.
This slightly changes the API calls, as seen in the modified utests.

#### a37814a5 cleanup dead code around `args.wait`

The current code does
```python3
  for vm in this_round_domains:
      try:
          vm.shutdown(force=force)
      except:
          if not args.wait:
              vm.log.error(...)  # THIS EXITS
          else:
              remaining_domains.add(vm)
```

From this point onward, if `not args.wait`, either we exited via `error`, or `remaining_domains` is empty (and then we `return` immediately after).
This makes most following code depending on `args.wait` obsolete.

#### c08d6eba add utests

*Note: this particular commit is mostly GenAI. Tests were reviewed manually.*

### DIFF
--- a/qubesadmin/tools/qvm_shutdown.py
+++ b/qubesadmin/tools/qvm_shutdown.py
@@ -25,15 +25,10 @@
 from __future__ import print_function
 
 import sys
-import time
 
 import asyncio
 
-try:
-    import qubesadmin.events.utils
-    have_events = True
-except ImportError:
-    have_events = False
+import qubesadmin.events.utils
 import qubesadmin.tools
 import qubesadmin.exc
 
@@ -74,9 +69,8 @@ def main(args=None, app=None):  # pylint: disable=missing-docstring
 
     force = args.force or (args.all_domains and not args.exclude)
 
-    if have_events:
-        loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
     remaining_domains = args.domains
     for _ in range(len(args.domains)):
         this_round_domains = set(remaining_domains)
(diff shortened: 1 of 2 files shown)


marmarek:
The `have_events=False` case is a remnant of python2 support, where asyncio wasn't a thing. Now it can be completely dropped, it's a dead code.

hippalectryon-0:
I removed the (not_)has_event branch and rebased on main. Tests fail as often on unrelated issues due to gitlab runners, I don't have the permissions to relaunch them,

NB: there's 2-3 other files which contain `have_events=False/True` logic, should those branches also be purged ? (if so I'll make a separate PR)

marmarek:
> NB: there's 2-3 other files which contain `have_events=False/True` logic, should those branches also be purged ? (if so I'll make a separate PR)

Yes, I think so.

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -71,33 +71,28 @@ def main(args=None, app=None):  # pylint: disable=missing-docstring
...
-                    else:
-                        remaining_domains.add(vm)
+                    shutdown_failed.add(vm)
         if not args.wait:
-            if remaining_domains:
-                parser.error_runtime(
-                    'Failed to shut down: ' +
-                    ', '.join(vm.name for vm in remaining_domains),
-                    len(remaining_domains))
+            assert not shutdown_failed
```
marmarek: Why assert? It is possible that vm.shutdown() will return an exception, especially if --force is not used. Or did you meant to not remove `else:` in the exception handling above?

marmarek (changes requested):

qubesadmin/tools/qvm_shutdown.py
hippalectryon-0: Inaccuracy on my end - `error_runtime` is clearly cleaner. fixed.

marmarek (approved):

PR #449, opened 2026-03-12
Fix mocks

marmarta:
Mocks broke due to https://github.com/QubesOS/qubes-core-admin-client/commit/84edd428c3b45fde9c4cd393fdecb8615c3f221b

Now they be fixed.


marmarek:
Sorry #450 is nicer :)

PR #451, opened 2026-03-13
Update docstring to better reflect difference between qvm-kill and qvm-shutdown --force

hippalectryon-0:
Meant to address the first half of https://github.com/QubesOS/qubes-issues/issues/10726

(note: after reading https://github.com/QubesOS/qubes-issues/issues/10650#issuecomment-3888949050 it seems that the above-mentioned issue is actually not about manpages but about the qubes-docs - still I believe this is helpful)


ben-grande:
> it seems that the above-mentioned issue is actually not about manpages but about the qubes-docs - still I believe this is helpful)

Both are helpful. Also see `doc/manpages` directory.

marmarek (approved):

PR #452, opened 2026-03-13
qvm_shutdown: do not --force if --all --exclude

### LINKED ISSUE #10440: Add a `--noforce` argument to `qvm-shutdown`

I have a keybind that does `qvm-shutdown --all --exclude=...` to shut down most qubes and then locks the screen. Unfortunately, `qvm-shutdown --all` also implies `--force`. This causes issues with preloaded dispVMs (such as "orphaned" `WaitForSession` processes consuming CPU time).

### The solution you'd like

Add a `--noforce` (or some other name) argument that would make it possible to use `--all` **without** implying `--force`.

### Alternative solutions considered

Allow for wildcard/regex matches in `--exclude` to exclude DispVMs, or a way to exclude paused qubes.

hippalectryon-0:
See discussion in https://github.com/QubesOS/qubes-issues/issues/10440
Closes https://github.com/QubesOS/qubes-issues/issues/10440

As described in https://github.com/QubesOS/qubes-issues/issues/10440#issuecomment-3975224829, also requires the following PRs to be merged in the same round:
* https://github.com/QubesOS/qubes-dist-upgrade/pull/24

Note: this IS a breaking change. Scripts using `--all --exclude` and relying on `--force` WILL break.

### DIFF
--- a/doc/manpages/qvm-shutdown.rst
+++ b/doc/manpages/qvm-shutdown.rst
@@ -25,7 +25,8 @@ Options
 
 .. option:: --all
 
-   perform the action on all qubes; implies :option:`--force`
+   perform the action on all qubes; implies :option:`--force` unless
+   :option:`--exclude` is also specified
 
 .. option:: --exclude=EXCLUDE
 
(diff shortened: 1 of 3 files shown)


ben-grande:
PR contents looks fine to me.

marmarek (approved):

PR #453, opened 2026-03-13
Add qvm-set-wallpaper

### LINKED ISSUE #215: Set Dom0 wallpaper service
**Partial PR:** https://github.com/QubesOS/qubes-core-admin-client/pull/453
**Partial PoC:** https://github.com/evadogstar/qvm-wallpaper-tool/

-----

**Reported by joanna on 11 Apr 2011 10:14 UTC**
Currently there is no easy way for the user to set a custom background (wallpaper) for Dom0 desktop. Of course, one can try to copy JPEG to Dom0 using tricks with qvm-run, but that is an unsafe operation (and this is precisely why there is no tool for file copy to Dom0).

Better idea is to let the user set the wallpaper in one of the AppVMs (they have the same screen resolution), using the already existing mechanisms (e.g. right click on some JPEG and choose 'Set as wallpaper'). Then, the gui-agent should inform the guid that the current AppVM's wallpaper should also be displayed in Dom0 as root window.

This way we allow for custom wallpapers in Dom0 without introducing any security problems.

Migrated-From: https://wiki.qubes-os.org/ticket/215

hippalectryon-0:
Closes https://github.com/QubesOS/qubes-issues/issues/215

This adds `qvm-set-wallpaper`, which safely converts an image from another VM using `qubes.GetImageRGBA` and sets it as the current VM's wallpaper.

*Note: parts of this PR, e.g. utests, are GenAI generated and human-reviewed*

TODO before marking as ready:
- test on actual VM, if possible with different DE


qubesadmin/tools/qvm_set_wallpaper.py
```diff
@@ -0,0 +1,203 @@
...
+    app = app or qubesadmin.Qubes()
+    parser = get_parser()
+    args = parser.parse_args(args, app=app)
+    vm = args.domains.pop()
+    filepath = args.filepath
+    output_path = args.output
+
+    try:
+        stdout, _stderr = vm.run_service_for_stdio(
+            "qubes.GetImageRGBA", input=filepath.encode()
```
ben-grande, line 150: Assuming I understood this correctly, it is doing the conversion in the same qube the image is in. This has the problem of trusting the image. Because you downloaded the image to that qube, doesn't mean you trust to open it in that qube. That is why conversions happen in disposable qubes.

In short:

- https://github.com/QubesOS/qubes-app-linux-img-converter/blob/main/rpm_spec/qimg-converter-dom0.spec.in
- https://github.com/QubesOS/qubes-app-linux-img-converter/blob/main/qvm-get-image

ben-grande:
I think this PR should be in https://github.com/QubesOS/qubes-app-linux-img-converter

qubesadmin/tools/qvm_set_wallpaper.py
marmarek: TBH, I'm not sure if this is really a valid concern. If somebody doesn't trust the image, they can always call `qvm-convert-img` before doing anything else with the file. Which would be a good idea _in that case_ anyway, to avoid for example file manager loading it for preview.

PR #456, opened 2026-03-15
re-add qubesadmin.tags import

marmarek:
Fixes: 4a282a2 "add type hints for app.py"

### DIFF
--- a/qubesadmin/vm/__init__.py
+++ b/qubesadmin/vm/__init__.py
@@ -35,6 +35,7 @@ import qubesadmin.features
 import qubesadmin.devices
 import qubesadmin.device_protocol
 import qubesadmin.firewall
+import qubesadmin.tags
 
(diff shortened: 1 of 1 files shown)


marmarek:
@hippalectryon-0 I'm not sure why unittests/pylint didn't failed, but it exploded on openQA: https://openqa.qubes-os.org/tests/169768#step/update2/50

hippalectryon-0:
Could it be that the tests mock the `QubesVM` class ? (in which case that seems a bit scary, that's a fairly important class to test)

marmarek:
Maybe it's because `qubesadmin.tests.tags` is run earlier and that imports `qubesadmin.tags`? That's the only other place where I can find this import...

hippalectryon-0:
That's definitely a possibility. Python import order-related issues can be fairly sneaky.

PR #457, opened 2026-03-15
tools/qvm-run: translate negative return codes (signals) to positive exit codes

### LINKED ISSUE #10468: `qvm-run` returns nonzero/error upon `SIGHUP` (and `SIGKILL`) but 0/success upon `SIGINT`

### Qubes OS release

4.2, presumably 4.3

### Brief summary

If I interrupt a long `qvm-run` invocation with ctrl-c, the exit code is 0:
```
[user@dom0 ~]$ qvm-run testvm "sleep 60"
Running 'sleep 60' on testvm
# (press ctrl-c)
^C[user@dom0 ~]$ echo $?
0
```

If I signal the `qvm-run` python process with `SIGHUP` then it returns nonzero. Maybe it should do the same for `SIGINT`?

### Steps to reproduce

(See above)

### Expected behavior

Return nonzero exit code? (but see Additional Information) This would seem to make the most sense for scripting `qvm-run` calls. And glancing at [`qvm_run.py`, I think that's what's actually intended](https://github.com/QubesOS/qubes-core-admin-client/blob/a5ea121957167475554e33a3eb2830e1e4e21ad0/qubesadmin/tools/qvm_run.py#L470), but I'm not certain.

### Actual behavior

Exit code 0

### Additional information

Searching through the issue database I grasp that `qrexec` signal handling/propagation is limited for well-considered reasons.

However, it appears `qvm-run` has a carve-out in [this block](https://github.com/QubesOS/qubes-core-admin-client/blob/a5ea121957167475554e33a3eb2830e1e4e21ad0/qubesadmin/tools/qvm_run.py#L470) to propagate ctrl-c to vm subprocesses and return the max of their return codes, i.e. preferring errors over successes. An interrupted `sleep` call will exit nonzero, so I'd expect that to propagate back upstream to `qvm-run`. That looks to me to be the intent, but I'm not familiar with the API. (In C, at least, the return value of `wait()` is not an exit code)

EliasDevv:
When qvm-run is interrupted via SIGINT (ctrl-c), subprocess.wait() 
returns a negative value (-signal). This translates negative return 
codes to positive exit codes using abs(retcode) + 128, consistent 
with shell signal handling conventions.

Fixes QubesOS/qubes-issues#10468

### DIFF
--- a/qubesadmin/tools/qvm_run.py
+++ b/qubesadmin/tools/qvm_run.py
@@ -483,6 +483,9 @@ def main(args=None, app=None):
         try:
             for vm, proc in procs:
                 this_retcode = proc.wait()
+                if this_retcode < 0:
+                    this_retcode = 128 + abs(this_retcode)
+
                 if this_retcode and verbose > 0:
                     print_no_color(
                         "{}: command failed with code: {}".format(
@@ -491,7 +494,7 @@ def main(args=None, app=None):
                         file=sys.stderr,
                         color=args.color_stderr,
                     )
-                retcode = max(retcode, proc.wait())
+                retcode = max(retcode, this_retcode)
         except KeyboardInterrupt:
             for vm, proc in procs:
                 with contextlib.suppress(ProcessLookupError):


EliasDevv:
I'm aware the commits are not GPG-signed yet. I'm setting up commit signing and will force-push signed commits shortly.

ben-grande:
Please change the title of the commit message. "Translating return code" is what had to be done to accomplish the proposed goal. What this commit is about is exiting with the appropriate code.

qubesadmin/tools/qvm_run.py
```diff
@@ -491,7 +494,7 @@ def main(args=None, app=None):
                         file=sys.stderr,
                         color=args.color_stderr,
                     )
-                retcode = max(retcode, proc.wait())
+                retcode = max(retcode, this_retcode)
         except KeyboardInterrupt:
```
marmarek, line 498: For Ctrl-c, the `except` section applies, and here similar conversion is needed.

marmarek (changes requested):

marmarek:
@EliasDevv are you still working on this?

PR #458, opened 2026-03-16
Typehint label / log / spinner / storage

hippalectryon-0:
Major: type-hint `storage.py`
Minor: type-hint `label.py spinner.py log.py`

### 1. safe assert
In `__lt__` we have

```python3
            if self._vid and other._vid:
                return (self._pool, self._vid) < (other._pool, other._vid)
```
Technically, `_vid` is supposed to be non-None if `pool` is non-None, not the other way around. Shouldn't it be `if self.pool and other.pool:` ?

### Type error raise

In `revert`:
```python3
        if not isinstance(revision, str):
            raise TypeError('revision must be a str')
```

but typing `def revert(self, revision: str) -> None:` raises no type-checker error, indicating that it's never called with a non-str argument. Is the raise really necessary ?

### None `__str__`

```python3
    def __str__(self):
        return self.name
```

but `self.name` can be `None`, and `__str__` should never be `None`. I added an `assert is not None` for now.

### LT assert

```python3
    def __lt__(self, other: object) -> bool:
        if isinstance(other, Pool):
            return self.name < other.name
```
What guarantees that `self.name` or `other.name` is defined here ?

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -148,8 +148,15 @@ class VMCollection:
 
     T = TypeVar("T")
 
-    def get(self, item: str | QubesVM, default: QubesVM | T=None)\
-            -> QubesVM | T:
+    @typing.overload
+    def get(self, item: str | QubesVM) -> QubesVM:
+        ...
+    @typing.overload
+    def get(self, item: str | QubesVM, default: T) -> QubesVM | T:
+        ...
+    # Overloaded to handle default None return type
+    def get(self, item: str | QubesVM, default: object=None)\
+            -> object:
         """
         Get a VM object, or return *default* if it can't be found.
         """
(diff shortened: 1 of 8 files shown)


marmarek:
> Technically, `_vid` is supposed to be non-None if `pool` is non-None, not the other way around. Shouldn't it be `if self.pool and other.pool:` ?

Yes, I think so. 

> but typing `def revert(self, revision: str) -> None:` raises no type-checker error, indicating that it's never called with a non-str argument. Is the raise really necessary ?

It's a library, the function may be called externally. That said, type hint instead of this is probably enough.

> but `self.name` can be `None`, and `__str__` should never be `None`. I added an `assert is not None` for now.

Can `Pool.name` really be `None`? I don't think that should be possible or allowed.

> What guarantees that `self.name` or `other.name` is defined here ?

See above.

hippalectryon-0:
> Can Pool.name really be None? I don't think that should be possible or allowed.

```python3
def __init__(self, app: QubesBase, name: str | None=None):
```

As per your comment I made `name` a required argument.

marmarek (approved):

PR #459, opened 2026-03-16
Typehint tags.py utils.py

### LINKED ISSUE #10680: [dev] `get_entry_point_one` uses an obsolete API
`get_entry_point_one` is a function defined in `qubes-core-admin-client` (where it's not used anywhere) and in `qubes-core-admin-client` (where it's used).

https://github.com/search?q=org%3AQubesOS%20get_entry_point_one&type=code

It has the following implementation:
```python
def get_entry_point_one(group, name):
    epoints = tuple(importlib.metadata.entry_points(group=group, name=name))
    if not epoints:
        raise KeyError(name)
    if len(epoints) > 1:
        raise TypeError(
            "more than 1 implementation of {!r} found: {}".format(
                name,
                ", ".join(
                    "{}.{}".format(ep.module_name, ".".join(ep.attrs))
                    for ep in epoints
                ),
            )
        )
    return epoints[0].load()
```

This was valid for older versions of python, but this interface is obsolete for python 3.10+, as described in the docs of `entry_points`.

As a result, `EntryPoint` objects no longer have fields `module_name` and `attrs`, which have been replaced by `module` and `attr` (which is now a single string instead of a list).

### Why make an issue rather than a PR

It's not clear how to proceed:
- an easy fix would be to replace `module_name` and `attrs` as described above:
```python3
 "{}.{}".format(ep.module, ep.attr)
```
- however this raises the following question: should automated tests have caught this earlier ? Do we need to update utests ?
- additionally, do we really need to keep `get_entry_point_one` in `qubes-core-admin-client` which doesn't use it ?

hippalectryon-0:
Also fixes https://github.com/QubesOS/qubes-issues/issues/10680

### DIFF
--- a/qubesadmin/label.py
+++ b/qubesadmin/label.py
@@ -19,8 +19,13 @@
 # with this program; if not, see <http://www.gnu.org/licenses/>.
 
 '''VM Labels'''
+from __future__ import annotations
+from typing import TYPE_CHECKING
 
 import qubesadmin.exc
+
+if TYPE_CHECKING:
+    from qubesadmin.app import QubesBase
 
 
 class Label:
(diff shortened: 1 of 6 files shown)


marmarek:
Pylint complains:
> qubesadmin/utils.py:33:0: W0611: Unused Any imported from typing (unused-import)

marmarek (approved):

PR #463, opened 2026-04-03
qvm_device: copy assign options to suggested attach command

Jayant-kernel:
Fixes QubesOS/qubes-issues#10203

## Problem

When using `qvm-<block,usb,pci> assign` with options like `--ro` or
`-o frontend-dev=xvdl`, the suggested attach command printed after a
successful assign omits those options:

```
Assigned. To attach you can now restart domain or run:
	qvm-block attach sys-borg-server sys-usb2:sda2
```

Running the suggested command without the original options would attach
the device with different settings than intended.

## Fix

Introduce `_build_options_str()` that converts the assignment's options
dict back to CLI flags (`--ro` for `read-only=yes`, `-o key=value` for
everything else), and append it to the suggested attach command:

```
Assigned. To attach you can now restart domain or run:
	qvm-block attach --ro -o frontend-dev=xvdl sys-borg-server sys-usb2:sda2
```

## Tests

- Updated `test_032_assign_ask_and_options` to assert `--ro` appears in
  the hint output.
- Added `test_032b_assign_option_in_hint` to cover `-o key=value` options.

### DIFF
--- a/qubesadmin/tools/qvm_device.py
+++ b/qubesadmin/tools/qvm_device.py
@@ -356,11 +356,18 @@ def assign_device(args):
         _print_attach_hint(assignment, vm)
 
 
+def _build_options_str(options):
+    """Build CLI option flags string from assignment options dict."""
+    parts = [f"-o {key}={value}" for key, value in options.items()]
+    return (" " + " ".join(parts)) if parts else ""
+
+
 def _print_attach_hint(assignment, vm):
     # pylint: disable=missing-function-docstring
     attached = vm.devices[assignment.devclass].get_attached_devices()
+    options_str = _build_options_str(assignment.options)
     ports = [
-        f"\tqvm-{assignment.devclass} attach {vm} "
+        f"\tqvm-{assignment.devclass} attach{options_str} {vm} "
         f"{assignment.backend_domain}:{dev.port_id}"
         for dev in assignment.devices
         if dev not in attached and not isinstance(dev, UnknownDevice)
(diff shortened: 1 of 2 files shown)


Jayant-kernel:
@andrewdavidwong @marmarek @ben-grande 
review the changes when you get the time

qubesadmin/tools/qvm_device.py
```diff
@@ -356,11 +356,24 @@ def assign_device(args):
         _print_attach_hint(assignment, vm)
 
 
+def _build_options_str(options):
+    """Build CLI option flags string from assignment options dict."""
+    parts = []
+    if options.get("read-only", "").lower() in ("yes", "true", "1"):
+        parts.append("--ro")
```
ben-grande: I don't think this workaround is necessary. Why not simply use it below as in `-o read-only=yes`?

ben-grande:
> @andrewdavidwong @marmarek @ben-grande review the changes when you get the time

Andrew doesn't review PRs, he only checks if PR description have correct references to issues. Me and Marek do review, but... no need to tag us when you have just done the PR, we will eventually see to it.

Jayant-kernel:
okay sir i will remember that

qubesadmin/tools/qvm_device.py
Jayant-kernel: Done, removed the special case.

Jayant-kernel: Good point, simplified to just use -o read-only=yes like any other option.

ben-grande:
It should be one commit, not separate ones. I think you have this knowledge, so please remember when to squash and when to keep commits separate.

Failing tests seems unspecific to this PR.

Jayant-kernel:
Squashed into one commit.

marmarek (approved):

PR #469, opened 2026-05-06
Add option to wait for domain shutdown

### LINKED ISSUE #6132: Don't retry shutdown on qube that was already shutdown in the same `qvm-shutdown` invocation using `--all`
1. ~~`qvm-shutdown --all` emits a warning for every halted VMs, both to stderr and as notification.~~ **EDIT**: Supposedly fixed in https://github.com/QubesOS/qubes-core-admin-client/commit/3f8200ec08050d1747182d09c9cbf690c7135844
2. `qvm-shutdown --all --wait` shuts down newly started VMs in the process of iteratively halting dependency VMs.

Introduce `--running` to `qvm-shutdown`, possibly with similar semantics to that of `qvm-ls` or `qvm-check`.

When the flag is specified, `qvm-shutdown` could snapshot a list of running VMs before the first iteration and filter it further as specified by command line arguments.

Users can avoid the annoyance of starting VMs in parallel to `qvm-shutdown`, only to find the system stopping them again.

**Describe alternatives you've considered**
Changing the meaning of `--all` flag breaks backwards compatibility. Introducing `--running` instead makes it consistent with other tools.

**Additional context**
TBD

TBD

- #3032

ben-grande:
By waiting for the full shutdown, it becomes possible to receive appropriate exception, that would allow adapting the client to react differently. A timeout can be interpreted and the method adjusted to kill the domain.

For: https://github.com/QubesOS/qubes-issues/issues/10835
Fixes: https://github.com/QubesOS/qubes-issues/issues/6132
Requires: https://github.com/QubesOS/qubes-core-admin/pull/807

---

Adding wait parameter to `qubesadmin.tools.qvm_shutdown` doesn't work well when specifying multiple domains. I'm thinking of a better way to handle it. I thought of looping the domains to create tasks for each of them with run_in_executor and asyncio.gather at last, but that will require changing the `qvm_shutdown` a bit.

### DIFF
--- a/qubesadmin/tests/tools/qvm_shutdown.py
+++ b/qubesadmin/tests/tools/qvm_shutdown.py
@@ -21,7 +21,6 @@
 # pylint: disable=missing-docstring
 
 import asyncio
-import unittest.mock
 
 import qubesadmin.tests
 import qubesadmin.tests.tools
@@ -88,27 +87,12 @@ class TC_00_qvm_shutdown(qubesadmin.tests.QubesTestCase):
         loop = asyncio.new_event_loop()
         asyncio.set_event_loop(loop)
 
-        mock_events = unittest.mock.AsyncMock()
-        patch = unittest.mock.patch(
-            'qubesadmin.events.EventsDispatcher._get_events_reader',
-            mock_events)
-        patch.start()
-        self.addCleanup(patch.stop)
-        mock_events.side_effect = qubesadmin.tests.tools.MockEventsReader([
-            b'1\0\0connection-established\0\0',
-            b'1\0some-vm\0domain-shutdown\0\0',
-            ])
-
         self.app.expected_calls[
-            ('some-vm', 'admin.vm.Shutdown', None, None)] = \
+            ('some-vm', 'admin.vm.Shutdown', 'wait', None)] = \
             b'0\x00'
         self.app.expected_calls[
             ('dom0', 'admin.vm.List', None, None)] = \
             b'0\x00some-vm class=AppVM state=Running\n'
-        self.app.expected_calls[
-            ('some-vm', 'admin.vm.CurrentState', None, None)] = \
-            [b'0\x00power_state=Running'] + \
-            [b'0\x00power_state=Halted']
         qubesadmin.tools.qvm_shutdown.main(['--wait', 'some-vm'], app=self.app)
         self.assertAllCalled()
 
@@ -117,27 +101,14 @@ class TC_00_qvm_shutdown(qubesadmin.tests.QubesTestCase):
         loop = asyncio.new_event_loop()
         asyncio.set_event_loop(loop)
 
-        mock_events = unittest.mock.AsyncMock()
-        patch = unittest.mock.patch(
-            'qubesadmin.events.EventsDispatcher._get_events_reader',
-            mock_events)
-        patch.start()
-        self.addCleanup(patch.stop)
-        mock_events.side_effect = qubesadmin.tests.tools.MockEventsReader([
-            b'1\0\0connection-established\0\0',
-            b'1\0sys-net\0domain-shutdown\0\0',
-            b'1\0some-vm\0domain-shutdown\0\0',
-            b'1\0other-vm\0domain-shutdown\0\0',
-            ])
-
         self.app.expected_calls[
-            ('some-vm', 'admin.vm.Shutdown', 'force', None)] = \
+            ('some-vm', 'admin.vm.Shutdown', 'force+wait', None)] = \
             b'0\x00'
         self.app.expected_calls[
-            ('other-vm', 'admin.vm.Shutdown', 'force', None)] = \
+            ('other-vm', 'admin.vm.Shutdown', 'force+wait', None)] = \
             b'0\x00'
         self.app.expected_calls[
-            ('sys-net', 'admin.vm.Shutdown', 'force', None)] = \
+            ('sys-net', 'admin.vm.Shutdown', 'force+wait', None)] = \
             b'0\x00'
         self.app.expected_calls[
             ('dom0', 'admin.vm.List', None, None)] = \
@@ -145,15 +116,6 @@ class TC_00_qvm_shutdown(qubesadmin.tests.QubesTestCase):
             b'sys-net class=AppVM state=Running\n' \
             b'some-vm class=AppVM state=Running\n' \
             b'other-vm class=AppVM state=Running\n'
-        self.app.expected_calls[
-            ('some-vm', 'admin.vm.CurrentState', None, None)] = \
-            b'0\x00power_state=Halted'
-        self.app.expected_calls[
-            ('other-vm', 'admin.vm.CurrentState', None, None)] = \
-            b'0\x00power_state=Halted'
-        self.app.expected_calls[
-            ('sys-net', 'admin.vm.CurrentState', None, None)] = \
-            b'0\x00power_state=Halted'
         qubesadmin.tools.qvm_shutdown.main(['--wait', '--all'], app=self.app)
         self.assertAllCalled()
 
@@ -162,31 +124,20 @@ class TC_00_qvm_shutdown(qubesadmin.tests.QubesTestCase):
         loop = asyncio.new_event_loop()
         asyncio.set_event_loop(loop)
 
-        mock_events = unittest.mock.AsyncMock()
-        patch = unittest.mock.patch(
-            'qubesadmin.events.EventsDispatcher._get_events_reader',
-            mock_events)
-        patch.start()
-        self.addCleanup(patch.stop)
-        mock_events.side_effect = qubesadmin.tests.tools.MockEventsReader([
-            b'1\0\0connection-established\0\0',
-            b'1\0sys-net\0domain-shutdown\0\0',
-            ])
-
         self.app.expected_calls[
-            ('some-vm', 'admin.vm.Shutdown', 'force', None)] = \
-            b'0\x00'
+            ('some-vm', 'admin.vm.Shutdown', 'force+wait', None)] = \
+            b'2\x00QubesVMShutdownTimeoutError\x00\x00Shutdown timed out\x00'
         self.app.expected_calls[
             ('some-vm', 'admin.vm.Kill', None, None)] = \
             b'2\x00QubesVMNotStartedError\x00\x00Domain is powered off\x00'
         self.app.expected_calls[
-            ('other-vm', 'admin.vm.Shutdown', 'force', None)] = \
-            b'0\x00'
+            ('other-vm', 'admin.vm.Shutdown', 'force+wait', None)] = \
+            b'2\x00QubesVMShutdownTimeoutError\x00\x00Shutdown timed out\x00'
         self.app.expected_calls[
             ('other-vm', 'admin.vm.Kill', None, None)] = \
             b'0\x00'
         self.app.expected_calls[
-            ('sys-net', 'admin.vm.Shutdown', 'force', None)] = \
+            ('sys-net', 'admin.vm.Shutdown', 'force+wait', None)] = \
             b'0\x00'
         self.app.expected_calls[
             ('dom0', 'admin.vm.List', None, None)] = \
@@ -194,27 +145,9 @@ class TC_00_qvm_shutdown(qubesadmin.tests.QubesTestCase):
             b'sys-net class=AppVM state=Running\n' \
             b'some-vm class=AppVM state=Running\n' \
             b'other-vm class=AppVM state=Running\n'
-        self.app.expected_calls[
-            ('some-vm', 'admin.vm.CurrentState', None, None)] = [
-            b'0\x00power_state=Running',
-            b'0\x00power_state=Running',
-            b'0\x00power_state=Running',
-        ]
-        self.app.expected_calls[
-            ('other-vm', 'admin.vm.CurrentState', None, None)] = [
-            b'0\x00power_state=Running',
-            b'0\x00power_state=Running',
-            b'0\x00power_state=Running',
-        ]
-        self.app.expected_calls[
-            ('sys-net', 'admin.vm.CurrentState', None, None)] = [
-            b'0\x00power_state=Halted',
-            b'0\x00power_state=Halted',
-            b'0\x00power_state=Halted',
-        ]
         with self.assertRaisesRegex(SystemExit, '2'):
             qubesadmin.tools.qvm_shutdown.main(
-                ['--wait', '--all', '--timeout=1'], app=self.app)
+                ['--wait', '--all'], app=self.app)
         self.assertAllCalled()
 
     def test_016_all_exclude_noforce(self):
@@ -266,25 +199,10 @@ class TC_00_qvm_shutdown(qubesadmin.tests.QubesTestCase):
             ['--dry-run', 'some-vm'], app=self.app)
         self.assertAllCalled()
 
-    def test_011_wait_retry(self):
+    def test_011_wait_retry_simple(self):
         '''test --wait retries VMs whose shutdown request failed'''
         loop = asyncio.new_event_loop()
         asyncio.set_event_loop(loop)
-
-        mock_events = unittest.mock.AsyncMock()
-        patch = unittest.mock.patch(
-            'qubesadmin.events.EventsDispatcher._get_events_reader',
-            mock_events)
-        patch.start()
-        self.addCleanup(patch.stop)
-        mock_events.side_effect = qubesadmin.tests.tools.MockEventsReader([
-            # round 1: wait for some-vm
-            b'1\0\0connection-established\0\0',
-            b'1\0some-vm\0domain-shutdown\0\0',
-            # round 2: wait for other-vm
-            b'1\0\0connection-established\0\0',
-            b'1\0other-vm\0domain-shutdown\0\0',
-            ])
 
         self.app.expected_calls[
             ('dom0', 'admin.vm.List', None, None)] = \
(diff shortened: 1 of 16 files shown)


ben-grande:
https://github.com/QubesOS/qubes-core-admin-client/pull/469/commits/6ebacfb97d9f7c7fa30905624dc1cc6adc11ec2a

It works, half of it, at least.

This will make some tests fails because it wasn't updated to have `wait` in the argument. But I am not sure if I will continue with the current method, since it currently blocks exiting the interpreter until the api call ceases (unblocks the thread). It is not "dangerous", but an inconvenience and might break expectations. For more information, please read the big "TODO" on top of "ThreadPoolexecutor".

The `future.cancel()` is `False`, as well as `future.cancelled()`. The `future.running()` is `True`. All this means is that it cannot be cancelled.

ben-grande:
I tried simper things and had the same problem.

```python3
    loop = asyncio.get_event_loop()
    with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
        for qube in domains:
            future = loop.run_in_executor(
               executor, partial(qube.shutdown, force=force, wait=args.wait)
            )
            tasks.append(future)
        loop.run_until_complete(asyncio.wait_for(asyncio.gather(*tasks), args.timeout))
```

```python3
async def main(...):

    ...

    loop = asyncio.get_running_loop()
    tasks = []
    with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
        for qube in domains:
            future = loop.run_in_executor(executor, qube.shutdown)
            tasks.append(future)
        async with asyncio.timeout(args.timeout):
            await asyncio.gather(*tasks)
    return
```
And change to `asyncio.run(main())` on `/usr/bin/qvm-shutdown`.


```python3
  File "/usr/lib/python3.13/site-packages/qubesadmin/tools/qvm_shutdown.py", line 111, in main
    await asyncio.gather(*tasks)
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

  File "/usr/lib/python3.13/site-packages/qubesadmin/tools/qvm_shutdown.py", line 110, in main
    async with asyncio.timeout(args.timeout):
               ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/timeouts.py", line 116, in __aexit__
    raise TimeoutError from exc_val
TimeoutError

During handling of the above exception, another exception occurred:

  File "/usr/bin/qvm-shutdown", line 7, in <module>
    sys.exit(asyncio.run(main()))
             ~~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/usr/lib64/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 712, in run_until_complete
    self.run_forever()
    ~~~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 683, in run_forever
    self._run_once()
    ~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.13/asyncio/base_events.py", line 2050, in _run_once
    handle._run()
    ~~~~~~~~~~~^^
  File "/usr/lib64/python3.13/asyncio/events.py", line 89, in _run
    self._context.run(self._callback, *self._args)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qubesadmin/tools/qvm_shutdown.py", line 102, in main
    with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/concurrent/futures/_base.py", line 647, in __exit__
    self.shutdown(wait=True)
    ~~~~~~~~~~~~~^^^^^^^^^^^
  File "/usr/lib64/python3.13/concurrent/futures/thread.py", line 239, in shutdown
    t.join()
    ~~~~~~^^
  File "/usr/lib64/python3.13/threading.py", line 1094, in join
    self._handle.join(timeout)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/lib64/python3.13/asyncio/runners.py", line 157, in _on_sigint
    raise KeyboardInterrupt()
KeyboardInterrupt

^CException ignored on threading shutdown:
  File "/usr/lib64/python3.13/threading.py", line 1536, in _shutdown
    atexit_call()
  File "/usr/lib64/python3.13/threading.py", line 1507, in <lambda>
    _threading_atexits.append(lambda: func(*arg, **kwargs))
  File "/usr/lib64/python3.13/concurrent/futures/thread.py", line 31, in _python_exit
    t.join()
  File "/usr/lib64/python3.13/threading.py", line 1094, in join
    self._handle.join(timeout)
KeyboardInterrupt:
```

ben-grande:
- https://discuss.python.org/t/graceful-exit-from-threadpoolexecutor-when-blocked-on-io-problem-and-possible-enhancement/80380

marmarek:
Maybe better to really do https://github.com/QubesOS/qubes-issues/issues/4719 first?

ben-grande:
> Maybe better to really do [QubesOS/qubes-issues#4719](https://github.com/QubesOS/qubes-issues/issues/4719) first?

Yes... I read some alternatives, but they are dirty: https://superfastpython.com/threadpoolexecutor-kill-running-tasks/

ben-grande:
> But I am not sure if I will continue with the current method, since it currently blocks exiting the interpreter until the api call ceases (unblocks the thread).

Technically, this is also the current behavior. With the current code (without this PR), `qvm-shutdown --timeout=3 HANGING_QUBE` will not cease the connection.

ben-grande:
The last commit, https://github.com/QubesOS/qubes-core-admin-client/pull/469/commits/b544cb0fc4f11c0f336193e97e3896efbe14bc1b, probably requires adapting the file created in `qvm-shutdown` to call `asyncio.run(main())`.

I will try again to not require the `main()` to be an async function, and pass the loop object... another day.

marmarek:
Besides the above, pylint complains, mostly about misplaced docstrings in tests.

ben-grande:
Sorry for the delay. I thought that rebooting a qube with libvirt was going to be easy.

qubesadmin/utils.py
```diff
@@ -229,3 +258,18 @@ def qbool(value: str | int | bool) -> bool:
...
+
+    :param func: synchronous function
+    :param args: args to pass to function
+    :param kwargs: kwargs to pass to function
+    :return: function result
+
+    >>> from qubesadmin.utils import async_thread
+    >>> await async_thread(vm.shutdown, force=True, wait=True)
+    """
+    return await asyncio.to_thread(functools.partial(func, *args, **kwargs))
```
marmarek: Does this really need a helper function for a single line?

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -20,116 +20,175 @@
...
+    remnants, timedout = [], []
+    tasks = [
+        asyncio.wait_for(
+            async_thread(qube.shutdown, force=args.force, wait=args.wait),
+            timeout=args.timeout,
+        )
+        for qube in domains
+    ]
+    results = await asyncio.gather(*tasks, return_exceptions=True)
+    for qube, res in zip(domains, results):
```
marmarek: This `zip()` relies on the same iteration order in domains as in the above call. Type hint says it's list (which should be fine), but actually you give it as set, which doesn't guarantee the same order (even if it works this way most of the time).

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -20,116 +20,175 @@
...
-                or (vm.klass == 'DispVM' and vm.get_power_state() == 'NA'))]
 
-def main(args=None, app=None):  # pylint: disable=missing-docstring
-    args = parser.parse_args(args, app=app)
+async def main_async(args=None, app=None):
+    # pylint: disable=missing-docstring
+    args, domains = parse_common(args=args, app=app)
+    if args.dry_run:
+        return
+    await run_async(args=args, domains=domains)
```
marmarek: Is this function used anywhere? Looks to be mostly in tests, and in one place in `qvm_template_postprocess.py` (at which I'll comment separately).

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -20,116 +20,175 @@
...
-                    'Failed to shut down: ' +
-                    ', '.join(vm.name for vm in shutdown_failed),
-                    len(shutdown_failed))
-            return
-        awaiting = remaining_domains - shutdown_failed
-        remaining_domains = shutdown_failed
-        if not awaiting:
-            # no VM shutdown request succeeded, no sense to try again
-            break
+    loop.run_until_complete(run_async(args=args, domains=domains))
```
marmarek: `asyncio.run(...)` instead?

qubesadmin/tools/qvm_template_postprocess.py
```diff
@@ -244,30 +238,11 @@ async def call_postinstall_service(vm):
...
-            if not vm.is_halted():
-                try:
-                    vm.kill()
-                except qubesadmin.exc.QubesVMNotStartedError:
-                    pass
+        timeout = qubesadmin.config.defaults['shutdown_timeout']
+        await qubesadmin.tools.qvm_shutdown.main_async(
+            args=["--timeout={}".format(timeout), "--", vm.name],
+            app=vm.app,
+        )
```
marmarek: This looks like overkill, you want to shutdown a single VM here. Just use vm.shutdown(wait=True), possibly via asyncio.

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -20,116 +20,175 @@
...
+    """
+    # pylint: disable=missing-docstring
+    remnants, timedout = [], []
+    tasks = [
+        asyncio.wait_for(
+            async_thread(qube.shutdown, force=args.force, wait=args.wait),
+            timeout=args.timeout,
+        )
+        for qube in domains
+    ]
```
marmarek: This approach significantly differs from what qvm-shutdown did before. Previously, shutdown was done in a retry loop, where shutdown was retried after previous round domains completed. This way, if any failed because of dependencies, it would retry shutdown when others are off.
This new version will fail at that. For example, assuming only `sys-net`, `sys-firewall` and `sys-whonix` are running, `qvm-shutdown --wait sys-net sys-firewall sys-whonix` (without --force) should:
1. First try to shutdown all, at which point sys-net and sys-firewall will refuse to shutdown because it's used by others. But sys-whonix shutdown will get initiated.
2. Wait for successfully initiated shutdown to complete.
3. Try again, this time sys-firewall will shutdown cleanly, but sys-net will still fail.
4. Wait for successfully initiated shutdown to complete.
5. Finally, sys-net shutdown will succeed cleanly.

qubesadmin/tests/tools/qvm_shutdown.py
```diff
@@ -266,50 +216,40 @@ def test_006_dry_run(self):
...
+            ('some-vm', 'admin.vm.Shutdown', 'wait', None)] = \
             b'0\x00'
         # other-vm fails first attempt, succeeds on retry
         self.app.expected_calls[
-            ('other-vm', 'admin.vm.Shutdown', None, None)] = [
-            b'2\x00QubesException\x00\x00Shutdown refused\x00',
+            ('other-vm', 'admin.vm.Shutdown', 'wait', None)] = [
+            b'2\x00QubesVMShutdownTimeoutError\x00\x00Shutdown timed out\x00',
             b'0\x00',
         ]
```
marmarek, line 308: This was the test for the behavior of retry-loop I described in the other comment. Your new version do that only for timeout (and changed test now does only timeout...), and only once.

qubesadmin/tools/qvm_kill.py
```diff
@@ -21,36 +21,44 @@
...
-                domain.name, e))
-
-    return exit_code
+
+def main(args=None, app=None):
+    # pylint: disable=missing-docstring
+    args = parser.parse_args(args, app=app)
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+    return loop.run_until_complete(run_async(domains=args.domains))
```
marmarek: `asyncio.run()` ?

marmarek (changes requested):

qubesadmin/utils.py
ben-grande: Not anymore, it was different then... Thanks for pointing it out.

```python3
return await asyncio.get_running_loop().run_in_executor(
    None, functools.partial(func, *args, **kwargs)
)
```

But checking `asyncio.to_thread` docs again, I don't need the `functools.partial` that was needed for `run_in_executor`, because the former accepts kwargs while the latter doesn't.

qubesadmin/tools/qvm_shutdown.py
ben-grande: It seems that conversion to set is unecessary. `app.domains` only show unique items. But, it doesn't keep order anyway, because `qubesadmin.tools.QubesArgumentParser.parse_qubes_app` makes a `set()` conversion... will fix that.

qubesadmin/tools/qvm_template_postprocess.py
ben-grande: But the old code is not just `vm.shutdown(wait=True)`, it replicates qvm_shutdown, it waits for the domain-shutdown event and kills if the qube was not shutdown.

qubesadmin/tools/qvm_shutdown.py
ben-grande: Also, I plan to use `qvm_shutdown.main_async` in other repositoryes that handled each vm.shutdown by their own way.

qubesadmin/tools/qvm_template_postprocess.py
marmarek: Then do try/except with kill here? I still think calling the whole tool (via cmdline entry point) is an overkill.

qubesadmin/tools/qvm_shutdown.py
marmarek: > in other repositoryes

Don't, the qubesadmin.tools is supposed to be just CLI wrappers around relevant functions. If you want to reuse some of the logic here elsewhere, move it to proper functions in other place (utils, _if no other idea_) use those.

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -20,116 +20,198 @@
...
+                len(used)
+            )
+        )
+    for _ in range(len(used)):
+        parser.print_error(
+            "Retrying shutdown of qubes that were in use: {}".format(
+                ", ".join(qube.name for qube in used)
+            )
+        )
+        unhandled_retry, used, timedout_retry = await shutdown(
```
marmarek: Two remarks:
- exit the loop if there is nothing to shutdown anymore
- abort if no qube succeeded - if there is no progress, there is no point in retrying

qubesadmin/tools/qvm_template_postprocess.py
```diff
@@ -244,30 +238,13 @@ async def call_postinstall_service(vm):
...
-                await asyncio.sleep(1)
-                timeout -= 1
-            if not vm.is_halted():
-                try:
-                    vm.kill()
-                except qubesadmin.exc.QubesVMNotStartedError:
-                    pass
+        try:
+            await asyncio.wait_for(
+                asyncio.to_thread(vm.shutdown, wait=True),
```
marmarek, line 243: wait=True already enforces the timeout, no need to wrap it in wait_for. And then adjust handled exception below.

ben-grande: Kinda... this is a bit confusing...

Timeout is known to the client only and never passed in the API call, which means that for the freezed qube that will timeout only after 60s (either by `shutdown_timeout` property or libvirt timeout, depending when the freeze happens), and will raise QubesVMShutdownTimeoutError, and the call is ceased.

But when setting the timeout on the client lower than 60s, and the qube is freezed, this timeout will raise the exception asyncio.TimeoutError, but still, the server has not ended the connection as it doesn't know of the client shutdown timeout, and the client can't end the connection because python can't cancel process running in a different thread, the error message is printed, but on SystemExit, it hangs until the connection is terminated by the server.

This could be solved either by

- having asynchronous caller that can cancel requests (still, the server would continue waiting for the qube to shutdown independently of client disconnect)
- sending the timeout in the api call as an argument, so the server ends the connection

marmarek: No need for any of that. Existence of this client-side timeout is only because `shutdown_timeout` property didn't existed before. Let the server side enforce the timeout, that's enough.

qubesadmin/tools/qvm_shutdown.py
ben-grande: Added tests for these, it wasn't tested...

ben-grande:
What do you think of https://github.com/QubesOS/qubes-core-admin-client/pull/469/commits/998b8df5c80dfbd3d1a0400fba65e19641dec76b?

See usage:

- https://github.com/QubesOS/qubes-core-admin-linux/pull/212
- https://github.com/QubesOS/qubes-desktop-linux-manager/pull/312/changes/561c22c0adb955bbdd7769af19817e8d6ad8db46

So when the admin client becomes async, things that I modified to use `asyncio.to_thread`, will be changed only in one place, just `qubesadmin.utils`. But I understand it might be hacky using kwargs.

One thing I noticed is that some clients wants to know the shutdown reason to set as a message, sometimes not a logger per se, as that I got it solved, but something related to the gui, such as `self.err` for error messages, see for example my workaround at: https://github.com/QubesOS/qubes-desktop-linux-manager/pull/312/changes/561c22c0adb955bbdd7769af19817e8d6ad8db46. Do you think that I should, instead of returning just failed qubes, return the exception together in a dict? `{qube: Exc("Failed cause this"), qube: OtherExc("Failed cause that")}`

Or do you prefer to handle on a per client basis, and not using qubesadmin.utils the way I did?

qubesadmin/tools/qvm_shutdown.py
```diff
@@ -20,116 +20,151 @@
...
+)
 
 parser.add_argument(
-    '--dry-run',
-    action='store_true', dest='dry_run', default=False,
-    help='don\'t really shutdown or kill the domains; useful with --wait')
+    "--timeout",
+    action="store",
+    default=None,
+    deprecated=True,
```
marmarek: Looks to be too new for Python 3.11 (Debian bookworm):
```
 TypeError: _StoreAction.__init__() got an unexpected keyword argument 'deprecated' 
```

marmarek:
> What do you think of [998b8df](https://github.com/QubesOS/qubes-core-admin-client/commit/998b8df5c80dfbd3d1a0400fba65e19641dec76b)?

Generally looks like a good idea.



> Do you think that I should, instead of returning just failed qubes, return the exception together in a dict? `{qube: Exc("Failed cause this"), qube: OtherExc("Failed cause that")}`

Yes, I think this would be more flexible. And then maybe return one dict, and let the caller decide based on exception type? You can still check for specific types in a compact way using `any()`, for example `if any(isinstance(e, QubesVMInUseError) for e in failed.values())`

marmarek:
pylint complains...

marmarek:
Is [cab1a91](https://github.com/QubesOS/qubes-core-admin-client/pull/469/commits/cab1a91c50a58dfb3e3a842e53aae0820e838724) intentionally in this PR? Looks unrelated to shutdown...

ben-grande:
> Is [cab1a91](https://github.com/QubesOS/qubes-core-admin-client/pull/469/commits/cab1a91c50a58dfb3e3a842e53aae0820e838724) intentionally in this PR? Looks unrelated to shutdown...

Unrelated, yes, made a commit first on the wrong branch, cherry picked to the correct one: https://github.com/QubesOS/qubes-core-admin-client/pull/471, and forgot to remove from this one.

ben-grande:
This is causing a RuntimeError for the event loop on Qube Manager, `qvm_start.main()` is being used there:

https://github.com/search?q=org%3AQubesOS+%22from+qubesadmin.tools%22+NOT+repo%3AQubesOS%2Fqubes-core-admin-client&type=code

I will see about moving part of the code to utils and modify qube manager to not use main(), or adapting the main() to support those requests.

ben-grande:
Marek, `qubesmanager` is using `qvm_start.main`. I can make it use `qvm_start.run_async`, or do you prefer that I make `qvm_start` with drive assignment reusable?

marmarek:
> Marek, `qubesmanager` is using `qvm_start.main`. I can make it use `qvm_start.run_async`, or do you prefer that I make `qvm_start` with drive assignment reusable?

That's yet another reason why calling other's tool main function is a bad idea... Yes, better make the drive assignment reusable.

ben-grande:
CI error: https://gitlab.com/QubesOS/qubes-core-admin-client/-/jobs/14581983314

> $ sudo dnf install -y python3-rpm python3-xlib
sudo: PAM account management error: Authentication service cannot retrieve authentication info
sudo: a password is required

ben-grande:
PipelineRetryFailed

marmarek:
Is one file per function (in utils) for some specific reason? Looks pointless...

marmarek:
And regardless of this case (where I'd prefer dfefd02b531f402759f2267f90662cb3ffaed47f be dropped), please be a bit more verbose in commit messages - it should tell not only "what" is changed, but also - more importantly - "why". That will help not only with review, but also you (or other developer) when looking at the change in half a year or so. https://www.datacamp.com/tutorial/git-commit-message has some good points, for example.

ben-grande:
> Is one file per function (in utils) for some specific reason? Looks pointless...

For most of the actions, it's just one function, but for `start`, it's multiple functions that had to be moved to utils. It became not so clear that they were all for the start action. If reading this you still think is pointless, I will revert that commit, else, I will improve the commit message.

ben-grande:
Improved the bad commit messages.

marmarek:
> For most of the actions, it's just one function, but for `start`, it's multiple functions that had to be moved to utils. It became not so clear that they were all for the start action. If reading this you still think is pointless, I will revert that commit, else, I will improve the commit message.

A comment/docstring would suffice...

marmarek (approved):

PR #470, opened 2026-05-23
add `qvm-restart`

### LINKED ISSUE #4747: [feature request] qvm-restart

RandyTheOtter:
An attempt to finalize Ali's work: https://github.com/QubesOS/qubes-core-admin-client/pull/387

Closes https://github.com/QubesOS/qubes-issues/issues/4747

I have decided to not handle shutdown in situ because I would like to avoid duplication


RandyTheOtter:
That should address the issues that are obvious-ish for me to fix.

Only thing left is [the qvm-start issue](https://github.com/QubesOS/qubes-core-admin-client/pull/387#discussion_r2494343837). @marmarek it seems to me `qvm-start` can raise the error mid-loop only if it is executed with `--drive` option (which we don't pass here), or domain is already running. It also errors out if some domains don't exist, but that shoudn't be a concern for `qvm-restart` as `qvm-shutdown` fails the same way.

Am I missing something?

marmarek:
> it seems to me `qvm-start` can raise the error mid-loop only

There is a bunch of cases where qvm-start can fail. Startup timeout, not enough memory, not enough disk space, and many more.

BTW https://github.com/QubesOS/qubes-core-admin-client/pull/469 does refactor to make at least shutdown function reusable. You might want to wait for it to be merged, and then use the new function.

RandyTheOtter:
> There is a bunch of cases where qvm-start can fail. Startup timeout, not enough memory, not enough disk space, and many more.

Shouldn't it get caught here?:
https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/tools/qvm_start.py#L201
```
        except (IOError, OSError, qubesadmin.exc.QubesException,
                ValueError) as e:
            if drive_assignment:
                try:
                    domain.devices['block'].detach(drive_assignment)
                except qubesadmin.exc.QubesException:
                    pass
            exit_code = 1
            parser.print_error(str(e))

    return exit_code
```

It returns status after the loop ends, so all start-able domains should start. They definitely do when I try a group with some of them having far too much memory to start.

marmarek:
Right, this case might be okay. But still, invoking other tool's main from another tool, parsing arguments again etc is not a good idea. This will also create several instances of `Qubes()` objects, which will cause issues with caching if that become relevant at some point...

RandyTheOtter:
> Right, this case might be okay. But still, invoking other tool's main from another tool, parsing arguments again etc is not a good idea.

What's wrong with that? If anything using the same `qvm-start` every time a qube is started feels more reliable

> This will also create several instances of `Qubes()` objects, which will cause issues with caching if that become relevant at some point...

Okay, should I create a separate pr for that, similar to Ben's `qvm-shutdown` refactor?

ben-grande:
> > Right, this case might be okay. But still, invoking other tool's main from another tool, parsing arguments again etc is not a good idea.
> 
> What's wrong with that? If anything using the same `qvm-start` every time a qube is started feels more reliable

- `main()` is intended for script invocation from the shell, sometimes, it may raise `SystemExit`, sometimes indirectly via `parser.error_runtime()`. Even if not the case now for some tools, don't rely on it.
- Most of that code is not necessary for a restart, only for a "start with drive for installation". You will end up just needing `domain.start()`

> > This will also create several instances of `Qubes()` objects, which will cause issues with caching if that become relevant at some point...
> 
> Okay, should I create a separate pr for that, similar to Ben's `qvm-shutdown` refactor?

No need for a separate PR, unless Github auto-closes this one when you rebase to a different upstream branch, and doesn't allow anyone to reopen it...

- Add my fork as a remote
- Rebase your PR on mine
- Use `qubesadmin.utils.(shutdown|start)`
- When my PR is merged, rebase your branch on the QubesOS remote back to the main branch

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,140 @@
...
+        qubesadmin.utils.start(
+            domains=[vm for vm in target_domains if vm not in failed['shutdown']]
+        )
+    )
+    if failed['start']:
+        sys.stderr.write("Failed to start domains back up:\n")
+        report_failure(failed['start'])
+
+    if args.strict and (failed['start'] or failed['shutdown']):
+        return 75
```
ben-grande, line 138: Returning non-zero on failure should be the default, `args.strict` is not necessary then.

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,140 @@
...
+        report_failure(failed['shutdown'])
+
+
+    failed['start'] = asyncio.run(
+        qubesadmin.utils.start(
+            domains=[vm for vm in target_domains if vm not in failed['shutdown']]
+        )
+    )
+    if failed['start']:
+        sys.stderr.write("Failed to start domains back up:\n")
```
ben-grande: See `parser.print_error`.

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,140 @@
...
+
+def main(args=None, app=None):
+    parser = get_parser()
+    args = parser.parse_args(args, app=app)
+    args.force = args.force or (args.all_domains and not args.exclude)
+
+    if args.all_domains:
+        target_domains = [
+            vm for vm in args.domains
+            if vm.name != 'dom0' # NOTE: allowing restart of non-dom0 AdminVMs
```
ben-grande: I don't think `dom0` is included in `args.domains` even when using `all_domains`. Can you check that?

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,140 @@
...
+"""Restart a qube / qubes"""
+
+import asyncio
+import sys
+
+import qubesadmin.tools
+import qubesadmin.utils
+
+
+def get_parser():
```
ben-grande: Unfortunately the parser has to be on the module level for the docs to be build, see CI.

ben-grande (changes requested):
Looking better. Please see CI failures.

qubesadmin/tools/qvm_restart.py
RandyTheOtter: Correct

qubesadmin/tools/qvm_restart.py
RandyTheOtter: Wow that method looks useless, and requires making parser available. Why does it exist?

Regardless, I can't use it if I want to have output without "Error:" in each print. Adopted for pre-report messages.

RandyTheOtter:
Pylint is complaining that test lines too long, when broken apart they're less readable.
Tests are failing due to qubesadmin.vm failing to get_power_state(). Doesn't seem like my fault, not sure what to think of this.

RandyTheOtter:
Seems like `get_power_state()` needs something else in the mock response, I should look into this later on

ben-grande:
> Pylint is complaining that test lines too long, when broken apart they're less readable.

Use `black -l80 FILE`.

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,141 @@
...
+            "Can not restart: "
+            + ", ".join(vm.name for vm in invalid_domains) +
+            "\nRefusing to shut down dom0 and unnamed DispVMs;"
+            " Ensure data is saved and shut down manually."
+        )
+
+    if not args.start:
+        target_domains = [
+            vm for vm in target_domains
+            if vm.get_power_state() == "Running"
```
ben-grande: Change this to `if vm.is_running()`?

ben-grande:
I have a use case for this to be in "utils" module (without expecting a command-line client), so I can easily use here (see how other actions were done):

- https://github.com/QubesOS/qubes-core-admin-client/pull/333

qubesadmin/tests/tools/qvm_restart.py
```diff
@@ -0,0 +1,170 @@
...
+        self.assertAllCalled()
+
+    def test_002_restart_halted_startopt(self):
+        """Trying to restart a halted qube with `--start` option"""
+        self.app.expected_calls[("dom0", "admin.vm.List", None, None)] = (
+            b"0\x00some-vm class=AppVM state=Halted\n"
+        )
+        self.app.expected_calls[
+            ("some-vm", "admin.vm.Shutdown", "wait", None)
+        ] = b"0\x00"
```
marmarek, line 64: This should be:

```suggestion
        ] = b"2\x00QubesVMNotStartedErrorDomain is powered off: 'some-vm'"
```

qubesadmin/tests/tools/qvm_restart.py
```diff
@@ -0,0 +1,170 @@
...
+        ] = b"0\x00default=True type=bool False"
+        self.app.expected_calls[
+            ("some-vm", "admin.vm.Shutdown", "force+wait", None)
+        ] = b"0\x00"
+        self.app.expected_calls[("some-vm", "admin.vm.Start", None, None)] = (
+            b"0\x00"
+        )
+        self.app.expected_calls[
+            ("dormant-vm", "admin.vm.Shutdown", "force+wait", None)
+        ] = b"0\x00"
```
marmarek, line 134: And similar here.

marmarek (changes requested):
Besides comments below, can you squash the commits (and possibly add co-authored-by tag to the commit message to retain info about Ali)?

qubesadmin/tests/tools/qvm_restart.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+class TC_00_qvm_restart(qubesadmin.tests.QubesTestCase):
+    def test_000_restart_running(self):
+        """Restarting just one already running qube"""
+        self.app.expected_calls[("dom0", "admin.vm.List", None, None)] = (
+            b"0\x00some-vm class=AppVM state=Running\n"
+        )
+        self.app.expected_calls[("some-vm", "admin.vm.CurrentState", None, None)] = (
+            b"0\x00some-vm mem=42069 mem_static_max=42069 cputime=1337 power_state=Running\n"
```
ben-grande: Example return type of `admin.vm.CurrentState`
```sh
% qubesd-query -e dom0 admin.vm.CurrentState dom0 | cat -A
0^@mem=4000000 mem_static_max=16000 cputime=440000000 power_state=Running% 
```
So know you know where the null is, just first character. My shell shows that there is no newline, by printing a `%` with reversevideo at the end.

If there were newline, it would have printed `$`:
```sh
% printf '\n' | cat -A
$
```
But on your shell, if it is not customized to handle output missing newline, your prompt will be on the same line as the output like this:

```bash
% bash --norc
bash-5.2$ printf '\0' | cat -A
^@bash-5.2$
```

If you want to see null representation and no newline without newline:
```sh
% printf '\0' | cat -A
^@%
```

marmarek: If you want to see precise bytes, I'd recommend `xxd` or `hexdump -C`.

ben-grande:
For the `line-too-long` pylint issue, use `black` formater with `black -l80 FILE`. Don't use `black -l80 .` as a lot of files on this repo are not formatted and there are open PRs that may be impacted.

qubesadmin/tests/tools/qvm_restart.py
RandyTheOtter: CurrentState responses are too long even with `black -l 80`, looking at the other tests it seems like I can cheat a little and mock only `power_state` part of the response. It fixed some failing tests as well

RandyTheOtter: Whoops, I replied to the wrong conversation...

Anyway, `QubesVMNotStartedErrorDomain` is wrong, shutdown doesn't fail if target is down. Correct me if it's an upcoming change I haven't seen yet, but with current admin it replies `0\x00`.

RandyTheOtter:
> For the `line-too-long` pylint issue, use `black` formater with `black -l80 FILE`. Don't use `black -l80 .` as a lot of files on this repo are not formatted and there are open PRs that may be impacted.

CurrentState responses are too long even with black -l 80, looking at the other tests it seems like I can cheat a little and mock only power_state part of the response. It have fixed some failing tests as well... Somehow...

ben-grande:
I was going to say that this is wrong, as you can break strings into multiple lines, but there are other tests, such as `tests.tools.qvm_start.py`, doing the same thing as only returning the necessary keys. It's still wrong, as it may cause problem in the future, but maybe it's fine for now?

qubesadmin/utils/__init__.py
```diff
@@ -519,3 +519,27 @@ def start_expert(
...
+    """
+    Core functionality of qvm-restart. Made to be imported from other python 
+    modules.
+
+    Unlike the main routine, doesn't do any input verification - this is up to
+    the invoking module.
+    """
+    failed = {}
+    failed['shutdown'] = asyncio.run(
+        qubesadmin.utils.shutdown(
```
ben-grande: We are in `utils` module, we can use just `shutdown`

qubesadmin/utils/__init__.py
```diff
@@ -519,3 +519,27 @@ def start_expert(
...
+def restart(domains,force=False):
+    """
+    Core functionality of qvm-restart. Made to be imported from other python 
+    modules.
+
+    Unlike the main routine, doesn't do any input verification - this is up to
+    the invoking module.
+    """
+    failed = {}
+    failed['shutdown'] = asyncio.run(
```
ben-grande: Let's avoid `asyncio.run`, follow this function format:

https://github.com/RandyTheOtter/qubes-core-admin-client/blob/169d1a9f66f881718ef8c776d095b1312c8314a8/qubesadmin/utils/__init__.py#L306

Same thing for `start` below.

ben-grande (changes requested):

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,134 @@
...
+    """Report failures as a table of `qube_name "reason"`. Returns a list of
+    strings with trailing newlines, ready for output in a single print call.
+
+    :param data: Information to report; a dictionary of:
+                 {qubesadmin.vm.QubesVM: str or Exception}
+    """
+    nwidth = str(len(max(
+        (vm.name for vm in data),
+        key=len
+    )))
```
marmarek:
```suggestion
    nwidth = str(max(len(vm.name) for vm in data))
```

qubesadmin/utils/__init__.py
```diff
@@ -519,3 +519,27 @@ def start_expert(
             except qubesadmin.exc.QubesException:
                 pass
         raise e
+
+
+def restart(domains,force=False):
```
ben-grande: Please:

- Make it `async def`
- Add space after the comma, as in: `domains, force=False`.
- add an option for toggling `shutdown` and `kill`, default to `shutdown`

qubesadmin/utils/__init__.py
ben-grande: So, make it 

```python3
        failed['shutdown'] = await shutdown(
            domains=domains, force=force, wait=True
        )
```

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,131 @@
...
+        target_domains = [
+            vm for vm in target_domains
+            if vm.is_running()
+        ]
+
+    failed = qubesadmin.utils.restart(target_domains,args.force)
+
+    if failed['shutdown']:
+        parser.print_error(
+            "Failed to restart domains:\n",
```
ben-grande, line 125: I think the word "shutdown" fits better here.

qubesadmin/utils/__init__.py
```diff
@@ -519,3 +519,27 @@ def start_expert(
...
+        )
+    )
+    failed['start'] = asyncio.run(
+        start(
+            domains=[
+                vm for vm in domains if vm not in failed['shutdown']
+            ]
+        )
+    )
+    return failed
```
ben-grande: We have to consider that every other method using `generic_action` has the same return type, which this functions changes from `dict[QubesVM, BaseException]` to `dict[str, dict[QubesVM, BaseException]]`. I am not sure what to do here, I don't like inconsistencies and would like that calling any of these actions returned the same type, so we don't need separate evaluation of the result depending on the method called, but I also understand that you use it for logging in `qvm_restart`. Maybe you could evaluate based on known exceptions for each call, and then do a safe fallback to generic logging when the exception is not known.

ben-grande (changes requested):

qubesadmin/utils/__init__.py
RandyTheOtter: Do we have a purpose for consistency here? Exception matching looks like avoidable maintenance burden considering that we already know which instruction returns what.

qubesadmin/tools/qvm_restart.py
RandyTheOtter: Why? Shutdown failure implies startup failure, thus it is a restart failure.

ben-grande: Restart is made of "shutdown + start". Being specific that it is the "shutdown" is to be explicit and avoid ambiguous meaning.

qubesadmin/utils/__init__.py
ben-grande: https://github.com/QubesOS/qubes-core-admin-client/pull/333

A single function that deals with all qube state wrappers. If the restart wrapper has a return signature, I'd need to create a separate handling just for that.

Ideally, there would be a parent class `QubesVMShutdownError` and `QubesVMStartError`, so that you could check if `QubesVMShutdownTimeoutError` or `QubesMemoryError` is part of which, so you would return with the same signature, and the command-line tool that intendes to log the information, would us `isinstance` to check hierarchy.

qubesadmin/tools/qvm_restart.py
```diff
@@ -0,0 +1,151 @@
...
+            end=""
+        )
+    failed_start = await restart.__anext__()
+    if failed_start:
+        parser.print_error(
+            "Failed to start domains back up:\n",
+            *report_failure(failed_start),
+            end=""
+        )
+
```
marmarek, line 136: `restart` generator needs to be closed:
```
await restart.aclose()
```

marmarek (changes requested):
CI complains about undocumented option:
```
Undocumented arguments for command 'qvm-restart': '--kill, -k'
```

PR #471, opened 2026-05-26
Enable some old expectedFailures for qvm_run

### LINKED ISSUE #2916: Fix tests for qvm-run
#### Qubes OS version (e.g., `R3.2`): R4.0

https://github.com/QubesOS/qubes-core-admin-client/pull/14#issuecomment-315910818

Since qvm-run use separate process for handling stdin, data handled there is not passed back to the main process (tests mock `proc.stdin` with `io.BytesIO()`). Needs some better way to test this.

ben-grande:
For: https://github.com/qubesos/qubes-issues/issues/2916

---

Didn't test on my system. Relying on unit tests.

### DIFF
--- a/qubesadmin/tests/tools/qvm_run.py
+++ b/qubesadmin/tests/tools/qvm_run.py
@@ -37,9 +37,6 @@ class TC_00_qvm_run(qubesadmin.tests.QubesTestCase):
             self.skipTest("qvm-run change behavior on redirected stdout/stderr")
         super().setUp()
 
-    def default_filter_esc(self):
-        return os.isatty(sys.stdout.fileno())
-
     def test_000_run_single(self):
         self.app.expected_calls[("dom0", "admin.vm.List", None, None)] = (
             b"0\x00test-vm class=AppVM state=Running\n"
@@ -358,22 +355,21 @@ class TC_00_qvm_run(qubesadmin.tests.QubesTestCase):
                     },
                 ),
                 # TODO: find a way to compare b'some-data\n' sent from another
-                # proces
+                # procces
             ],
         )
         self.assertAllCalled()
 
-    @unittest.expectedFailure
     def test_002_color_output(self):
         self.app.expected_calls[("dom0", "admin.vm.List", None, None)] = (
             b"0\x00test-vm class=AppVM state=Running\n"
         )
         self.app.expected_calls[
+            ('test-vm', 'admin.vm.CurrentState', None, None)
+        ] = b'0\x00power_state=Running\n'
+        self.app.expected_calls[
             ("test-vm", "admin.vm.feature.CheckWithTemplate", "os", None)
         ] = b"2\x00QubesFeatureNotFoundError\x00\x00Feature 'os' not set\x00"
-        # self.app.expected_calls[
-        #     ('test-vm', 'admin.vm.List', None, None)] = \
-        #     b'0\x00test-vm class=AppVM state=Running\n'
         stdout = io.StringIO()
         with unittest.mock.patch("sys.stdout", stdout):
             ret = qubesadmin.tools.qvm_run.main(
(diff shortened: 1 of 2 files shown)


qubesadmin/tests/tools/qvm_run.py
```diff
@@ -283,17 +283,16 @@ def test_002_passio(self):
         with subprocess.Popen(
             ["echo", "some-data"], stdout=subprocess.PIPE
         ) as echo:
-            with unittest.mock.patch("sys.stdin", echo.stdout):
```
ben-grande, line 286: Why remove this? Cause later its already testing is stdin, stdout and stderr are None.

ben-grande: Reverted this, not sure if I should touch that in this PR or commit. Maybe call a subprocess for shell and pass stdin that way, and capture stdout and stderr also that way?

qubesadmin/tests/tools/qvm_run.py
```diff
@@ -1352,7 +1347,7 @@ def test_029_command_is_dashdash(self):
             [
                 (
                     "@dispvm:test-vm",
-                    "qubes.VMExec+----",
+                    "qubes.VMExec+----+----",
```
marmarek, line 1354: I don't think this is correct. `--` (the first one) means end of options, it shouldn't be considered part of the command itself.

qubesadmin/tests/tools/qvm_run.py
```diff
@@ -1321,7 +1316,7 @@ def test_028_argparse_bug_workaround(self):
             [
                 (
                     "@dispvm:test-vm",
-                    "qubes.VMExec+command+----",
+                    "qubes.VMExec+command+----+----",
```
marmarek, line 1322: And here too, the old behavior was correct.

qubesadmin/tests/tools/qvm_run.py
```diff
@@ -409,26 +405,27 @@ def test_002_color_output(self):
...
-        # self.app.expected_calls[
-        #     ('test-vm', 'admin.vm.List', None, None)] = \
-        #     b'0\x00test-vm class=AppVM state=Running\n'
+        self.app.expected_calls[
+            ('test-vm', 'admin.vm.CurrentState', None, None)
+        ] = b'0\x00power_state=Running\n'
+        self.app.expected_calls[
+            ('test-vm', 'admin.vm.feature.CheckWithTemplate', 'os', None)
+        ] = b"2\x00QubesFeatureNotFoundError\x00\x00Feature 'os' not set\x00"
         stdout = io.StringIO()
```
marmarek, line 418: This is unused after your change (the assert at the end is pointless with the mock.patch removed).

marmarek (changes requested):

qubesadmin/tests/tools/qvm_run.py
ben-grande: Command: `["--no-gui", "--dispvm=test-vm", "command", "--", "--"]`.

So when the command `command` is already the first positional argument, everything else should be passed to the `VMExec` and not be trimmed by qvm-run.

qubesadmin/tests/tools/qvm_run.py
ben-grande: In this case, yes. Old one was correct.

qubesadmin/tests/tools/qvm_run.py
marmarek: `--` is "end of options" marker, not "start of positional arguments" marker. It only means options shouldn't be interpreted after it, nothing more. See [POSIX spec](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap12.html) (Guideline 10).
See for example behavior of `ls`:
- `ls - l foo -- bar`
- `ls foo -l -- -- bar`
- `ls foo -- bar -l`

That said, `qvm-run` already enforces all options to be passed before the command (for example `qvm-run VM ls -p` is not the same as `qvm-run -p VM ls`). So, maybe indeed in this case your approach is more consistent.

ben-grande: This is hard to do with argparse. It eats "--" anywhere even though option arguments are not allowed anymore at the place "--" is, because a preceding positional argument exists. Marking this as draft, but probably won't be solved.

ben-grande:
The dash-dash handling is haunted, I will drop the changes related to that so only the normal skipped tests can be enabled again and the issue can be closed.

qubesadmin/tests/tools/qvm_run.py
marmarek: What about this comment? Your change made the test not testing intended thing anymore.

marmarek (changes requested):

ben-grande:
PipelineRetryFailed

qubesadmin/tests/tools/qvm_run.py
marmarek: So, this change should be reverted.

marmarek (approved):

PR #486, opened 2026-07-29
Add standalone disposable template to mock_app

ben-grande:
Helps test applications that tries to query different property set from the more common disposable template klass, AppVM. StandaloneVMs don't have "template" attribute for example.

For: https://github.com/QubesOS/qubes-desktop-linux-manager/pull/323

### DIFF
--- a/qubesadmin/tests/mock_app.py
+++ b/qubesadmin/tests/mock_app.py
@@ -1073,6 +1073,16 @@ class MockQubesComplete(MockQubes):
             running=True,
         )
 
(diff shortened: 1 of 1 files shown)


marmarek:
Have you tried running full tests of desktop-linux-manager and manager with this change? AFAIR some of them are sensitive to changes in the number of qubes.

ben-grande:
Tested on:

- [x] `desktop-linux-manager`
- [x] `manager` (one error there but not related to this PR)

marmarek (approved):

PR #490, opened 2026-08-07
Fix man page about qvm-run --gui and --no-gui

iasds:
Per https://github.com/QubesOS/qubes-issues/issues/9181, --gui/--no-gui do not control GUI forwarding. They control whether qvm-run waits for the qube's GUI session to start before running the command (and which RPC service is used for DisposableVMs). Update the man page to describe the actual behavior. Still affects 4.3.

### DIFF
--- a/doc/manpages/qvm-run.rst
+++ b/doc/manpages/qvm-run.rst
@@ -80,13 +80,16 @@ Options
 
 .. option:: --gui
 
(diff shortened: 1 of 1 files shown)


crass:
@marmarek I was confused about the `--gui` option also. This seems like an easy merge.

marmarek (approved):

PR #495, opened 2026-08-17
Reject usernames containing colon in run_service

### LINKED ISSUE #9154: Colon in user results in invalid qrexec command

### Qubes OS release

R4.2, but this problem is probably older than R4.1.

### Brief summary

If one passes `--user=a:b` (or any other string containing a colon) to `qvm-run` in dom0, the colon is interpreted as the end of the username, rather than as part of it.

### Steps to reproduce

Either:

1. Pass a username to `qvm-run --user=` that contains a colon.
2. Include a `user=` keyword in a qrexec policy that contains a colon.

### Expected behavior

Error stating that usernames in qrexec cannot contain a colon.

### Actual behavior

Colon is interpreted as end of the username, and everything after it is interpreted as the command.  For instance, `qvm-run --user=DEFAULT:nogui` is equivalent to `qvm-run --no-gui`.

datenbar:
Fixes QubesOS/qubes-issues#9154

## Problem
run_service() in qubesadmin/app.py builds the qrexec command as "{user}:QUBESRPC ..." without validating that user itself doesn't contain a ' : '. On the receiving end, parse_qubes_rpc_command() (qubes-core-qrexec/libqrexec/exec.c) splits on the first colon it finds, so a colon inside user gets misread as the delimiter silently truncating the username and misinterpreting the rest as part of the command.

## Fix
Added a check in run_service() that raises ValueError if user contains a ' : ' before user is used to build the command string.

## Testing
Added test_015_run_service_user_colon in qubesadmin/tests/app.py. Full suite: 63 passed, 1 pre-existing unrelated failure (test_014_run_service_prefix_data, fails identically on unmodified main appears to be an os.isatty mocking issue unrelated to this change).

### Notes 
I noticed parse_qubes_rpc_command() on the C side (qubes-core-qrexec) has no independent safeguard against this either. Happy to open that as a separate follow-up if useful, since fixing it here felt out of scope.

### Disclosure
This is my first contribution. I used Claude AI to help locate the relevant repositories and functions across qubes-core-admin-client and qubes-core-qrexec, and to review the fix and test before submitting. The fix, test, and verification are my own.

### DIFF
--- a/qubesadmin/app.py
+++ b/qubesadmin/app.py
@@ -1016,6 +1016,8 @@ class QubesLocal(QubesBase):
             qrexec_opts.extend(["-l", localcmd])
         if user is None:
             user = "DEFAULT"
+        if ":" in user:
+            raise ValueError("Username in qrexec cannot contain a colon")
         if not wait:
             qrexec_opts.extend(["-e"])
         if prefix_data:
--- a/qubesadmin/tests/app.py
+++ b/qubesadmin/tests/app.py
@@ -1119,6 +1119,14 @@ class TC_20_QubesLocal(unittest.TestCase):
             self.get_request(), b"admin.vm.Start+ dom0 name some-vm\0"
         )
 
+    @mock.patch.object(sys, 'stderr')
+    @mock.patch('os.isatty', lambda fd: fd == 2)
+    def test_015_run_service_user_colon(self, mock_stderr):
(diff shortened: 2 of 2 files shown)


marmarek:
PipelineRetry

marmarek (approved):

PR #497, opened 2026-08-18
Execute backup even if summary is denied by qrexec

ben-grande:
The summary might be denied in case the user didn't allow the call to avoid leaking more information about the backup than necessary. This can happen on non-dom0, where the caller doesn't have the backup passphrase or know about the included and excluded qubes, as some information can be only seen from dom0, such as the backup profile.

For: https://github.com/qubesos/qubes-issues/issues/11062

### DIFF
--- a/qubesadmin/tools/qvm_backup.py
+++ b/qubesadmin/tools/qvm_backup.py
@@ -34,7 +34,7 @@ try:
 except ImportError:
     have_events = False
 import qubesadmin.tools
-from qubesadmin.exc import QubesException
+from qubesadmin.exc import QubesException, QubesDaemonAccessError
 
 backup_profile_dir = '/etc/qubes/backup'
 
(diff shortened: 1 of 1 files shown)


qubesadmin/app.py
```diff
@@ -1083,9 +1083,10 @@ def qubesd_call(
             ) as p:
                 (stdout, stderr) = p.communicate(payload)
         if p.returncode != 0:
-            raise qubesadmin.exc.QubesDaemonAccessError(
-                "Service call error: %s", stderr.decode()
-            )
+            msg = "Service call error: {}".format(stderr.decode())
+            if p.returncode == 126:
+                raise qubesadmin.exc.QrexecDeniedError(msg)
```
marmarek, line 1088: I don't think it needs to be a separate exception here. While it may allow differentiating one specific case, it wouldn't for example cover backup info being denied by an admin-permission event handler (which may happen even if qvm-backup is running in dom0). `QubesDaemonAccessError` should be good enough for the purpose of this PR.

But even if going with separate exception, I don't like "qrexec" in its name - it's too specific to a particular place and can't be re-used. For example, when call is made from dom0, qrexec isn't involved at all, yet the Admin API call still can be denied by other means. So, **if** leaving separate exception, maybe something like `QubesDaemonAccessDeniedError`?

marmarek (changes requested):

qubesadmin/app.py
ben-grande: Opted to use `QubesDaemonAccessError` to not block this PR from being merged.

marmarek (approved):

PR #499, opened 2026-08-24
Fix ordering of dependent qube actions

besendorf:
## Summary

Addresses the qvm-start portion of QubesOS/qubes-issues#10860.

When multiple qubes are requested, dependency auto-starts can change another requested qube's state before its concurrent start task checks it, causing a false "Domain is already running" error.

This change:

- preflights all requested qubes before starting any;
- reports already-running qubes unless `--skip-if-running` is used;
- starts only qubes halted during preflight;
- tolerates state changes caused by dependency auto-starts.

`qvm-shutdown` is intentionally unchanged. Its dependency checks occur server-side and cover audiovm, guivm, netvm, and assigned devices, so a client-side NetVM-only ordering fix would be incomplete.

## Testing

- 38 focused unit tests passed
- `python -m compileall -q qubesadmin`
- `git diff --check`

## AI assistance disclosure

This change was developed with assistance from OpenAI Codex. I reviewed and adapted the implementation and ran the test suite.

### DIFF
--- a/qubesadmin/tests/tools/qvm_start.py
+++ b/qubesadmin/tests/tools/qvm_start.py
@@ -70,6 +70,64 @@ class TC_00_qvm_start(qubesadmin.tests.QubesTestCase):
             1)
         self.assertAllCalled()
 
+    def test_004_preflight_before_start(self):
+        self.app.expected_calls[
+            ('dom0', 'admin.vm.List', None, None)] = \
+            b'0\x00some-vm class=AppVM state=Running\n' \
+            b'other-vm class=AppVM state=Running\n'
+        self.app.expected_calls[
+            ('some-vm', 'admin.vm.CurrentState', None, None)] = \
+            b'0\x00power_state=Running'
+        self.app.expected_calls[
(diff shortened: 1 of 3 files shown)


marmarek:
I don't like this, it makes the common case execute yet another requests (to check the power state), which `start_expert()` already does. Maybe change `start_expert()` to allow skipping that check?
But also, I have an impression that with `--skip-running`, the check can be skipped  completely (both in the qvm-start itself, and in `start_expert()`). `domain.start()` seems to be a no-op for already running qube.

marmarek:
Looks good now, but can you squash the commits together?

besendorf:
Yes sure. Is this something you always would like me to do? In the repo I maintain I just always use the button that github provides to do that

marmarek:
Generally yes, unless there are indeed multiple separate changes in a single PR. Theoretically, I could do that myself, but it would invalidate commit signature, which I'd like to preserve.

marmarek (approved):

PR #502, opened 2026-09-03
Optimize QubesVM._get_derived_vms()

### LINKED ISSUE #11083: qubesadmin QubesVM.derived_vms property is worst-case O(N^2)
### Qubes OS release

Qubes OS 4.3

### Brief summary

@deeplow discovered that the `derived_vms` property is unexpectedly slow: https://github.com/freedomofpress/securedrop-workstation/pull/1793#pullrequestreview-4908622584. When I looked into it, I noticed it was using recursion. I used Claude to verify my suspicion, which is that the current implementation is worst-case O(N^2) but it could be implemented as O(N).

### Steps to reproduce

1. Access the `derived_vms` property on a VM using qubesadmin

### Expected behavior

It only needs to traverse the list of VMs once, i.e. O(N) performance.

### Actual behavior

The implementation goes over the list of VMs for each VM, which is roughly O(N^2)

### Additional information

The simple workaround is to merely enable caching, which is what we've done in SecureDrop for now. I can submit a PR for this.

legoktm:
Instead of recursing in a way that effectively made it O(N^2), build the map of template -> children once, and then iterate over it, giving us effectively O(N) performance.

Fixes <https://github.com/QubesOS/qubes-issues/issues/11083>.

----

As mentioned in the issue, I used Claude Opus 5 to figure out this bug and then to write the initial patch, which I then cleaned up and rewrote the comments.

### DIFF
--- a/qubesadmin/vm/__init__.py
+++ b/qubesadmin/vm/__init__.py
@@ -438,9 +438,28 @@ class QubesVM(qubesadmin.base.PropertyHolder):
         Return `set` of all domains based on the current TemplateVM
         at any level of inheritance.
         """
-        result = set(vm.appvms)
-        for appvm in vm.appvms:
-            result.update(QubesVM._get_derived_vms(appvm))
+        # first build a map of templates -> children
+        children = {}
+        for domain in vm.app.domains:
+            try:
+                template = domain.template
+            except AttributeError:
+                continue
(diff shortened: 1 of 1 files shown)


ben-grande:
Looks correct.

marmarek (approved):

PR #504, opened 2026-09-06
Cache direct VM feature and tag reads

### LINKED ISSUE #11082: VM features and tags in qubes-core-admin-client are not cached the way properties are

`qubesadmin` caches properties, but not features or tags.

With `cache_enabled` set, the first property access on a VM triggers `_fetch_all_properties()` in `qubesadmin/base.py`, which issues a single `admin.vm.property.GetAll` call and serves every later property read from `_properties_cache`.

`Features` in `qubesadmin/features.py` has no equivalent. `__getitem__` issues `admin.vm.feature.Get` on every read, and `check_with_template` issues `admin.vm.feature.CheckWithTemplate` on every read, whether or not `cache_enabled` is set. `Tags` in `qubesadmin/tags.py` behaves the same way: `__contains__` issues `admin.vm.tag.Get` on every membership test.

This is expensive for GUI tools that build lists of qubes, because reading one feature per qube becomes one qubesd round trip per qube. Counting calls in Qubes Global Config against `qubesadmin.tests.mock_app.MockQubesComplete`, which has 20 qubes, constructing all nine settings pages costs 584 Admin API calls:

| Method                     | Calls |
|----------------------------|-------|
| `admin.vm.feature.Get`     |   343 |
| `admin.vm.tag.Get`         |    57 |
| `admin.vm.property.GetAll` |    20 |

Features and tags account for just over two thirds of the calls. Properties, read from the same 20 qubes, cost 20 calls in total because they are cached.

The counts come from the mock backend rather than wall-clock measurement on a real system, where each call is a round trip to qubesd.

### The solution you'd like

Cache features and tags per VM when `cache_enabled` is set, in the same manner as properties: fetch the whole set once, then serve reads from the cache. `admin.vm.feature.List` and `admin.vm.tag.List` already exist, so the change should not need a new Admin API method, and `_invalidate_cache_all()` already provides the invalidation hook that callers such as Global Config use around saves.

`check_with_template` needs care, since its result depends on the template as well as the qube, so it may need either its own cache keyed on both or to stay uncached at first.

This would benefit every Qubes GUI tool that builds a list of qubes, not only Global Config. Global Config is simply where the cost was measured, in QubesOS/qubes-issues#11081.

AI disclosure: I researched everything on my own, but i let my local LLM generate this description out of my bullet points.

andreasglashauser:
This adds caching to VM feature and tag reads. I implemented this as stated in the issue and I respected @ben-grande's comment.

Fixes https://github.com/QubesOS/qubes-issues/issues/11082

AI disclosure: some (less than half) of the tests where implemented by gpt-6-astra. I also used it to review my code.

### DIFF
--- a/qubesadmin/features.py
+++ b/qubesadmin/features.py
@@ -57,19 +57,38 @@ class Features:
         self._missing_cache.clear()
         self._names_cache = None
 
+    def record_value(self, key: str, value: str) -> None:
+        '''Cache a feature value confirmed by qubesd.'''
+        if not self.vm.app.cache_enabled:
+            self.clear_cache()
+            return
+        self._missing_cache.pop(key, None)
+        self._values_cache[key] = value
+        if self._names_cache is not None and key not in self._names_cache:
+            self._names_cache.append(key)
+
+    def record_removal(self, key: str) -> None:
+        '''Drop a feature that qubesd confirmed as removed.'''
+        if not self.vm.app.cache_enabled:
+            self.clear_cache()
+            return
+        self._values_cache.pop(key, None)
+        if self._names_cache is not None and key in self._names_cache:
+            self._names_cache.remove(key)
+
     def __delitem__(self, key: str) -> None:
-        self.clear_cache()
         self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Remove', key)
+        self.record_removal(key)
 
     def __setitem__(self, key: str, value: object) -> None:
-        self.clear_cache()
         if isinstance(value, bool):
             # False value needs to be serialized as empty string
-            self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
-                b'1' if value else b'')
+            serialized = '1' if value else ''
         else:
-            self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
-                str(value).encode())
+            serialized = str(value)
+        self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
+                            serialized.encode())
+        self.record_value(key, serialized)
 
     def __getitem__(self, item: str) -> str:
         if not self.vm.app.cache_enabled:
@@ -77,7 +96,6 @@ class Features:
         if item in self._values_cache:
             return self._values_cache[item]
         if item in self._missing_cache:
-            # QubesException formats messages even without arguments.
             raise QubesFeatureNotFoundError(
                 self._missing_cache[item].replace('%', '%%'))
         try:
--- a/qubesadmin/features.py
+++ b/qubesadmin/features.py
@@ -25,6 +25,8 @@ import typing
 from typing import TypeVar
 from collections.abc import Iterator, Generator
 
+from qubesadmin.exc import QubesFeatureNotFoundError
+
 if typing.TYPE_CHECKING:
     from qubesadmin.vm import QubesVM
 
@@ -45,11 +47,22 @@ class Features:
     def __init__(self, vm: QubesVM):
         super().__init__()
         self.vm = vm
+        self._values_cache: dict[str, str] = {}
+        self._missing_cache: dict[str, str] = {}
+        self._names_cache: list[str] | None = None
+
+    def clear_cache(self) -> None:
+        '''Discard cached direct values, missing features and feature names.'''
+        self._values_cache.clear()
+        self._missing_cache.clear()
+        self._names_cache = None
 
     def __delitem__(self, key: str) -> None:
+        self.clear_cache()
         self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Remove', key)
 
     def __setitem__(self, key: str, value: object) -> None:
+        self.clear_cache()
         if isinstance(value, bool):
             # False value needs to be serialized as empty string
             self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
@@ -59,13 +72,36 @@ class Features:
                 str(value).encode())
 
     def __getitem__(self, item: str) -> str:
-        return self.vm.qubesd_call(
-            self.vm.name, 'admin.vm.feature.Get', item).decode('utf-8')
+        if not self.vm.app.cache_enabled:
+            self.clear_cache()
+        if item in self._values_cache:
+            return self._values_cache[item]
+        if item in self._missing_cache:
+            # QubesException formats messages even without arguments.
+            raise QubesFeatureNotFoundError(
+                self._missing_cache[item].replace('%', '%%'))
+        try:
+            value = self.vm.qubesd_call(
+                self.vm.name, 'admin.vm.feature.Get', item).decode('utf-8')
+        except QubesFeatureNotFoundError as error:
+            if self.vm.app.cache_enabled:
+                self._missing_cache[item] = error.args[0]
+            raise
+        if self.vm.app.cache_enabled:
+            self._values_cache[item] = value
+        return value
 
     def __iter__(self) -> Iterator[str]:
+        if not self.vm.app.cache_enabled:
+            self.clear_cache()
+        if self._names_cache is not None:
+            return iter(self._names_cache)
         qubesd_response = self.vm.qubesd_call(self.vm.name,
             'admin.vm.feature.List')
-        return iter(qubesd_response.decode('utf-8').splitlines())
+        names = qubesd_response.decode('utf-8').splitlines()
+        if self.vm.app.cache_enabled:
+            self._names_cache = names
+        return iter(names)
 
     keys = __iter__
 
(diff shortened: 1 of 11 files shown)


qubesadmin/vm/__init__.py
```diff
@@ -77,6 +77,8 @@ def name(self) -> str:
 
     @name.setter
     def name(self, new_value):
+        self.features.clear_cache()
+        self.tags.clear_cache()
```
ben-grande, line 81: Is there are reason this is on top? Does this somehow affects `qubesd_call`? Can the `clear_cache` be kept together with `self.app.domains.clear_cache()`?

qubesadmin/features.py
```diff
@@ -45,11 +47,22 @@ class Features:
...
+        self._names_cache: list[str] | None = None
+
+    def clear_cache(self) -> None:
+        '''Discard cached direct values, missing features and feature names.'''
+        self._values_cache.clear()
+        self._missing_cache.clear()
+        self._names_cache = None
 
     def __delitem__(self, key: str) -> None:
+        self.clear_cache()
```
ben-grande, line 61: Why discarding the whole cache when a single feature is set or removed? Why not keep up with it?

Also, I think that invalidation should come after `qubesd_call`, because if the call fails, for example, `permission-denied`, the cache shouldn't be cleared as the client didn't receive a successful response.

qubesadmin/features.py
```diff
@@ -45,11 +47,22 @@ class Features:
...
         self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Remove', key)
 
     def __setitem__(self, key: str, value: object) -> None:
+        self.clear_cache()
```
ben-grande, line 65: https://github.com/QubesOS/qubes-core-admin-client/pull/504/changes#r3948192032

qubesadmin/features.py
```diff
@@ -59,13 +72,36 @@ def __setitem__(self, key: str, value: object) -> None:
...
-        return self.vm.qubesd_call(
-            self.vm.name, 'admin.vm.feature.Get', item).decode('utf-8')
+        if not self.vm.app.cache_enabled:
+            self.clear_cache()
+        if item in self._values_cache:
+            return self._values_cache[item]
+        if item in self._missing_cache:
+            # QubesException formats messages even without arguments.
+            raise QubesFeatureNotFoundError(
+                self._missing_cache[item].replace('%', '%%'))
```
ben-grande, line 82: You can avoid escaping if you use the unsanitized value as a value instead of format,

```python3
"%s" % self._missing_cache[item]
```

qubesadmin/features.py
```diff
@@ -98,6 +134,8 @@ def check_with_template(self, item: str, default: T) -> str | T: ...
     def check_with_template(self, feature: str,
                             default: object = None) -> object:
         ''' Check if the vm's template has the specified feature. '''
+        if not self.vm.app.cache_enabled:
+            self.clear_cache()
```
ben-grande, line 138: Why the cache is cleared here?

qubesadmin/tags.py
```diff
@@ -61,11 +70,25 @@ def discard(self, elem: str) -> None:
...
-        return iter(qubesd_response.decode('utf-8').splitlines())
+        names = qubesd_response.decode('utf-8').splitlines()
+        if self.vm.app.cache_enabled:
+            self._names_cache = names
+        return iter(names)
 
     def __contains__(self, elem: str) -> bool:
         '''Does the VM have a tag'''
+        if not self.vm.app.cache_enabled:
+            self.clear_cache()
```
ben-grande, line 87: Why the cache is cleared here? If you want to invalidate the cache if `cache_enabled` was once enabled and then disabled, maybe there could be a setter for that property?

ben-grande (changes requested):

qubesadmin/vm/__init__.py
andreasglashauser: Not really and actually its not needed at all. I removed it from the setter, the property-set:name in the events dispatcher and the two tests for it.

qubesadmin/features.py
andreasglashauser: Good points, both changed.

qubesadmin/features.py
andreasglashauser: already fixed

qubesadmin/features.py
andreasglashauser: Tried that, but it didnt surve the tests because __init__ in QubesCeption turns arguments like "001" into "1" and it crashed on "²". I think thats an pre-existing but which should get fixed in a separate PR because it hits every error argument.

qubesadmin/features.py
andreasglashauser: Gone now

qubesadmin/tags.py
andreasglashauser: Yeah that was the intention. The setter is the better place for it. Fixed.

qubesadmin/tests/app.py
```diff
@@ -1321,3 +1323,32 @@ def test_016_run_service_prefix_data(self):
...
+
+class TC_30_ContainerCacheInvalidation(qubesadmin.tests.QubesTestCase):
+    def test_disabling_cache_invalidates_everything(self) -> None:
+        flow = []
+        with mock.patch.object(self.app, '_invalidate_cache_all',
+                               side_effect=lambda: flow.append('invalidate')):
+            for is_enabled in (False, True, True, False, False):
+                self.app.cache_enabled = is_enabled
+                flow.append(str(is_enabled))
+        self.assertEqual(', '.join(flow),
```
marmarek: I don't think you need to convert it to string, you can simply compare lists.

qubesadmin/tests/app.py
```diff
@@ -1321,3 +1323,32 @@ def test_016_run_service_prefix_data(self):
...
+        with ExitStack() as patches:
+            for vm in (first_vm, second_vm):
+                for container_name in ('features', 'tags'):
+                    patches.enter_context(mock.patch.object(
+                        getattr(vm, container_name), 'clear_cache',
+                        side_effect=partial(flow.append,
+                                            f'{vm.name} {container_name}')))
+            self.app._invalidate_cache_all()  # pylint: disable=protected-access
+        self.assertEqual(
+            f'{", ".join(sorted(flow))}; calls={len(self.app.actual_calls)}',
```
marmarek: same here

qubesadmin/tests/events.py
```diff
@@ -264,3 +265,83 @@ def test_030_events_device(self):
...
+                 False, 'new, False; calls=1'),
+                ('domain-tag-add:tag', {'tag': 'tag'}, False,
+                 'old, True; calls=0'),
+                ('domain-tag-delete:tag', {'tag': 'tag'}, True,
+                 'old, False; calls=0')):
+            with self.subTest(event=event):
+                self.prime_caches(has_tag)
+                snapshots = self.read_callback_snapshots(event, **kwargs)
+                other = self.read_snapshot(self.other_vm)
+                self.assertEqual(
```
marmarek: Same here (the `expected` value can be a tuple if you want).

qubesadmin/tests/app.py
```diff
@@ -1321,3 +1323,32 @@ def test_016_run_service_prefix_data(self):
...
+                self.app.cache_enabled = is_enabled
+                flow.append(str(is_enabled))
+        self.assertEqual(', '.join(flow),
+                         'False, True, True, invalidate, False, False')
+
+    def test_invalidate_existing_vm_containers_without_api_calls(self) -> None:
+        first_vm = self.app.domains.get_blind('first-vm')
+        second_vm = self.app.domains.get_blind('second-vm')
+        flow = []
+        with ExitStack() as patches:
```
marmarek: This looks overly complex, can't you simply patch `qubesadmin.features.Features.clear_cache` (and `.Tags.clear_cache`) using `mock.patch` as a context manager? You do get access to the VM name in a side_effect callback, but probably you don't even need that, you can simply compare `mock_calls` of normal mocks there.

qubesadmin/features.py
```diff
@@ -45,27 +47,73 @@ class Features:
...
-            self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
-                b'1' if value else b'')
+            serialized = '1' if value else ''
         else:
-            self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
-                str(value).encode())
+            serialized = str(value)
+        self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
+                            serialized.encode())
+        self.record_value(key, serialized)
```
marmarek: This looks correct, I think, but I must say it feels risky. I'll test it a bit more...

marmarek (changes requested):

PR #506, opened 2026-09-17
Document reboot properties

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/8196
For: https://github.com/QubesOS/qubes-core-admin/pull/810


crass:
Instead of a boolean `rebootable`, I'd suggest an integer `allowed_reboots` that is the number of reboot requests allowed by the guest. The counter gets reset on shutdown. A value of `-1` would be equivalent to `rebootable` being true and `0` for `rebootable` false. The default could be `revisions_to_keep - 1`. Initially, you could have only the values 0 and -1 being supported if the full implementation is more work than can be given to it right now. This would allow flexibility for future behavior without needed to change the API.

marmarek:
Sounds like a smart idea, especially about resetting the counter on shutdown instead of a time.

ben-grande:
> `allowed_reboots` that is the number of reboot requests allowed by the guest. The counter gets reset on shutdown. A value of `-1` would be equivalent to `rebootable` being true and `0` for `rebootable` false.

The intention of the time threshold is to prevent a vicious reboot cycle, but if we do add a counter, than that only happens at the end of the counter, which might be much better for Windows reboot installations, than guessing the time expiration.

> The default could be `revisions_to_keep - 1`

The default storage type is LVM, the `revisions_to_keep=2`, If I use `revisions_to_keep - 1` means that we'd allow `1` rebooot. Now let's say that a malicious actor wants to delete all sane volumes, all it takes is

1. adding a script that prevents the next boot
2. ask for reboot, on shutdown, a volume is removed
3. on the next boot, the qube shuts down, a volume is removed

The `allowed_reboots` counter will have to remain `0` by default until `revisions_to_keep >= 3`, which I don't think will happen soon.

Usability wise `allowed_reboots`, seems much better, cause once the user doesn't worry about keeping sane volumes (as it is expensive to keep many revisions), not preventing many sudden reboots when the users asks for it, seems much better.

